PR #1, opened 2022-06-16
proof of concept of check execution orchestration, step 1

nelsonkopliku:
Spoiler Alert: the code sucks, gonna make it better :sweat_smile: 

The idea here is that wanda 
- reacts to incoming execution requests from the server
- instruments the target hosts to start gathering facts for checks
- waits for gathered facts published by the target hosts
- launches check evaluation as soon as we're able to do so (ie: all facts are gathered)
- emits checks results
- emits execution progess (running, completed, whatever)

wanda


lib/wanda/node_instrumentation.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Wanda.Instrumentation do
+  @moduledoc """
+  Remote targets instrumentation module.
+  """
+
+  def trigger_facts_gathering(execution_id, target, checks) do
```
arbulu89, line 6: This is the tricky part, right?
Because here, we would need to trigger the fact gathering, and wait until the agents sends back the facts.
I see it nicely done with different GenServer calls in any case.
I'm not sure how the DynamicServer plays here though hehe

arbulu89 (commented):
Hey @nelsonkopliku ,
I find the logic nice, as a first glance of how the orchestration of the checks execution looks like.
Starting this way, already points us out to start thinking how the communications between the checksengine and agent are (dsl and its relatives).

lib/wanda/notes.md
```diff
@@ -0,0 +1,40 @@
...
+
+  information at hand:
+  - an identifier for the currently triggered execution (by the user or by time passing)
+  - cluster id
+  - list of the target hosts (aka agents) of the cluster
+  - list of the selected checks (12ER4, FD54R, ...) - each agent has its own selection
+  - provider (atm this is the cluster's provided - which is simply the discovered cluster of a host...)
+
+  Assumptions:
+  - the selected checks may vary per target agent
```
rtorrero, line 19: You mean within the same cluster?

lib/wanda/node_instrumentation.ex
nelsonkopliku: The `DynamicSupervisor` allows us to spawn an indefinite number of executions, where a normal `Supervisor` accepts static children.

Whatever we call the method, here is where I imagine Wanda publishing towards the agents messages about starting facts gathering for the checks.

Here's a touch-point with the Agent and your comment about the
```
# sbd_config is a hardcoded key to identify which fact gathering to do
# keys is a list with the keys from the list to return
[{"name": "sbd_config", "keys": ["SBD_DEVICE", "SBD_TIMEOUT_ACTION"]}]
```
What information do we have at hand?
What are we gonna send to the Agent?
Most likely here we need to make some steps ahead with the DSL itself

nelsonkopliku: Then, for the receiving (waiting) part, I imagine a component listening for incoming facts gathered and calls the apppropriate method on the ExecutionServer (what we simulate now)
```elixir
Wanda.ExecutionServer.collect_gathered_facts(...)
```
(Gonna improve naming)

lib/wanda/notes.md
```diff
@@ -0,0 +1,40 @@
...
+  what does it mean "requesting an execution for the selection"?
+  could it be something like the following?
+
+  starting the process of gathering facts form targets and evaluating expectations **for all the hosts** on a given cluster
+  and recalculate the health of the single hosts and of the cluster as a whole given the checks results
+
+  is it a usecase to start execution for the check selection of a **single host**?
+
+  information at hand:
+  - an identifier for the currently triggered execution (by the user or by time passing)
```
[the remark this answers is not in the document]
nelsonkopliku: A Check Execution on a Cluster is triggered by either a User clicking on the **Start Execution** button or by a Scheduled action that takes place every X minutes.

lib/wanda/notes.md
```diff
@@ -0,0 +1,40 @@
+  Scenario: we have selected some checks for the hosts of a given cluster and the user triggers an execution
+
+  what does it mean "requesting an execution for the selection"?
+  could it be something like the following?
+
+  starting the process of gathering facts form targets and evaluating expectations **for all the hosts** on a given cluster
+  and recalculate the health of the single hosts and of the cluster as a whole given the checks results
+
+  is it a usecase to start execution for the check selection of a **single host**?
```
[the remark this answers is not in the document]
nelsonkopliku: Nice thinking but yes, sounds like something not strictly necessary, right now at least.

lib/wanda/notes.md
nelsonkopliku: Yes.

As of now what we call a CheckSelection is a list of checks that we want to execute on a target Cluster and the same selection would be applied to all the hosts of that cluster.

However, the user needs the flexibility to configure a more fine grained CheckSelection in order to better suite the complexity of checking SAP infrastructures.
So we're going in the direction where each host has its own Check Selection.

lib/wanda/notes.md
```diff
@@ -0,0 +1,40 @@
...
+
+  Assumptions:
+  - the selected checks may vary per target agent
+  - some agents might not have any selected check
+  - different agents might run on different providers (? this might be skipped atm)
+
+  Process
+  - listen on incoming messages (from the server) about starting the execution for the check selection of a cluster (or of a host)
+  - start an ExecutionServer process for the requested execution
+  - instrument the target agents to run the facts gathering for each of its selected checks (dsl parsing, publishing messages to the agents...)
```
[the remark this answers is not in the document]
nelsonkopliku: Yes, maybe this goes a bit too far in the future but there's an interesting point.

First I believe we need to clarify what a CheckExecution is.
My understanding is the following

> A Check Execution the process of gathering facts form targets and evaluating expectations **for all the hosts** on a given cluster.
> Results are propagated and the hosts' healths and cluster health are recalculated.

---
On facts gathering execution one option is to gather facts on demand, as soon as it is required.

Could you explain a bit more what you see in regards od _fact-gathering doesn't need to happen necessarily during the check execution_?
How do you see that working?

lib/wanda/notes.md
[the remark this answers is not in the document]
rtorrero: Ah, the sentence makes total sense now

PR #3, opened 2022-07-12
Refactor execution pt1

fabriziosestito:
This PR refactors the execution GenServer, separating the functional core from the side effects.
It adds tests and the concept of the result of execution.

Next PRs will add grouped expectations and rabbit messaging.


arbulu89 (approved):
Brilliant work folks!
Really really nice!
Waiting to do the mob session to have a deep dive. 
But I could mostly understand how the code works just in a first initial review.

nelsonkopliku (approved):
:heart: :tropical_drink:

PR #4, opened 2022-08-03
Group expectations evaluation

fabriziosestito:
This PR adds the group expectation evaluation and removes the local one.
An expectation declared with `expect` is expected to return `true` on all the agents that execute  the check.
An expectation declare with `expect_same` is expected to return `the same value` on all the agents that execute the check.

It also refactors the shape of the result struct, aggregating the agent evaluations in check results and the check results in an execution result, so we can remove this responsibility from the dashboard.

ATM we provide basic error handling if an expression is invalid or a fact is missing, we will provide more errors in followup PRs and if we add the rustler module for evaluating expressions.


lib/wanda/execution/evaluation.ex
```diff
@@ -0,0 +1,169 @@
...
+    expectation_evaluations
+    |> Enum.uniq_by(& &1.return_value)
+    |> Kernel.length() == 1
+  end
+
+  defp eval_expectation_result(:expect, expectations_evaluations) do
+    Enum.all?(expectations_evaluations, &(&1.return_value == true))
+  end
+
+  defp aggregate_result(%CheckResult{expectation_results: expectation_results} = check_result) do
```
arbulu89, line 142: Maybe better `aggregate_check_result` ? To clearly identify this function as something that works on an individual check

lib/wanda/execution/evaluation.ex
```diff
@@ -0,0 +1,169 @@
...
+
+  defp aggregate_execution_result(%Result{check_results: check_results} = execution_result) do
+    result =
+      check_results
+      |> Enum.map(& &1.result)
+      |> Enum.map(&{&1, result_weight(&1)})
+      |> Enum.max_by(fn {_, weight} -> weight end)
+      |> elem(0)
+
+    %Result{execution_result | result: result}
```
arbulu89, line 161: What to we store here at the end? `:passing|:warning...` or the weight? I guess the 1st one, but I don't really understand how hehe

test/execution/evaluation_test.exs
```diff
@@ -0,0 +1,174 @@
...
+    AgentCheckResult,
+    CheckResult,
+    Evaluation,
+    ExpectationEvaluation,
+    ExpectationEvaluationError,
+    ExpectationResult,
+    Result
+  }
+
+  describe "evaluation of an execution" do
```
arbulu89, line 14: Nice work with the tests. I don't want to add any overload here, but would it make sense to cover more scenarios. I only see the "passing" scenario covered XD

arbulu89 (approved):
Excellent piece of work @fabriziosestito and @CDimonaco , really fantastic.
The work on the `evaluation.ex` file is totally great.

I took my time to check the evaluation flow, and even though it is not that simple, with some effort it is easy to understand.

Maybe some  docstring explanation above about the general process would help others to understand, something like, but in any case the code is enough meaningful. So don't give to much thought to this hehe

lib/wanda/execution/evaluation.ex
fabriziosestito: this will store the `:passing | :warning ...`
you can check it here: https://github.com/trento-project/wanda/pull/4/commits/b22b534147d771dc8a479d312228d030939290f8#diff-0a5bf24059f0d65a81c7c45f97bbf5c1b712a07e66e1044fe14a85db757ebbf8

always refer to the type specs because the dialyzer catches if we are using values other than the specified

test/execution/evaluation_test.exs
fabriziosestito: ehrm we had it in mind but this morning i totally forgot to add them we can have them in a separate PR, adding a task in jira

lib/wanda/execution/evaluation.ex
fabriziosestito: thanks, makes sense

lib/wanda/execution/check_result.ex
```diff
@@ -1,21 +1,21 @@
 defmodule Wanda.Execution.CheckResult do
   @moduledoc """
-  Represents the result of check.
+  Represents the result of a a check.
```
nelsonkopliku, line 3: typo, double `a`

nelsonkopliku (approved):
LGTM

test/execution/evaluation_test.exs
fabriziosestito: @arbulu89  added moar tests

dottorblaster (approved):
LGTM

CDimonaco (approved):

PR #5, opened 2022-08-03
Setup amqp

fabriziosestito:
This PR adds a basic RabbitMQ setup leveraging `gen_rmq`.
We are configuring a single exchange of type: topic and we use a routing key for publishing messages to the agent and the dashboard.


lib/wanda/messaging/adapters/amqp.ex
```diff
@@ -5,6 +5,16 @@ defmodule Wanda.Messaging.Adapters.AMQP do
 
   @behaviour Wanda.Messaging.Adapters.Behaviour
 
+  alias Wanda.Messaging.Adapters.AMQP.Publisher
+
+  @impl true
+  def publish(routing_key, message), do: Publisher.publish_message(message, routing_key)
```
arbulu89, line 11: Isn't it a bit strange that in the 1st signature we have the `routing_key` 1st and the message 2nd, and in the second function just in the opposite order?

config/config.exs
```diff
@@ -1,6 +1,26 @@
...
+
+config :wanda, Wanda.Facts.Messenger.Supervisor,
+  children: [Wanda.Messaging.Adapters.AMQP.Publisher, Wanda.Messaging.Adapters.AMQP.Consumer]
+
+config :wanda, :messaging,
+  adapter: Wanda.Messaging.Adapters.AMQP,
+  amqp: [
+    consumer: [
+      queue: "events_wanda",
+      exchange: "events",
```
arbulu89, line 13: Why do we need an exchange in the consumer?

lib/wanda/messaging.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Wanda.Messaging do
+  @moduledoc """
+  Publishes messages to the message bus
+  """
+
+  @spec publish(String.t(), any()) :: :ok | {:error, any()}
+  def publish(topic, message), do: adapter().publish(topic, message)
```
arbulu89, line 7: Could we maybe use the `routing_key` wording here?
Rather than using `routing_key` and `topic`

arbulu89 (commented):
I don't entirely understand the code and what you wrote in the PR description.
Where is declared to which agent to send the message? 
I don't see the AgendID being used anywhere to build the routing key.

I don't even understand where is defined that the exchange is of `topic` type

lib/wanda/messaging.ex
fabriziosestito: Routing key is an internal of rabbitmq, it doesn't exist in other contexts/message brokers. The common term to define this is "topic", so from the abstract messaging implementation perspective, which doesn't know about the underlying adapter, topic sounds more fitting to me.

lib/wanda/messaging/adapters/amqp.ex
fabriziosestito: This is to mimic what phoenix channel does, which has the topic as first parameter and the message as the second. We used this in trento web so I would vote to keep the (topic, message) order in this way.

fabriziosestito:
> I don't entirely understand the code and what you wrote in the PR description.
> Where is declared to which agent to send the message?
> I don't see the AgendID being used anywhere to build the routing key.
> 
> I don't even understand where is defined that the exchange is of topic type

This is just the setup of the library, as you can see from other parts the messaging bus is not really used yet.
The RabbitMQ topic exchange seems the fitting type for this use case, we basically have one exchange and we can use routing key wildcard to consume all or part of the events. 
We still have to add the producing part where the agent routing key will be specified.

config/config.exs
fabriziosestito: Yes we need to bind the queue with the exchange if it doesn't exist the first time we start the application.

lib/wanda/messaging.ex
arbulu89: Ok. I got confused with the fact that the "rabbitmq` adapter is named "amqp", and the terminologies are a bit mixed (routing_key is used in some places and topic in other, what rabbitmq is not referred anywhere.
Maybe the adapter should be named `rabbitmq`?
Anyway, I see that you already had a lot of thinking on this, so I don't want to extend it more.
It is fine as it is.

arbulu89 (approved):
Green light from my side!
PD: I see that the merge request from `main` brought a lot of changes to PR, is this correct? Shouldn't be all this changes already removed from the diff?
PD2: I saw that the PR now is pointing the `group_expectations`. I don't know why, but well, I guess there is some reason hehe

fabriziosestito:
> Green light from my side! PD: I see that the merge request from `main` brought a lot of changes to PR, is this correct? Shouldn't be all this changes already removed from the diff? PD2: I saw that the PR now is pointing the `group_expectations`. I don't know why, but well, I guess there is some reason hehe

yes I think I need to rebase --onto, i am on it

nelsonkopliku (approved):
Great! Green light to go forward.

---

I wanted to share this thought:

> Wanda will consume all the events (`checks.#` ), the agents will consume only events matching `checks.agents.agent_id` while the dashboard will consume events matching with `checks.execution` for instance.
>

I have only one doubt about the agents consuming from `checks.agents.{agent_id}`.
If I get this correctly it means that, for a single execution, Wanda would be sending N messages to the exchange, one per agent.

`checks.agents.agent-id-1`
`checks.agents.agent-id-2`
`checks.agents.agent-id-3`

I am wondering if it would make sense that wanda, when facts gathering should start on the agents, publishes only one message (ie. `checks.gathering.{cluster_id|group_id|whatever}` ) that then all the interested agents would consume, similary to a pub sub, where the same message gets processed by many consumers :thinking: 

In the [Topic example](https://www.rabbitmq.com/tutorials/tutorial-five-elixir.html) on rabbit website there's this statement, which I found helpful.

> **Topic exchange**
> Topic exchange is powerful and can behave like other exchanges.
> 
> When a queue is bound with "#" (hash) binding key - it will receive all the messages, regardless of the routing key - like in fanout exchange.
> 
> When special characters "*" (star) and "#" (hash) aren't used in bindings, the topic exchange will behave just like a direct one.
> 

No blocker for the PR, just a discussion topic

PR #9, opened 2022-08-18
Receive Execution Requested event

fabriziosestito:
Basic setup to receive an execution requested event.
For now, events are treated as maps and validated inside Wanda.
After were switch to the contract repo, we can move the schema and cloudevents unwrapping part.


nelsonkopliku (approved):

dottorblaster (approved):
LGTM

PR #10, opened 2022-08-22
Adds json schema for emitted ExecutionCompletedV1

nelsonkopliku:
Adds the json schema for `ExecutionCompleted(V1)` (a json representation of [lib/wanda/execution/result.ex](https://github.com/trento-project/wanda/blob/main/lib/wanda/execution/result.ex)) event emitted by wanda when all the agents have gathered facts and evaluation was triggered.

Here's a bunch of json samples that would satisfy the schema https://gist.github.com/nelsonkopliku/3ff61acfe69ace5d200e0c2e21d0cd96

It can be tried out here https://www.jsonschemavalidator.net/ or here https://jsonschemalint.com/#!/version/draft-07/markup/json

Next steps:
- make sure wanda is able to codify in json the result
- produce a cloud-event enveloped version
- publish the resulting json on rabbit

### DIFF
--- a/priv/json_schema/trento.checks.v1.ExecutionCompleted.schema.json
+++ b/priv/json_schema/trento.checks.v1.ExecutionCompleted.schema.json
@@ -105,6 +105,9 @@
                 "facts": {
                   "type": "object",
                   "title": "GatheredFacts",
+                  "propertyNames": {
+                    "pattern": "^[A-Za-z0-9_][A-Za-z0-9_]*$"
+                  },
                   "patternProperties": {
                     ".*": {
                       "type": [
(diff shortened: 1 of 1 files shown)


priv/json_schema/trento.checks.v1.ExecutionCompleted.schema.json
```diff
@@ -0,0 +1,178 @@
...
+                      "type": [
+                        "number",
+                        "string",
+                        "boolean",
+                        "object",
+                        "array",
+                        "null"
+                      ]
+                    }
+                  }
```
dottorblaster, line 116: Wouldn't it be better to shape this as a proper gathered fact structure?

dottorblaster (changes requested):
Just one comment, then we are good to go I think

dottorblaster (approved):
I was wrong about the key pattern, LGTM

PR #11, opened 2022-08-23
Serialize an ExecutionCompleted json cloud event

nelsonkopliku:
This PR enables producing a json representation of `trento.checks.v1.ExecutionCompleted` as a cloud event.

It opens possibilities to support serialization for different events, yet I did not go too far considering that part of what is being done in wanda in relation to Cloud Events, would converge in trento contracts.

I had a bit of an issue with the cached dependencies, in [this previous run](https://github.com/trento-project/wanda/runs/7989424887?check_suite_focus=true) for instance, where the schema was not found, while actually being there in the code.
~~I ended up adding the branch name (`${GITHUB_REF##*/}`) as part of the cache key used for the dependencies, which gave me updated deps and a green pipeline :smile:~~ 
~~I understand it might slow down a bit because each branch would have its own cache, so reused only for subsequent runs of THAT branch and not cross-branch... at the same time I am wondering whether this might be actually helpful as it fixes cases like this :sweat_smile:~~
~~What do you think? Suggestions?~~
opened #12 in this regard


rtorrero:
@nelsonkopliku great work!

About the CI problems: Does it really make sense to use the branch name as part of the key for the cache? Won't this cause trouble again when we merge it into `main`, as `main` didn't previously implement the schema either?

nelsonkopliku:
> About the CI problems: Does it really make sense to use the branch name as part of the key for the cache? Won't this cause trouble again when we merge it into `main`, as `main` didn't previously implement the schema either?
>
thanks @rtorrero !

Not sure it will cause the same issue, as the cache for main branch as well would be regenerated.
What else am I missing? :thinking:

dottorblaster (changes requested):
Can we split this into a separate PR?

.github/workflows/ci.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
             deps
             _build/test
             priv/plts
-          key: ${{ runner.os }}-${{ env.OTP_VERSION }}-${{ env.ELIXIR_VERSION }}-${{ hashFiles('mix.lock') }}
+          key: ${{ runner.os }}-${{ env.OTP_VERSION }}-${{ env.ELIXIR_VERSION }}-${{ hashFiles('mix.lock') }}-${GITHUB_REF##*/}
```
dottorblaster, line 44: Can we split this into another PR?

lib/wanda/messaging/event_registry.ex
```diff
@@ -0,0 +1,13 @@
...
+  Cloudevents type registry of the internal representations
+  """
+
+  @execution_completed_event "trento.checks.v1.ExecutionCompleted"
+
+  @spec extract_metadata(any) :: {:ok, String.t(), String.t()} | {:error, any()}
+  def extract_metadata(%Wanda.Execution.Result{execution_id: execution_id}),
+    do: {:ok, @execution_completed_event, execution_id}
+
+  def extract_metadata(_), do: {:error, :unsupported_event}
```
dottorblaster, line 12: Can we have this moved into the Mapper module?

lib/wanda/messaging/event_registry.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Wanda.Messaging.EventsRegistry do
+  @moduledoc """
+  Cloudevents type registry of the internal representations
+  """
+
+  @execution_completed_event "trento.checks.v1.ExecutionCompleted"
+
+  @spec extract_metadata(any) :: {:ok, String.t(), String.t()} | {:error, any()}
+  def extract_metadata(%Wanda.Execution.Result{execution_id: execution_id}),
+    do: {:ok, @execution_completed_event, execution_id}
```
dottorblaster, line 10: I think this should be `{:ok, {@execution_completed_event, execution_id}}` as tagged tuples always hold two elements in my experience

lib/wanda/messaging/mapper.ex
```diff
@@ -15,4 +17,23 @@ defmodule Wanda.Messaging.Mapper do
...
+  defp build_cloud_event(id, type, event) do
+    %{
+      "specversion" => "1.0",
+      "id" => id,
+      "type" => type,
+      "source" => "https://github.com/trento-project/wanda",
+      "data" => event
+    }
+    |> CloudEvent.from_map()
+  end
```
dottorblaster, line 38: My doubt is that we end up doing the exact same work in the contract repo, are we sure we wanna move forward with this?

lib/wanda/messaging/mapper.ex
```diff
@@ -15,4 +17,23 @@ defmodule Wanda.Messaging.Mapper do
       {:ok, cloud_event}
     end
   end
+
+  @spec to_json(struct()) :: binary() | {:error, any()}
+  def to_json(event) do
```
dottorblaster, line 22: I don't think we need to return `:ok` or `:error` here, because conceptually nobody shouldn't be passing unsupported events to this function.

dottorblaster (changes requested):

.github/workflows/ci.yaml
nelsonkopliku: This PR would fail :sweat_smile: 
Maybe we can merge the _another PR_ first

What do you prefer?

lib/wanda/messaging/event_registry.ex
nelsonkopliku: Fine by me, it doesn't harm eitherway :smile:

lib/wanda/messaging/event_registry.ex
nelsonkopliku: Ok, gonna make it like that! Thanks!

.github/workflows/ci.yaml
dottorblaster: I'd prefer to split and merge the other PR, but I don't really get why we're getting into this behavior ._.

nelsonkopliku: I could observe is that the [linked broken run](https://github.com/trento-project/wanda/runs/7989424887?check_suite_focus=true) fails with a `{:error, :schema_not_found}`
```
1) test serializing outgoing events should produce an ExecutionCompletedV1 with a valid event data, as per the schema (Wanda.MapperTest)
     test/messaging/mapper_test.exs:31
     match (=) failed
     code:  assert :ok = Wanda.JsonSchema.validate(event_type, event_data)
     left:  :ok
     right: {:error, :schema_not_found}
     stacktrace:
       test/messaging/mapper_test.exs:40: (test)
```
meaning it cannot find the file `priv/json_schema/trento.checks.v1.ExecutionCompleted.schema.json` which is actually there committed and pushed.

By triggering a cache rebuild the pipeline is happy again.

This is at least _what_ happens. I am not sure about the _why_: it looks like what's in `priv/json_schema/` gets bundled in the deps cache. 
No idea :sweat_smile:

lib/wanda/messaging/mapper.ex
nelsonkopliku: I am not sure I fully understand. 
What do you mean? Could you elaborate a bit more?

dottorblaster: The caller of this function should know if an event is supported or not, hence the tagged tuple is not needed here IMHO

.github/workflows/ci.yaml
nelsonkopliku: > I'd prefer to split and merge the other PR, but I don't really get why we're getting into this behavior ._.

Here you go https://github.com/trento-project/wanda/pull/12

lib/wanda/messaging/event_registry.ex
nelsonkopliku: Changed here https://github.com/trento-project/wanda/pull/11/commits/402a03b6dc8e4ad95ba02f620bcddb5d0c3e9792

lib/wanda/messaging/event_registry.ex
nelsonkopliku: Changed here https://github.com/trento-project/wanda/pull/11/commits/402a03b6dc8e4ad95ba02f620bcddb5d0c3e9792

dottorblaster (changes requested):
LGTM

dottorblaster (approved):
LGTM (wth is wrong with this UI?)

CDimonaco (approved):

PR #13, opened 2022-08-30
Publish facts gathering initiation event

nelsonkopliku:
This PR adds the ability to react on an `ExecutionRequested` from the web, with a `FactsGatheringRequested` event that will contain information for the interested agents to actually start gathering facts.

This implementation goes in the direction of publishing only one message to all agents, and interested ones will actually do something upon consumption, while not interested ones might just ack/nack the message and do nothing.
**example routing_key**:  `checks.(execution|facts_gathering).${execution_id}`

The alternative would be sending one message per agent, each message containing the facts for a specific agent.
**example routing_key**:  `checks.agents.${agent_id}`

I see both acceptable, yet not sure which way is the best. 
This needs to be investigated a bit more, and possibly tried in integration with the agent.

I have no strong opinions, and I am open to change my mind and the code, yet this is at least a starting point.

<details>
  <summary>Here's some notes i used for reasoning on the thing</summary>
  
  ```
Each target represents a node in the cluster plus the selected checks to be executed on that node
 One-to-one relationship Agent:CheckSelection
 Different Agents may have partially or totally same selection, or differ completely

 The following is a valid scenario
 Target{agent-1, [c1, c2, c3]}
 Target{agent-2, [c1, c5]}
 Target{agent-3, [c1, c5]}
 Target{agent-5, [c6, c7, c8, c9]}

Given the checks selection we need to communicate each agent which facts to gather.
Shoot your thoughts.

pseudo-message

 execution_id uuid
 group_id uuid
 targets [agent1, agentX] ->  the consuming agent checks if its agent id is in the list, so it should react
 facts [
  AgentFacts{ -> Facts to be gathered per each agent
    agent_id: agent1, -> an interested agents extracts only its own facts to be gathered
    facts: [
      Fact{check_id-1, name-1, gatherer-1, argument-1},
      Fact{check_id-2, name-2, gatherer-2, argument-2}
    ]
  },
  AgentFacts{...}
 ]
```
</details>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ env:
   ELIXIR_VERSION: 1.13.4
   OTP_VERSION: 24
   MIX_ENV: test
-  CACHE_VERSION: 1.0
+  CACHE_VERSION: 1.1
 
 jobs:
   elixir-deps:
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -3,7 +3,7 @@ defmodule Wanda.Catalog do
   Function to interact with the checks catalog.
   """
 
-  alias Wanda.Catalog.Expectation
+  alias Wanda.Catalog.{Expectation, Fact}
 
   @catalog_path Application.compile_env!(:wanda, [__MODULE__, :catalog_path])
 
@@ -22,6 +22,17 @@ defmodule Wanda.Catalog do
     check_id
     |> get_expectations()
     |> Enum.find(&(&1.name == name))
+  end
+
+  @doc """
+  Get a list of facts for a given check.
+  """
+  @spec get_facts(String.t()) :: [Fact.t()]
+  def get_facts(check_id) do
+    check_id
+    |> get_check()
+    |> Map.fetch!("facts")
+    |> Enum.map(&map_fact(check_id, &1))
   end
 
   defp get_check(check_id) do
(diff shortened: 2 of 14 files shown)


arbulu89 (commented):
Hey @nelsonkopliku ,
Letting the code aside...

I find the send 1 message to all agents architecture more "complex" (in eyes to understand and follow the whole flow of messages) than the 1 message per agent flow.

This is why I think this:
- When you send 1 message per agent the message payload is simpler. There is not need to add information about other agents, or even the agent id. 
- The agent needs to read, parse and decide if the message is for it or not. This already adds some complexity in the agent side code. Otherwise, the complexity would go entirely to the queue system (and a bit to wanda, which would need to publish 1 message per agent).
- Sending 1 message to all makes the ACK of the message more difficult. When do you remove the message from the queue? 1 agent doesn't know if the others have already received the message or not, so they cannot remove the message themselves. When you have 1 queue per agent, the agent itself is responsable of saying if the message was consumed properly and nothing else needs to be done.
- How would be handled NACKs to the queue? More if only some of the agents and not all send it. Is the message requeued? (adding the message for the other agents back again).

I understand that the routing system is ideal to already filter the messages to go to unwanted agents, so we don't need to implement this in our code.

My 2 cents hehe
It is true that I didn't research much the advantages of publishing 1 message to all, rather than that one, that you only publish 1 message from wanda hehe

nelsonkopliku:
Thanks for the feedback @arbulu89!

I need to try out the integration with the agent and make them do something together.
I'll come back :wink:

PR #16, opened 2022-09-02
One msg per agent integration exp

nelsonkopliku:


lib/wanda/execution/facts_request.ex
```diff
@@ -15,3 +15,20 @@ defmodule Wanda.Execution.FactsRequest do
           facts: [AgentFacts.t()]
         }
 end
+
+# TODO: decide between one msg or many msgs
+defmodule Wanda.Execution.AgentFactsRequest do
```
nelsonkopliku, line 20: Variant for one message per agent scenario

config/config.exs
```diff
@@ -1,17 +1,35 @@
...
+  ]
 
 config :wanda, :messaging,
   adapter: Wanda.Messaging.Adapters.AMQP,
   amqp: [
     consumer: [
       queue: "events_wanda",
       exchange: "events",
-      routing_key: "checks.*",
+      routing_key: "checks.executions",
```
nelsonkopliku, line 32: No strong opinions about the routing key.

It can be different things depending on what we want to achieve.
`checks.*` would match 
`checks.execution_requested`, `checks.facts_gathered` 

`checks.#` would match 
`checks.execution_requested.<execution_id>`, `checks.facts_gathered.<agent_id>`

sample [here](https://www.rabbitmq.com/tutorials/tutorial-five-elixir.html)

In this example a fixed `checks.executions` is just to try it out.
The fact is that wanda should be able to consume messages from the web and the agents
- web -> `ExecutionRequested`
- agents -> `FactsGathered`

lib/wanda/execution/server.ex
```diff
@@ -44,13 +44,44 @@ defmodule Wanda.Execution.Server do
           targets: targets
         } = state
       ) do
-    :ok =
+    case Wanda.Messaging.ProcessCache.get(execution_id) do
```
nelsonkopliku, line 47: The thing here is this:
I was getting many times the same message, so I thought about at least once message delivery by rabbit.

The point is: what happens if we get an `ExecutionRequested` event (for the same execution id) more than once?

I believe that we need to be able to know whether an execution is already started, start it if needed, or just ignore if already running.

I thought of a simple cache where we track started executions.

Doubts are:
- maybe I am missing some other nice way with supervisors/genservers
- if using a cache would be acceptable, I am not sure doing this operation in the `Server` is fine, another option is doing it somewhere else, let's say closer to the consumer layer
- any other thing I overlooked :eyes:

lib/wanda/messaging/adapters/amqp/consumer.ex
```diff
@@ -26,15 +26,29 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
   @impl GenRMQ.Consumer
   def handle_message(%GenRMQ.Message{payload: payload} = message) do
     case Mapper.from_json(payload) do
-      {:ok, event} -> Policy.handle_event(event)
-      {:error, reason} -> handle_error(message, reason)
+      {:ok, event} ->
```
nelsonkopliku, line 29: Was experimenting with acks, just feel free to ditch.

lib/wanda/messaging/mapper.ex
```diff
@@ -47,5 +47,11 @@ defmodule Wanda.Messaging.Mapper do
   defp extract_metadata(%FactsRequest{execution_id: execution_id}),
     do: {:ok, {@facts_gathering_requested_event, execution_id}}
 
+  defp extract_metadata(%{"execution_id" => execution_id, "group_id" => _, "targets" => _}),
+    do: {:ok, {"trento.checks.v1.ExecutionRequested", execution_id}}
```
nelsonkopliku, line 51: wanda will not emit a `ExecutionRequested`, so this should go.

I am just using wanda to pretend wanda is the web kek

priv/catalog/156F64.yaml
```diff
@@ -14,4 +14,4 @@ facts:
     argument: totem.token
 expectations:
   - name: timeout
-    expect: corosync_token_timeout == 30000
+    expect: corosync_token_timeout == "30000"
```
nelsonkopliku, line 17: This is an interesting one, when the parts clicked the check failed :sweat_smile: 

The agent returns a string `"30000"`. while the expression containing a number.

PR #12, opened 2022-08-24
Adds cache version to pipeline

nelsonkopliku:
Relates to https://github.com/trento-project/wanda/pull/11
Adds a `CACHE_VERSION` env, used in cache key, allowing a cache update by bumping the version

PR #18, opened 2022-09-06
Revert "Adds cache version to pipeline"

fabriziosestito:
Reverts trento-project/wanda#12

We should never change the CI, which is standard.

The problem is the JSON schema macro that doesn't realize that a new file is added and does not trigger the recompilation.
Since the contract repo supersedes the current macro, we could remove it instead of fixing it. (Coming in a follow up PR)


dottorblaster (approved):
LGTM

rtorrero (approved):
LGTM

PR #20, opened 2022-09-13
Remove JSON schema, add new protobuf contracts 

fabriziosestito:
Changes JSON schema related code to the newly generated contracts from protobuf.
Refactors inbound events to use the new contracts code.

Refactors `Result.t()` to use `Fact.t()` instead of a map for clarity and mapping reasons.

Follow up PR will take care of outbound events.


lib/wanda/execution/evaluation.ex
```diff
@@ -71,6 +74,16 @@ defmodule Wanda.Execution.Evaluation do
     }
   end
 
+  defp add_facts(%AgentCheckResult{} = agent_check_result, facts, check_id) do
```
arbulu89, line 77: Why do you need to receive the `check_id` as an independent parameter?
The `check_id` already comes in the facts result together with the name and value

lib/wanda/policy.ex
```diff
@@ -3,43 +3,44 @@ defmodule Wanda.Policy do
...
+      Enum.map(targets, fn %{agent_id: agent_id, checks: checks} ->
         %Target{agent_id: agent_id, checks: checks}
       end)
     )
   end
 
-  def handle_event(%CloudEvent{
-        type: @facts_gathered_event,
-        data: %{"execution_id" => execution_id, "agent_id" => agent_id, "facts" => facts}
+  def handle_event(%FactsGathered{
```
arbulu89, line 31: I guess that the case where the value is a `FactError` is not handled yet, right?

lib/wanda/policy.ex
```diff
@@ -3,43 +3,44 @@ defmodule Wanda.Policy do
...
+        execution_id: execution_id,
+        agent_id: agent_id,
+        facts_gathered: facts_gathered
       }) do
     execution_impl().receive_facts(
       execution_id,
       agent_id,
-      Enum.map(facts, fn %{"check_id" => check_id, "name" => name, "value" => value} ->
+      Enum.map(facts_gathered, fn %{check_id: check_id, name: name, value: value} ->
```
arbulu89, line 39: It looks like the usage of `oneof` in elixir is much simpler than in golang, as I see that you can simply use the `value` field, without casting the the different options. It looks super nice.

arbulu89 (approved):
Great!

dottorblaster (approved):
LGTM

CDimonaco (approved):
Hello Proto

Lgtm

lib/wanda/policy.ex
fabriziosestito: when building the event you need to pass a tuple `{:type, struct}`, but when getting the value back you get the pure struct it seems

lib/wanda/execution/evaluation.ex
fabriziosestito: The facts variable here is a hashmap that we use internally in gathering business logic as a way to organize data to be accessible in an easy way.
It does not have any reference to the `check_id` field. In this function, we rebuild the original struct to send it over alongside the results for debugging/ux purposes.

rtorrero (approved):
lgtm

PR #21, opened 2022-09-14
Timeout business logic implementation

dottorblaster:
~Still have to write tests, but~ this PR implements the business logic of the timeout handling.

### DIFF
--- a/lib/wanda/execution/agent_check_error.ex
+++ b/lib/wanda/execution/agent_check_error.ex
@@ -1,0 +1,18 @@
+defmodule Wanda.Execution.AgentCheckError do
+  @moduledoc """
+  Represents the result of a check on a specific agent.
+  """
+
+  @derive Jason.Encoder
+  defstruct [
+    :agent_id,
+    :error,
+    :message
+  ]
+
+  @type t :: %__MODULE__{
+          agent_id: String.t(),
+          error: atom(),
+          message: String.t()
+        }
+end
--- a/lib/wanda/execution/agent_check_error.ex
+++ b/lib/wanda/execution/agent_check_error.ex
@@ -6,13 +6,13 @@ defmodule Wanda.Execution.AgentCheckError do
   @derive Jason.Encoder
   defstruct [
     :agent_id,
-    :error,
+    :type,
     :message
   ]
 
   @type t :: %__MODULE__{
           agent_id: String.t(),
-          error: atom(),
+          type: :timeout,
           message: String.t()
         }
 end
--- a/lib/wanda/execution/evaluation.ex
+++ b/lib/wanda/execution/evaluation.ex
@@ -123,8 +123,12 @@ defmodule Wanda.Execution.Evaluation do
        ) do
     expectation_results =
       agents_check_results
-      |> Enum.flat_map(fn %AgentCheckResult{expectation_evaluations: expectation_evaluations} ->
-        expectation_evaluations
+      |> Enum.flat_map(fn
+        %AgentCheckError{} ->
+          []
+
+        %AgentCheckResult{expectation_evaluations: expectation_evaluations} ->
+          expectation_evaluations
       end)
       |> Enum.group_by(& &1.name)
       |> Enum.map(fn {name, expectation_evaluations} ->
@@ -170,10 +174,14 @@ defmodule Wanda.Execution.Evaluation do
   end
 
   defp aggregate_check_result(
-         %CheckResult{expectation_results: expectation_results} = check_result
+         %CheckResult{
+           expectation_results: expectation_results,
+           agents_check_results: agents_check_results
+         } = check_result
        ) do
     result =
-      if Enum.all?(expectation_results, &(&1.result == true)) do
+      if Enum.all?(expectation_results, &(&1.result == true)) and
+           not errors?(agents_check_results) do
         :passing
       else
         :critical
@@ -192,6 +200,13 @@ defmodule Wanda.Execution.Evaluation do
 
     %Result{execution_result | result: result}
   end
+
+  defp errors?(agents_check_results),
+    do:
+      Enum.any?(agents_check_results, fn
+        %AgentCheckError{} -> true
+        _ -> false
+      end)
 
   # TODO: is unknown needed?
   # defp result_weight(:unknown), do: 3
--- a/lib/wanda/execution/evaluation.ex
+++ b/lib/wanda/execution/evaluation.ex
@@ -57,7 +57,11 @@ defmodule Wanda.Execution.Evaluation do
     agents_results =
       Enum.map(agents_facts, fn
         {agent_id, :timeout} ->
-          %AgentCheckError{agent_id: agent_id, error: :timeout, message: "Timeout dude"}
+          %AgentCheckError{
+            agent_id: agent_id,
+            type: :timeout,
+            message: "Agent timed out during the execution"
+          }
 
         {agent_id, facts} ->
           %AgentCheckResult{agent_id: agent_id}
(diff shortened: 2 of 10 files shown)


lib/wanda/execution/agent_check_error.ex
```diff
@@ -0,0 +1,18 @@
...
+  @derive Jason.Encoder
+  defstruct [
+    :agent_id,
+    :error,
+    :message
+  ]
+
+  @type t :: %__MODULE__{
+          agent_id: String.t(),
+          error: atom(),
```
fabriziosestito, line 15: `type:  :timeout`

lib/wanda/execution/server.ex
```diff
@@ -73,8 +73,22 @@ defmodule Wanda.Execution.Server do
...
-        %State{} = state
+        %State{
+          execution_id: execution_id,
+          group_id: group_id,
+          gathered_facts: gathered_facts,
+          targets: targets,
+          agents_gathered: agents_gathered
+        } = state
       ) do
+    timed_out_agents =
```
fabriziosestito, line 84: we should refer to this as "targets'

lib/wanda/execution/evaluation.ex
```diff
@@ -119,8 +119,12 @@ defmodule Wanda.Execution.Evaluation do
        ) do
     expectation_results =
       agents_check_results
-      |> Enum.flat_map(fn %AgentCheckResult{expectation_evaluations: expectation_evaluations} ->
-        expectation_evaluations
+      |> Enum.flat_map(fn
```
fabriziosestito, line 122: imho being explicit here is better than being implicit, I would `Enum.filter` `AgentCheckError` and then keep the flat map as it was

dottorblaster: :+1: done

lib/wanda/execution/server.ex
```diff
@@ -80,9 +80,24 @@ defmodule Wanda.Execution.Server do
...
+        } = state
       ) do
-    :ok = Messaging.publish("results", :timeout)
+    targets =
+      Enum.filter(targets, fn %Target{agent_id: agent_id} ->
+        agent_id not in agents_gathered
+      end)
+
+    gathered_facts = Gathering.put_gathering_timeouts(gathered_facts, targets)
+    result = Evaluation.execute(execution_id, group_id, checks, gathered_facts)
```
arbulu89, line 98: This will most probably fail. We fixed this using:
`execution_completed = Messaging.Mapper.to_execution_completed(result)` and publishing that variable

arbulu89 (commented):
Just to open a discussion.
It makes me wonder if we should send a different message in this situation. We already can send different messages with protobuf, and in this case is strange that we need to fill all the expected checks with timeout, instead of sending something like `ExecutionTimeout` struct.

At the end, this timeout scenario is pretty specific -> we didn't get the facts so we couldn't run the checks. It is something generic, not check specific.

My 2 cents

fabriziosestito:
> Just to open a discussion. It makes me wonder if we should send a different message in this situation. We already can send different messages with protobuf, and in this case is strange that we need to fill all the expected checks with timeout, instead of sending something like `ExecutionTimeout` struct.


> At the end, this timeout scenario is pretty specific -> we didn't get the facts so we couldn't run the checks. It is something generic, not check-specific.
> 
> My 2 cents


@arbulu89  the idea was to have a best-effort approach to the execution. `ExecutionTimeout` would not allow us to send an event which is partial.
We thought about sending a list of the agents who timed out, in a separate field, or even a `ExecutionCompletedWithTimeouts` event, maybe we can follow up to this.

lib/wanda/execution/server.ex
dottorblaster: Ah yeah you're right, my bad

lib/wanda/execution/evaluation.ex
```diff
@@ -54,10 +55,14 @@ defmodule Wanda.Execution.Evaluation do
          agents_facts
        ) do
     agents_results =
-      Enum.map(agents_facts, fn {agent_id, facts} ->
-        %AgentCheckResult{agent_id: agent_id}
-        |> add_agent_expectation_result(facts, expectations)
-        |> add_facts(facts, check_id)
+      Enum.map(agents_facts, fn
+        {agent_id, :timeout} ->
+          %AgentCheckError{agent_id: agent_id, type: :timeout, message: "Timeout dude"}
```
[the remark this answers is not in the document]
dottorblaster: Actually half a leftover half an "I'm glad you noticed it so you can give me an advice about a better string" :smile:

Do you have any better idea? I really didn't know what to put in there :sweat_smile:

lib/wanda/execution/server.ex
```diff
@@ -80,9 +80,24 @@ defmodule Wanda.Execution.Server do
...
+          agents_gathered: agents_gathered
+        } = state
       ) do
-    :ok = Messaging.publish("results", :timeout)
+    targets =
+      Enum.filter(targets, fn %Target{agent_id: agent_id} ->
+        agent_id not in agents_gathered
+      end)
+
+    gathered_facts = Gathering.put_gathering_timeouts(gathered_facts, targets)
```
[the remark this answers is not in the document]
dottorblaster: Well actually here we did, because maybe we have some gathered facts and some timeouts. But I'm ok changing it

dottorblaster:
@arbulu89 I think @fabriziosestito told you the big picture already :sweat_smile: yeah I'm absolutely open to add an outer field in a follow-up. Also this kind of infrastructure allows us to set up the field for more cases in which we have an error, and reuse basically the same pattern and the same functions if we find more corner cases. Thanks for the feedback!

test/execution/server_test.exs
```diff
@@ -142,5 +142,30 @@ defmodule Wanda.Execution.ServerTest do
 
       assert_receive {:DOWN, ^ref, _, ^pid, :normal}
     end
+
+    test "should go down when the timeout function gets called" do
```
fabriziosestito, line 146: here we should test that the facts were partially gathered by sending some of the agents `receive_facts`

test/execution/server_test.exs
```diff
@@ -142,5 +142,30 @@ defmodule Wanda.Execution.ServerTest do
...
+             config: [timeout: 99_999]
+           ]}
+        )
+
+      ref = Process.monitor(pid)
+
+      Process.send(pid, :timeout, [:noconnect])
+
+      assert_receive {:DOWN, ^ref, _, ^pid, :normal}
+    end
```
fabriziosestito, line 169: a test that proves that the timeout is invoked correctly after the timeout period is missing

fabriziosestito (changes requested):

lib/wanda/execution/gathering.ex
```diff
@@ -19,6 +19,19 @@ defmodule Wanda.Execution.Gathering do
     )
   end
 
+  @doc """
+  Adds timeout data to gathered facts.
+  """
+  @spec put_gathering_timeouts(map(), [Target.t()]) :: map()
+  def put_gathering_timeouts(gathered_facts, timed_out_agents) do
```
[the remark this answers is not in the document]
fabriziosestito: @arbulu89 @dottorblaster sorry guys this should be called `gathered_facts`, as it is the map that represents all the `gathered_facts`  in the GenServer. Also, the function above uses the same naming and we use the same naming in the Execution server to distinguish facts that have been gathered from facts in the catalog.

Could you please revert to the original naming?

lib/wanda/execution/server.ex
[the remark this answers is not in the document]
fabriziosestito: same as above, the naming is correct

lib/wanda/execution/evaluation.ex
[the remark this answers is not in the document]
fabriziosestito: @dottorblaster  "Agent timed out during the execution"

fabriziosestito (approved):
only one thing to change then green light from my side

lib/wanda/execution/evaluation.ex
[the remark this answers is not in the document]
arbulu89: Yes, that message is better hehe

arbulu89 (approved):
Looks good!

PR #23, opened 2022-09-15
Add checks to execution server state

fabriziosestito:
Refactors execution server and evaluation functional core so we don't do any side effects to load the catalog.
Checks are preloaded and passed to the server when it is started by the `DynamicSupervisor`, if something bad happens or if the targets are asking checks that do not exist this is where we will do the error handling.

### DIFF
--- a/lib/wanda/execution/evaluation.ex
+++ b/lib/wanda/execution/evaluation.ex
@@ -3,7 +3,7 @@ defmodule Wanda.Execution.Evaluation do
   Evaluation functional core.
   """
 
-  alias Wanda.Catalog
+  alias Wanda.Catalog.{Check, Expectation}
 
   alias Wanda.Execution.{
     AgentCheckResult,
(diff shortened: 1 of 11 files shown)


lib/wanda/execution/evaluation.ex
```diff
@@ -114,11 +119,9 @@ defmodule Wanda.Execution.Evaluation do
       end)
       |> Enum.group_by(& &1.name)
       |> Enum.map(fn {name, expectation_evaluations} ->
-        %Catalog.Check{expectations: expectations} = Catalog.get_check(check_id)
```
arbulu89, line 117: I understand that all of this code is simply to remove the need to run this `Catalog.get_check`, right?
This way all the checks execution flow depends on the coming parameters and it doesn't get anything during the flow.
Great

arbulu89 (approved):
A lot of code for a simple change, but hey, it was needed hehe
so pure now XD

lib/wanda/execution/evaluation.ex
fabriziosestito: Yes as `Catalog.get_check` reads a file now and in the future will probably retrieve checks from a persistence layer.

PR #24, opened 2022-09-15
Publish facts gathering requested

fabriziosestito:
Publish FactsGatheringRequested event

### DIFF
--- a/lib/wanda/messaging/adapters/amqp.ex
+++ b/lib/wanda/messaging/adapters/amqp.ex
@@ -14,7 +14,7 @@ defmodule Wanda.Messaging.Adapters.AMQP do
 
   def publish(routing_key, message) do
     message
-    |> Trento.Contracts.to_event(source: "wanda")
+    |> Trento.Contracts.to_event(source: "github.com/trento-project/wanda")
     |> Publisher.publish_message(routing_key)
   end
 end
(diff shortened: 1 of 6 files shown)


rtorrero (approved):
LGTM

lib/wanda/messaging/adapters/amqp.ex
```diff
@@ -8,7 +8,13 @@ defmodule Wanda.Messaging.Adapters.AMQP do
   alias Wanda.Messaging.Adapters.AMQP.Publisher
 
   @impl true
+
+  # FIXME: fix Trento.Contracts.to_event/2
```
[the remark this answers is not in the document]
fabriziosestito: the dialyzer complains about a function that will not return in the contracts repo, which depends on the `protobuf` library. Unfortunately, this bubbles up here, but the dialyzer is wrong (wrong spec in the probouf code?, not sure)

lib/wanda/messaging/adapters/amqp.ex
```diff
@@ -8,7 +8,13 @@ defmodule Wanda.Messaging.Adapters.AMQP do
...
+  @dialyzer {:nowarn_function, publish: 2}
+
   def publish(routing_key, message) do
-    Publisher.publish_message(message, routing_key)
+    message
+    |> Trento.Contracts.to_event(source: "wanda")
```
arbulu89, line 17: Maybe the source could be `github.com/trento-project/wanda`

arbulu89 (approved):
Greaaaat!

lib/wanda/messaging/adapters/amqp.ex
fabriziosestito: good one, we could take this as inspiration: https://github.com/googleapis/google-cloudevents/blob/main/examples/structured/pubsub/MessagePublishedData.json

PR #25, opened 2022-09-16
Fix amqp message consumption

arbulu89:
Fix amqp messages consumption for fact gathered event

### DIFF
--- a/lib/wanda/execution.ex
+++ b/lib/wanda/execution.ex
@@ -17,15 +17,21 @@ defmodule Wanda.Execution do
       |> Enum.uniq()
       |> Enum.map(&Catalog.get_check/1)
 
-    DynamicSupervisor.start_child(
-      Supervisor,
-      {Server,
-       execution_id: execution_id,
-       group_id: group_id,
-       targets: targets,
-       checks: checks,
-       config: config}
-    )
+    case DynamicSupervisor.start_child(
+           Supervisor,
+           {Server,
+            execution_id: execution_id,
+            group_id: group_id,
+            targets: targets,
+            checks: checks,
+            config: config}
+         ) do
+      {:ok, _} ->
+        :ok
+
+      {:error, _} = error ->
+        error
+    end
   end
 
   @impl true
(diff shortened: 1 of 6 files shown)


lib/wanda/messaging/adapters/amqp/consumer.ex
```diff
@@ -25,10 +25,10 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
 
   @impl GenRMQ.Consumer
   def handle_message(%GenRMQ.Message{payload: payload} = message) do
-    case Contracts.from_event(payload) do
-      {:ok, event} ->
-        Policy.handle_event(event)
-
+    with {:ok, event} <- Contracts.from_event(payload),
```
arbulu89, line 28: Some doubts with this `with` XD

fabriziosestito (approved):

lib/wanda/execution.ex
```diff
@@ -17,15 +17,21 @@ defmodule Wanda.Execution do
...
+            group_id: group_id,
+            targets: targets,
+            checks: checks,
+            config: config}
+         ) do
+      {:ok, _} ->
+        :ok
+
+      {:error, _} = error ->
+        error
```
dottorblaster, line 33: Why weren't we happy with the previous one?

dottorblaster (approved):
LGTM

lib/wanda/execution.ex
fabriziosestito: execution behaviour is defined to return :ok or error, returning a pid is an implementation detail of this specific implementation. we are just interested to know that the execution was started properly

lib/wanda/messaging/adapters/amqp/consumer.ex
fabriziosestito: that's ok, someone does not like using else with `with` blocks but I find it too strict

dottorblaster: `someone` is me but I think it's ok, no big deal

fabriziosestito: it depends. if you have several `else` most probably you are better using other constructs, here we have a default error case that it does not just return but needs to call an error handling function

lib/wanda/execution.ex
arbulu89: Glad to see that I had the same solution mind XD

arbulu89 (commented):
Thanks @fabriziosestito

PR #26, opened 2022-09-16
Do not requeue amqp message on error

arbulu89:
Change the `requeue` option to not do it. Otherwise, the messages are read in a loop constantly. At this point, we could publish in a dead letters queue or something.
Besides that, I have changed the logging line as it wasn't logging anything. To make this work the `message` atom should be whitelisted in the logger configuration in the `config.ex` file.

### DIFF
--- a/lib/wanda/messaging/adapters/amqp/consumer.ex
+++ b/lib/wanda/messaging/adapters/amqp/consumer.ex
@@ -36,9 +36,9 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
 
   @impl GenRMQ.Consumer
   def handle_error(message, _reason) do
-    Logger.error("Unable to handle message", message: inspect(message))
+    Logger.error("Unable to handle message: #{inspect(message)}")
 
-    GenRMQ.Consumer.reject(message, true)
+    GenRMQ.Consumer.reject(message)
   end
 
   def child_spec(opts) do


dottorblaster:
> To make this work the message atom should be whitelisted in the logger configuration in the config.ex file
I would be fine doing that, @fabriziosestito what do you think?

lib/wanda/messaging/adapters/amqp/consumer.ex
```diff
@@ -36,9 +36,9 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
 
   @impl GenRMQ.Consumer
   def handle_error(message, _reason) do
-    Logger.error("Unable to handle message", message: inspect(message))
+    Logger.error("Unable to handle message: #{inspect(message)}")
 
-    GenRMQ.Consumer.reject(message, true)
+    GenRMQ.Consumer.reject(message, false)
```
fabriziosestito, line 41: ```elixir GenRMQ.Consumer.reject(message)``` as the default is `false` https://hexdocs.pm/gen_rmq/GenRMQ.Consumer.html#reject/2

fabriziosestito (changes requested):

arbulu89:
> > To make this work the message atom should be whitelisted in the logger configuration in the config.ex file
> 
> I would be fine doing that, @fabriziosestito what do you think?

With the metadata like printing, the metadata elements are printed before the message itself, so it looks a pretty strange scenario for this use case at least.

fabriziosestito:
> > > To make this work the message atom should be whitelisted in the logger configuration in the config.ex file
> > 
> > 
> > I would be fine doing that, @fabriziosestito what do you think?
> 
> With the metadata like printing, the metadata elements are printed before the message itself, so it looks a pretty strange scenario for this use case at least. They are usually used to print return codes and things like that (as far as I have seen)

it's ok also to inspect, usually metadata are used when sending the log outside (kibana etc)

fabriziosestito (approved):

PR #27, opened 2022-09-19
Map execution completed

fabriziosestito:
Maps domain structs to `ExecutionCompleted` event


lib/wanda/messaging/mapper.ex
```diff
@@ -25,30 +29,123 @@ defmodule Wanda.Messaging.Mapper do
...
+         type: type
+       }) do
+    ExpectationResult.new!(
+      name: name,
+      result: result,
+      type: to_expectation_type(type)
+    )
+  end
+
+  defp to_result(:passing), do: :PASSING
```
arbulu89, line 140: How much I love protobuf XD

lib/wanda/messaging/mapper.ex
```diff
@@ -25,30 +29,123 @@ defmodule Wanda.Messaging.Mapper do
...
+      agent_id: agent_id,
+      facts: Enum.map(facts, &to_fact/1),
+      evaluations: Enum.map(expectation_evaluations, &to_expectation_evaluation/1)
+    )
+  end
+
+  defp to_fact(%Execution.Fact{} = fact) do
+    fact
+    |> Map.from_struct()
+    |> Fact.new!()
```
arbulu89, line 95: I think that this wouldn't work when the struct is encoded to protobuf. It would throw this error:
```
Got error when encoding Trento.Checks.V1.AgentCheckResult#facts: ** (Protobuf.EncodeError) Trento.Checks.V1.Fact#value should be {key, val},
```

And we would need to map as:
```
Fact.new!(check_id: check_id, name: name, value: {:numeric_value, value})
```
(checking the type of the value, to know if we need a numeric/string/boolean, etc).

This is my experience when I tried to make it work at least

In order to make the test most robust, we would need to try to encode the output of the mapper,

lib/wanda/messaging/mapper.ex
```diff
@@ -25,30 +29,123 @@ defmodule Wanda.Messaging.Mapper do
...
+  end
+
+  defp to_expectation_evaluation(%Execution.ExpectationEvaluation{
+         name: name,
+         return_value: return_value,
+         type: type
+       }) do
+    evaluation_value =
+      ExpectationEvaluation.new!(
```
arbulu89, line 104: I think this wouldn't work neither. We would need something like:
```
ExpectationEvaluations.new!(
                  evaluations: {
                    :evaluation_value,
                    ExpectationEvaluation.new!(
                      name: name,
                      return_value: {:boolean_value, return_value},
                      type:to_expectation_type(type)
                    )
                  }
                )
```
For the same reason as above. The encoding to the protobuf message would fail.

arbulu89 (commented):
Hey @fabriziosestito ,
I think some of the conversions would raise a failure once the struct is encoded to the protobuf event. This was my experience when I tried to make it work at least. Unless there is something I'm missing hehe (which is possible).
I would suggest to create a test with the `Trento.Contracts.to_event()` function using the created struct.

arbulu89:
> @arbulu89 I see thanks! I modified the test to encode and then instantly decode it and, in fact, it raises. I think we should have a conversation about where to store the check results before proceeding further with the encoding, so maybe the mapping of the nested structs,/ is not even needed. Moving this to draft.

Perfect! I would be really happy if we remove as many as `oneof` possibles hehe

fabriziosestito:
Superseded by architectural changes

PR #30, opened 2022-09-26
Handle fact gathering errors

arbulu89:
Add fact gathering error handling. In this case, the fact will come with the `:error_value` key, together with a type and message.
If there is any fact gathering error for a certain check, the result will come as a `AgentCheckError`.

There is a possibility to simply let Abacus deal with it, as it won't get any expected value, the expectation will simply fail. But I didn't want to attach the logic to abacus, so the result is simply set to a failure if any fact gathered for a certain check fails.


lib/wanda/execution/evaluation.ex
```diff
@@ -66,39 +66,59 @@ defmodule Wanda.Execution.Evaluation do
...
+  defp add_facts(%AgentCheckError{} = agent_check_result, facts, check_id) do
+    %AgentCheckError{
+      agent_check_result
+      | facts: build_facts(check_id, facts)
     }
   end
 
+  defp build_facts(check_id, facts) do
+    Enum.map(facts, fn
+      {name, %{type: type, message: message}} ->
```
fabriziosestito, line 114: I think the map should be populated with `{:error, fact_error}` or `{:ok, fact}` this way you can easily match without having to match the shape of the map that is a bit unclear

fabriziosestito (changes requested):

fabriziosestito:
Great work! just a comment about the shape of the gathered facts state values.

memo to self: activate credo pipe rule

lib/wanda/execution/evaluation.ex
fabriziosestito: the same goes for the `:timeout` condition

test/execution/evaluation_test.exs
```diff
@@ -160,6 +161,82 @@ defmodule Wanda.Execution.EvaluationTest do
...
+                       agent_id: "agent_2",
+                       expectation_evaluations: [
+                         %ExpectationEvaluation{
+                           name: "timeout",
+                           return_value: false,
+                           type: :expect
+                         }
+                       ],
+                       facts: [
+                         %FactError{
```
fabriziosestito, line 209: nice!

lib/wanda/execution/evaluation.ex
arbulu89: Does it make sense to include an `ok, error` distinction if they are already different struct types?
It wouldn't be better simply adding the whole data struct so we can map them by the type?
Let me send a draft commit to show what I mean hehe

arbulu89: @fabriziosestito Here the proposed change: https://github.com/trento-project/wanda/pull/30/commits/af69d60899ff6c643c109ae1156c9eb2a6564b70

At the end, I have changed how the fact gathering map is composed (that was pretty much attached on how Abacus works). No, the maps stores a list of Fact/FactError so we don't need to transform them back again. 

I didn't include any change in the `AgentCheckResult/AgentCheckError` usage in the `:timeout` scenario. I was feeling that the change was going to far away from the initial purpose of the PR. And in fact, the current logic doesn't look that bad, as we let the evaluation code handle all the final data creation

fabriziosestito: no worries, if we feel to touch the timeout logic we will in the future.
i like this logic better, it seems way more readable. big LGTM

fabriziosestito (approved):

nelsonkopliku (approved):

PR #31, opened 2022-09-27
Add amqp consumer integration tests

fabriziosestito:


test/messaging/adapters/amqp/consumer_test.exs
```diff
@@ -0,0 +1,81 @@
...
+  alias Trento.Checks.V1.{
+    ExecutionRequested,
+    FactsGathered
+  }
+
+  alias Wanda.Messaging.Adapters.AMQP.Publisher
+
+  setup [:set_mox_from_context, :verify_on_exit!]
+
+  @moduletag :integration
```
arbulu89, line 15: How does this `:integration` tag work?

arbulu89 (approved):
Hey @fabriziosestito ,
The code looks good. I have a question about how it is executed.
I don't see any change in the CI, on the `mix test` command. 
How do we run the integration tests? Or better asked, how do we run the tests without integration tests? We will want to continue running the tests without starting a rabbitmq service

test/messaging/adapters/amqp/consumer_test.exs
fabriziosestito: you can skip integration tests in local with `mix test --exclude integration`

fabriziosestito:
> Hey @fabriziosestito , The code looks good. I have a question about how it is executed. I don't see any change in the CI, on the `mix test` command. How do we run the integration tests? Or better asked, how do we run the tests without integration tests? We will want to continue running the tests without starting a rabbitmq service

@arbulu89 the I've added rabbitmq to the gh action, not sure I am understanding correctly what you are referring to.

arbulu89:
> @arbulu89 the I've added rabbitmq to the gh action, not sure I am understanding correctly what you are referring to.

Yes, I saw that. I thought that the integration tests should be enabled somehow, but as you commented above, it works in the other way around, you need to disable them if you don't want them.
Clear now!

CDimonaco (approved):

PR #32, opened 2022-09-27
Message content_type from Contracts

CDimonaco:


lib/wanda/messaging/adapters/amqp/publisher.ex
```diff
@@ -16,7 +18,7 @@ defmodule Wanda.Messaging.Adapters.AMQP.Publisher do
   def publish_message(message, routing_key \\ "") do
     Logger.info("Publishing message #{inspect(message)}")
 
-    GenRMQ.Publisher.publish(__MODULE__, message, routing_key)
+    GenRMQ.Publisher.publish(__MODULE__, message, routing_key, [{:content_type, Contracts.content_type}])
```
arbulu89, line 21: `Contracts.content_type` is a function, right? Don't you need to finish with `()`?
I think at least `Credo` should complain about

CDimonaco: You're right, the rubyst in me take over everything 🤣

CDimonaco: fixed, make ruby great again! 🤣

arbulu89 (approved):

dottorblaster (approved):
LGTM

PR #35, opened 2022-09-28
Add a storable Wanda.Result.ExecutionResult entity

nelsonkopliku:
# Description

This pr adds a `Wanda.Result.ExecutionResult` entity storable on the database.
This entity is as simple as containing:
- an `execution_id`
- a `group_id`
- a `payload` with the whole content of a `Wanda.Result.ExecutionResult`

image

image

this entity in upcoming PRs would:
- be stored when an execution completed
- read and exposed in the list API
- read and exposed in the detail API

## How was this tested?

At the moment manually by running a `mix ecto.setup` or a `mix ecto.reset`.
By doing so we should have one record in the database.

image

### DIFF
--- a/lib/wanda/execution/fact.ex
+++ b/lib/wanda/execution/fact.ex
@@ -3,6 +3,7 @@ defmodule Wanda.Execution.Fact do
   A fact is a piece of information that was gathered from a target.
   """
 
+  @derive Jason.Encoder
   defstruct [
     :check_id,
     :name,
--- a/priv/repo/migrations/20220928131105_add_execution_result.exs
+++ b/priv/repo/migrations/20220928131105_add_execution_result.exs
@@ -1,0 +1,16 @@
+defmodule Wanda.Repo.Migrations.AddExecutionResult do
+  use Ecto.Migration
+
+  def change do
+    create table(:execution_results, primary_key: false) do
+      add :execution_id, :uuid, primary_key: true
+      add :group_id, :uuid, null: false
+      add :payload, :map, null: false
+
+      timestamps()
+    end
+
+    create index(:execution_results, [:group_id])
+    create unique_index(:execution_results, [:execution_id, :group_id])
+  end
+end
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -9,3 +9,66 @@
 #
 # We recommend using the bang functions (`insert!`, `update!`
 # and so on) as they will fail if something goes wrong.
+
+execution_id = "00000000-0000-0000-0000-000000000001"
+group_id = "00000000-0000-0000-0000-000000000002"
+
+%Wanda.Results.ExecutionResult{
+  execution_id: execution_id,
+  group_id: group_id,
+  payload: %Wanda.Execution.Result{
+    execution_id: execution_id,
+    group_id: group_id,
+    check_results: [
+      %Wanda.Execution.CheckResult{
+        agents_check_results: [
+          %Wanda.Execution.AgentCheckResult{
+            agent_id: "agent_1",
+            expectation_evaluations: [
+              %Wanda.Execution.ExpectationEvaluation{
+                name: "timeout",
+                return_value: true,
+                type: :expect
+              }
+            ],
+            facts: [
+              %Wanda.Execution.Fact{
+                check_id: "expect_check",
+                name: "corosync_token_timeout",
+                value: 30_000
+              }
+            ]
+          },
+          %Wanda.Execution.AgentCheckResult{
+            agent_id: "agent_2",
+            expectation_evaluations: [
+              %Wanda.Execution.ExpectationEvaluation{
+                name: "timeout",
+                return_value: true,
+                type: :expect
+              }
+            ],
+            facts: [
+              %Wanda.Execution.Fact{
+                check_id: "expect_check",
+                name: "corosync_token_timeout",
+                value: 30_000
+              }
+            ]
+          }
+        ],
+        check_id: "expect_check",
+        expectation_results: [
+          %Wanda.Execution.ExpectationResult{
+            name: "timeout",
+            result: true,
+            type: :expect
+          }
+        ],
+        result: :passing
+      }
+    ],
+    result: :passing
+  }
+}
+|> Wanda.Repo.insert!()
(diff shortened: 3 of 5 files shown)


priv/repo/seeds.exs
```diff
@@ -9,3 +9,66 @@
 #
 # We recommend using the bang functions (`insert!`, `update!`
 # and so on) as they will fail if something goes wrong.
+
+execution_id = "fa5f4440-14ee-4169-8723-7cc7660f8024"
+group_id = "e71961b1-6c94-4b0a-84fb-7e00cc614d9d"
```
dottorblaster, line 14: Is it necessary that these are fixed? Could we just generate them?

dottorblaster (approved):
LGTM but see the comments

priv/repo/seeds.exs
arbulu89: Actually, why do we need any seed here? I would completely remove this

lib/wanda/execution/fact.ex
```diff
@@ -3,6 +3,7 @@ defmodule Wanda.Execution.Fact do
   A fact is a piece of information that was gathered from a target.
   """
 
+  @derive Jason.Encoder
```
arbulu89, line 6: `FactError` should have the json encoding as well

priv/repo/seeds.exs
dottorblaster: Well, if the choice was between this and the photofinish fixture... :smile: maybe we can remove it in a second moment? At the moment it does no harm and it allows to get started with some work that implements the read side of this

dottorblaster: I think we need seeds to test the API, right?

dottorblaster: Sorry missed your comments @nelsonkopliku, yeah maybe in the next PR you can move it in a test

priv/repo/migrations/20220928131105_add_execution_result.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Wanda.Repo.Migrations.AddExecutionResult do
+  use Ecto.Migration
+
+  def change do
+    create table(:execution_results, primary_key: false) do
+      add :execution_id, :uuid, primary_key: true
+      add :group_id, :uuid, null: false
```
arbulu89, line 7: What are these `null: false` for?

lib/wanda/result/execution_result.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Wanda.Result.ExecutionResult do
```
arbulu89, line 1: I would find other place to store this schemas. Somewhere within `execution` (as it belongs to this domain) or entities folder. Maybe `Wanda.Execution.Entities.ExecutionResult`

priv/repo/migrations/20220928131105_add_execution_result.exs
CDimonaco: It should mean that the field cannot be nullable

CDimonaco (approved):

priv/repo/seeds.exs
nelsonkopliku: > I think we need seeds to test the API, right?

Yes, it might be useful sometimes to have wanda in dev mode filled with data.

priv/repo/migrations/20220928131105_add_execution_result.exs
nelsonkopliku: Thanks Carmine, it's exactly that! At a database level.

In the following steps we will discover if we need some `validate_required` magic at the entity level.

lib/wanda/execution/fact.ex
nelsonkopliku: Thanks, done here db14ef366491aaca830f26850fcd35e5bfb289cc

lib/wanda/result/execution_result.ex
nelsonkopliku: Thanks Xabi, this is a good point.

We could indeed have an `wanda/execution/entities` directory where we could place this `execution_result.ex` entity.
Also the proposed namespacing `Wanda.Execution.Entities.ExecutionResult` sounds quite ok to me.

I'd love to have other inputs on this, though. @trento-project/developers what do you think?

fabriziosestito: We could call the folder schema since entity is ambiguous and it is not really used anywhere in the code. We could also add the context functionality to the `wanda/execution.ex `. which already has start execution receive facts and it will also have the responsibility of retrieving results.

dottorblaster: I think that having `result/something.ex` is idiomatic and super elixirish, about the name of the file I really don't know, no hard feeling

priv/repo/seeds.exs
fabriziosestito, line 14: `00000000-....1` is a valid uuid

fabriziosestito: 00000000-....1 is a valid uuid

lib/wanda/result/execution_result.ex
nelsonkopliku: Alright, before this becomes an endless thread, let's stick with the standard provided by the framework.

image

So, going with:
`wanda/results` as the directory
`wanda/results/execution_result.ex` as the schema file
`wanda/results.ex` as the context/service

Let's see what kind of refactoring it needs in the future.

priv/repo/seeds.exs
nelsonkopliku:
```elixir
execution_id = "00000000-0000-0000-0000-000000000001"
group_id = "00000000-0000-0000-0000-000000000002"
```

lib/wanda/result/execution_result.ex
arbulu89: I would go with @fabriziosestito suggestion in any case `wanda/execution/execution_result_schema.ex`. I still see that this logic should go in the `execution` folder, as it belongs there. And yes, you can put the functions to query/save the entries in `execution.ex` if you like, it looks good to me.

nelsonkopliku: keeping that in mind for the following round! Thanks!

PR #36, opened 2022-09-29
Store execution result on Check execution completion

nelsonkopliku:
# Description

This pr adds a `Wanda.Results` context exposing an `append` function to append check execution results to the history log.

The same context might be used to expose functionalities to list/search/fetch-an-item

Following up there is:
- read and exposed in the list API, with filtering
- read and exposed in the detail API

## How was this tested?

It was tested by 
- extending `test/execution/server_test.exs` and expect some items are or aren't there in the log
- adding a `test/results_test.exs` that test appending on empty/non-empty logs, and expecting specific items to be in the log

### ~~Doubts~~

~~Is the following enough to assure that if something goes wrong with storing the log item we do not publish an integration event?~~
```elixir
defp store_and_publish_execution_result(%Wanda.Execution.Result{} = result) do
    Wanda.Results.append(result)

    execution_completed = Messaging.Mapper.to_execution_completed(result)
    :ok = Messaging.publish("results", execution_completed)
  end
```
~~Not sure and not sure yet how to test that case.~~

Addressed by making sure to bang if something goes wrong during insertion `Wanda.Repo.insert!()`
```elixir
defmodule Wanda.Results do
  @spec save(Result.t()) :: ExecutionResult.t()
  def save(...) do
    %ExecutionResult{...}
    |> Wanda.Repo.insert!()
  end
end

```

### DIFF
--- a/lib/wanda/execution/execution_result_schema.ex
+++ b/lib/wanda/execution/execution_result_schema.ex
@@ -1,4 +1,4 @@
-defmodule Wanda.Results.ExecutionResult do
+defmodule Wanda.Execution.ExecutionResult do
   @moduledoc """
   Represents a storable ExecutionResult, available after check evaluation
   """
@@ -9,9 +9,9 @@ defmodule Wanda.Results.ExecutionResult do
 
   @primary_key false
   schema "execution_results" do
-    field :execution_id, Ecto.UUID, primary_key: true
-    field :group_id, Ecto.UUID
-    field :payload, :map
+    field(:execution_id, Ecto.UUID, primary_key: true)
+    field(:group_id, Ecto.UUID)
+    field(:payload, :map)
 
     timestamps(type: :utc_datetime_usec)
   end
--- a/lib/wanda/execution/result.ex
+++ b/lib/wanda/execution/result.ex
@@ -5,7 +5,7 @@ defmodule Wanda.Execution.Result do
 
   alias Wanda.Execution.CheckResult
 
-  @derive Jason.Encoder
+  @derive {Jason.Encoder, except: [:execution_id, :group_id]}
   defstruct [
     :execution_id,
     :group_id,
--- a/lib/wanda/execution/server.ex
+++ b/lib/wanda/execution/server.ex
@@ -133,7 +133,7 @@ defmodule Wanda.Execution.Server do
   end
 
   defp store_and_publish_execution_result(%Wanda.Execution.Result{} = result) do
-    Wanda.Results.append(result)
+    Wanda.Results.save_result(result)
 
     execution_completed = Messaging.Mapper.to_execution_completed(result)
     :ok = Messaging.publish("results", execution_completed)
--- a/lib/wanda/execution/server.ex
+++ b/lib/wanda/execution/server.ex
@@ -98,9 +98,8 @@ defmodule Wanda.Execution.Server do
 
     gathered_facts = Gathering.put_gathering_timeouts(gathered_facts, targets)
     result = Evaluation.execute(execution_id, group_id, checks, gathered_facts, timedout_agents)
-    execution_completed = Messaging.Mapper.to_execution_completed(result)
 
-    :ok = Messaging.publish("results", execution_completed)
+    store_and_publish_execution_result(result)
 
     {:stop, :normal, state}
   end
@@ -125,13 +124,19 @@ defmodule Wanda.Execution.Server do
     if Gathering.all_agents_sent_facts?(agents_gathered, targets) do
       result = Evaluation.execute(execution_id, group_id, checks, gathered_facts)
 
-      execution_completed = Messaging.Mapper.to_execution_completed(result)
-      :ok = Messaging.publish("results", execution_completed)
+      store_and_publish_execution_result(result)
 
       {:stop, :normal, state}
     else
       {:noreply, state}
     end
+  end
+
+  defp store_and_publish_execution_result(%Wanda.Execution.Result{} = result) do
+    Wanda.Results.append(result)
+
+    execution_completed = Messaging.Mapper.to_execution_completed(result)
+    :ok = Messaging.publish("results", execution_completed)
   end
 
   defp via_tuple(execution_id),
--- a/lib/wanda/results.ex
+++ b/lib/wanda/results.ex
@@ -1,0 +1,24 @@
+defmodule Wanda.Results do
+  @moduledoc """
+  This module exposes functionalities to interact with the historycal log of ExecutionResults.
+  It allows to:
+   - append an item to the log
+  """
+
+  alias Wanda.Execution.{ExecutionResult, Result}
+
+  @spec append(Result.t()) :: {:ok, any() } | {:error, any()}
+  def append(
+        %Result{
+          execution_id: execution_id,
+          group_id: group_id
+        } = result
+      ) do
+    %ExecutionResult{
+      execution_id: execution_id,
+      group_id: group_id,
+      payload: result
+    }
+    |> Wanda.Repo.insert()
+  end
+end
--- a/lib/wanda/results.ex
+++ b/lib/wanda/results.ex
@@ -7,7 +7,7 @@ defmodule Wanda.Results do
 
   alias Wanda.Execution.{ExecutionResult, Result}
 
-  @spec append(Result.t()) :: {:ok, any() } | {:error, any()}
+  @spec append(Result.t()) :: {:ok, any()} | {:error, any()}
   def append(
         %Result{
           execution_id: execution_id,
--- a/lib/wanda/results.ex
+++ b/lib/wanda/results.ex
@@ -5,10 +5,11 @@ defmodule Wanda.Results do
    - append an item to the log
   """
 
-  alias Wanda.Execution.{ExecutionResult, Result}
+  alias Wanda.Execution.Result
+  alias Wanda.Results.ExecutionResult
 
-  @spec append(Result.t()) :: {:ok, any()} | {:error, any()}
-  def append(
+  @spec save_result(Result.t()) :: ExecutionResult.t()
+  def save_result(
         %Result{
           execution_id: execution_id,
           group_id: group_id
@@ -19,6 +20,6 @@ defmodule Wanda.Results do
       group_id: group_id,
       payload: result
     }
-    |> Wanda.Repo.insert()
+    |> Wanda.Repo.insert!()
   end
 end
--- a/test/execution/server_test.exs
+++ b/test/execution/server_test.exs
@@ -8,7 +8,8 @@ defmodule Wanda.Execution.ServerTest do
   alias Trento.Checks.V1.FactsGatheringRequested
   alias Wanda.Catalog
 
-  alias Wanda.Execution.{ExecutionResult, Server}
+  alias Wanda.Execution.Server
+  alias Wanda.Results.ExecutionResult
 
   setup [:set_mox_from_context, :verify_on_exit!]
 
@@ -34,10 +35,8 @@ defmodule Wanda.Execution.ServerTest do
                   ]}
                )
 
-      history_log = Repo.all(ExecutionResult)
-
       assert pid == :global.whereis_name({Server, execution_id})
-      assert Enum.empty?(history_log)
+      assert ExecutionResult |> Repo.all() |> Enum.empty?()
     end
   end
 
@@ -64,10 +63,8 @@ defmodule Wanda.Execution.ServerTest do
          ]}
       )
 
-      history_log = Repo.all(ExecutionResult)
-
       assert_receive :wandalorian
-      assert Enum.empty?(history_log)
+      assert ExecutionResult |> Repo.all() |> Enum.empty?()
     end
 
     test "should exit when all facts are sent by all agents" do
@@ -113,9 +110,7 @@ defmodule Wanda.Execution.ServerTest do
       assert_receive :executed
       assert_receive {:DOWN, ^ref, _, ^pid, :normal}
 
-      history_log = Repo.all(ExecutionResult)
-      assert 1 == length(history_log)
-      assert [%ExecutionResult{execution_id: ^execution_id}] = history_log
+      assert [%ExecutionResult{execution_id: ^execution_id}] = Repo.all(ExecutionResult)
     end
 
     test "should timeout" do
@@ -153,10 +148,8 @@ defmodule Wanda.Execution.ServerTest do
 
       assert_receive {:DOWN, ^ref, _, ^pid, :normal}
 
-      history_log = Repo.all(ExecutionResult)
-      assert 1 == length(history_log)
-
-      assert [%ExecutionResult{execution_id: ^execution_id, group_id: ^group_id}] = history_log
+      assert [%ExecutionResult{execution_id: ^execution_id, group_id: ^group_id}] =
+               Repo.all(ExecutionResult)
     end
 
     test "should go down when the timeout function gets called" do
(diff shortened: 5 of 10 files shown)


lib/wanda/results.ex
```diff
@@ -0,0 +1,24 @@
...
+  def append(
+        %Result{
+          execution_id: execution_id,
+          group_id: group_id
+        } = result
+      ) do
+    %ExecutionResult{
+      execution_id: execution_id,
+      group_id: group_id,
+      payload: result
```
arbulu89, line 20: I would remove the `execution_id` and `group_id` from the payload map, otherwise they are duplicated in the main columns and the map, which looks "ugly" so to say

arbulu89: @fabriziosestito The credo option to warn the one line and one parameter pipes, not enabled yet I guess XD

lib/wanda/execution/server.ex
```diff
@@ -125,15 +124,21 @@ defmodule Wanda.Execution.Server do
...
+      store_and_publish_execution_result(result)
 
       {:stop, :normal, state}
     else
       {:noreply, state}
     end
   end
 
+  defp store_and_publish_execution_result(%Wanda.Execution.Result{} = result) do
+    Wanda.Results.append(result)
```
arbulu89, line 136: To keep consistency with the publishing error handling, maybe we should pattern match like:
`{:ok, _} = Wanda.Results.append(result)`

test/results_test.exs
```diff
@@ -0,0 +1,94 @@
...
+            %Fact{name: "corosync_token_timeout", check_id: "expect_check", value: 30_000}
+          ]
+        }
+      }
+
+      execution_id = UUID.uuid4()
+      group_id = UUID.uuid4()
+      checks = [Catalog.get_check("expect_check")]
+
+      result = Evaluation.execute(execution_id, group_id, checks, gathered_facts)
```
arbulu89, line 30: Should we depend on external functions to test? I think we should create the result object directly, to avoid false negatives if the `execute` (which involves a lot of stuff) changes

arbulu89 (dismissed):
I think we should extract the `execution_id` and `group_id` from the payload to avoid duplicates. Otherwise, we could remove the columns, as it is pretty simple to query fields in jsonb as well.

test/results_test.exs
fabriziosestito: i agree with @arbulu89 here we should not depend on any external function as this is not a wide test.
we should instead craft a result struct with a factory or similar and see if the stored result is what we expect

test/support/factory.ex
```diff
@@ -113,4 +114,15 @@ defmodule Wanda.Factory do
       result: Map.get(attrs, :result, Enum.random([:passing, :warning, :critical]))
     }
   end
+
+  def execution_result_log_item_factory(attrs) do
```
fabriziosestito, line 118: i find this names very hard to remember, to me this is an execution_result or a result. 
Consider splitting the factories if you have name clashing

lib/wanda/execution/execution_result_schema.ex
```diff
@@ -9,9 +9,9 @@ defmodule Wanda.Results.ExecutionResult do
 
   @primary_key false
   schema "execution_results" do
-    field :execution_id, Ecto.UUID, primary_key: true
-    field :group_id, Ecto.UUID
-    field :payload, :map
+    field(:execution_id, Ecto.UUID, primary_key: true)
```
fabriziosestito, line 12: something went wrong with the formatter, parenthesis should not appear here, add the ecto format dep to .formatter.exs

lib/wanda/results.ex
```diff
@@ -0,0 +1,24 @@
...
+  @moduledoc """
+  This module exposes functionalities to interact with the historycal log of ExecutionResults.
+  It allows to:
+   - append an item to the log
+  """
+
+  alias Wanda.Execution.{ExecutionResult, Result}
+
+  @spec append(Result.t()) :: {:ok, any()} | {:error, any()}
+  def append(
```
fabriziosestito, line 11: i find append a bit too much revealing of the underlying implementation detail, `save_result` would be more friendly to read in some months

test/execution/server_test.exs
```diff
@@ -33,7 +34,10 @@ defmodule Wanda.Execution.ServerTest do
                   ]}
                )
 
+      history_log = Repo.all(ExecutionResult)
```
fabriziosestito, line 37: why not `execution_results`?

test/execution/server_test.exs
```diff
@@ -105,6 +112,10 @@ defmodule Wanda.Execution.ServerTest do
 
       assert_receive :executed
       assert_receive {:DOWN, ^ref, _, ^pid, :normal}
+
+      history_log = Repo.all(ExecutionResult)
+      assert 1 == length(history_log)
```
fabriziosestito, line 117: this assert is not needed because of the pattern match above

test/execution/server_test.exs
```diff
@@ -60,7 +64,10 @@ defmodule Wanda.Execution.ServerTest do
          ]}
       )
 
+      history_log = Repo.all(ExecutionResult)
```
fabriziosestito, line 67: you could `assert [] == Repo.all(ExecutionResult)`

test/execution/server_test.exs
```diff
@@ -141,6 +152,11 @@ defmodule Wanda.Execution.ServerTest do
       assert_receive :timeout, 200
 
       assert_receive {:DOWN, ^ref, _, ^pid, :normal}
+
+      history_log = Repo.all(ExecutionResult)
+      assert 1 == length(history_log)
```
fabriziosestito, line 157: assert not needed

lib/wanda/execution/server.ex
fabriziosestito: not sure about this, as we bang in the context so it will raise an exception but it will never return an :ok tuple

test/execution/server_test.exs
```diff
@@ -168,6 +184,23 @@ defmodule Wanda.Execution.ServerTest do
       Process.send(pid, :timeout, [:noconnect])
 
       assert_receive {:DOWN, ^ref, _, ^pid, :normal}
+
+      history_log = Repo.all(ExecutionResult)
+      assert 1 == length(history_log)
```
fabriziosestito, line 189: this assert is not needed

test/results_test.exs
```diff
@@ -0,0 +1,94 @@
...
+
+      Results.append(result)
+
+      history_log = Repo.all(ExecutionResult)
+      assert 4 == length(history_log)
```
fabriziosestito, line 75: assert not needed

lib/wanda/results.ex
fabriziosestito: @arbulu89 we thought about having them as external fields because of indexing
we could remove those two fields from the payload indeed.

nelsonkopliku: Not a blocking thing to me, but hey if you prefer so :smile: 

That would mean doing a struct to map transformation of the payload and then remove the `execution_id` and `group_id`.
The doubt I have is that `Map.from_struct` is not recursive and it will transform only the top level struct to a map (which could be enough to remove the two fields).
Are you ok with this?

lib/wanda/execution/server.ex
nelsonkopliku: We are not banging in the context at the moment.

I can change the context to bang `Wanda.Repo.insert!()` and leave this line as it is.
Agreed?

test/execution/server_test.exs
rtorrero: You mean below?

rtorrero (commented):
Appart from the already commented stuff, LGTM

lib/wanda/execution/execution_result_schema.ex
nelsonkopliku: you mean this?

https://github.com/trento-project/wanda/blob/main/.formatter.exs#L3
```exs
# Used by "mix format"
[
  import_deps: [:ecto, :phoenix],
  inputs: ["*.{ex,exs}", "priv/*/seeds.exs", "{config,lib,test}/**/*.{ex,exs}"],
  subdirectories: ["priv/*/migrations"]
]
```
What am I missing?

test/execution/server_test.exs
nelsonkopliku: I tried that but credo was complaining, and it suggested me to use `Enum.empty?` :sweat_smile:

lib/wanda/results.ex
arbulu89: If the json encoders are there the `Map.from_struct` should work on nested things (this is actually how it is able to store the struct in the database).

This maybe would work:
```
transformed_result = result
|> Map.to_struct()
|> Map.drop([:execution_id, :group_id])
```

lib/wanda/execution/execution_result_schema.ex
fabriziosestito: maybe the formatter did something bad, try to remove parenthesis and format it again

test/execution/server_test.exs
fabriziosestito: yes :D

test/execution/server_test.exs
fabriziosestito: oh i see, so you cold `assert ExecutionResult |>  Repo.all() |> Enum.empty?()`

lib/wanda/results.ex
arbulu89: @nelsonkopliku , edit:
You were right, the `Map.from_struct` doesn't map the nested structs, but as you said, execution_id and group_id are in the 1st layer, so the code snippet still would do the work I think

lib/wanda/execution/server.ex
arbulu89: I just saw that the `append` method was returning the `{:ok, any}` or `{:error, any}` options, so to have consistency with the publish. No strong opinions though

test/execution/server_test.exs
nelsonkopliku: oh yeah I like that! Thanks for reminding!

test/support/factory.ex
nelsonkopliku: I see, there's different contexts involved.

The internal execution result computed by the engine and the execution result representation that gets appended to the log of execution results (reads _the one that gets saved in the database_)

However,
we can change `execution_result_log_item_factory` to `result_factory` if that sounds better to you, and keep it in the same factory.

`execution_result` would probably sound a bit clearer, unfortunately it would conflict with https://github.com/trento-project/wanda/blob/main/test/support/factory.ex#L69 and so would require splitting the factory.
For reference, something like this? https://hexdocs.pm/ex_machina/readme.html#where-to-put-your-factories

If you agree we can go lean and just rename `execution_result_log_item_factory` to `result_factory` at the moment.

lib/wanda/results.ex
nelsonkopliku: Alright added in the last commit.

test/results_test.exs
nelsonkopliku: Right, I cut it too short :sweat_smile: my bad.

Fixed in the last commit.

lib/wanda/execution/execution_result_schema.ex
nelsonkopliku: formatter went bananas :smile:

lib/wanda/results.ex
nelsonkopliku: Changed to `save` so it looks like `Wanda.Results.save()` instead of `Wanda.Results.save_result()`

Sounds good?

test/execution/server_test.exs
nelsonkopliku: adjusted tests in the last commit.

test/execution/server_test.exs
nelsonkopliku: addressed in the last commit.

lib/wanda/execution/server.ex
nelsonkopliku: Banged with `Wanda.Repo.insert!()` in `Wanda.Results.save/1`

test/support/factory.ex
dottorblaster: `result_factory` is ok to me

lib/wanda/results.ex
fabriziosestito: `save_result` is more idiomatic as the Results context could be dealing with more schemas, we could also find a better name for the context itself. See for example https://hexdocs.pm/phoenix/contexts.html#starting-with-generators

test/support/factory.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Wanda.Factory do
   @moduledoc false
 
-  use ExMachina
+  use ExMachina.Ecto, repo: Wanda.Repo
```
arbulu89, line 4: What is this `Wanda.Repo` here for? Is it needed because the struct is databas schema?

arbulu89 (commented):
@nelsonkopliku Pending the other conversations about naming and the `assert` thing in the tests (it looks like there are still some that are not needed), the remaining code looks good to me.

test/support/factory.ex
nelsonkopliku: it allows to do `insert(:something)` and indeed have stuff in the db as I understood reading [this](https://hexdocs.pm/ex_machina/readme.html#overview)

test/support/factory.ex
nelsonkopliku: changed to `result_factory`

```elixir
def result_factory(attrs) do
    ...
end

insert_list(3, :result)
```

test/support/factory.ex
arbulu89: Ok, I see. Well, we don't use it yet, but we will, so feel free to leave it.

arbulu89: Edit: I see fome `insert_list`, so yet, we use it hehe

nelsonkopliku: we use it here
https://github.com/trento-project/wanda/pull/36/files#diff-b5a8fea17b9ce629cd14125b9a03a7c9a81f206ce6b0b2d135624b1baa28cc0aR41

lib/wanda/results.ex
nelsonkopliku: Interesting piece of doc :smile: 

`Wanda.Results.save_result/1` just sounded a bit redundant, yet since it's more idiomatic I'll keep it like that.
And let's discover if the context name needs revisiting :+1:

nelsonkopliku:
@trento-project/developers all feedbacks should have been addressed :rocket:

arbulu89 (approved):
LGTM. I don't like the last change on how execution_id and group_id are removed, but the PR already got too long, so we could change afterwards if needed

test/support/factory.ex
```diff
@@ -113,4 +115,15 @@ defmodule Wanda.Factory do
       result: Map.get(attrs, :result, Enum.random([:passing, :warning, :critical]))
     }
   end
+
+  def result_factory(attrs) do
```
fabriziosestito, line 119: `execution_result_factory` returns a `Result` while `result_factory` returns an `ExecutionResult`.
Could we maybe invert the name of those two?

fabriziosestito (changes requested):

test/support/factory.ex
nelsonkopliku: There you go dc1e66a59932e2b794e029dcfcba938e1e6d6be0

fabriziosestito: LGTM :shipit:

fabriziosestito (approved):

lib/wanda/execution/result.ex
```diff
@@ -5,7 +5,7 @@ defmodule Wanda.Execution.Result do
 
   alias Wanda.Execution.CheckResult
 
-  @derive Jason.Encoder
+  @derive {Jason.Encoder, except: [:execution_id, :group_id]}
```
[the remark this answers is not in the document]
nelsonkopliku: I see your point, I had a similar thought.

At the moment, though, we do not really have different usages of a json serialized version of a `Wanda.Execution.Result`, so whatever way would be acceptable.

Let's keep an eye on this and decide later what approach is more suitable.

dottorblaster (approved):
LGTM

@nelsonkopliku mind that I think that @fabriziosestito referred to this when he told you to split the factories

https://github.com/thoughtbot/ex_machina#splitting-factories-into-separate-files

PR #37, opened 2022-09-30
Add dockerfile

fabriziosestito:
Written by @arbulu89 :
Wanda Dockerfile. The uploaded container is named `Wanda` for convenience, we can change it in the future (but I'm getting used to the name and I would keep it hehe)

It needs to downgrade the obfuscation library as it requires OTP version 23.2 (which we don't have in opensuse repos...).


CDimonaco (approved):

nelsonkopliku (approved):
:eyes:

PR #38, opened 2022-09-30
Some needed improvements to make the code runnable on prod environment

arbulu89:
Make the amqp url configurable on runtime and update how the catalog path is obtained.

This 2nd thing is needed to run the code in our docker file, as we don't copy the whole release files. We would need to copy the `/build/_build/prod/lib/wanda/priv/` folder as it is, with that exact path. The old `@catalog_path` was build on compilation time, which is less flexible. (Most probably there is some better way for doing this properly hehe)

### DIFF
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -9,16 +9,18 @@ defmodule Wanda.Catalog do
     Fact
   }
 
-  @catalog_path Application.compile_env!(:wanda, [__MODULE__, :catalog_path])
-
   @doc """
   Get a check from the catalog.
   """
   def get_check(check_id) do
-    @catalog_path
+    get_catalog_path()
     |> Path.join("#{check_id}.yaml")
     |> YamlElixir.read_from_file!()
     |> map_check()
+  end
+
+  defp get_catalog_path do
+    Application.fetch_env!(:wanda, Wanda.Catalog)[:catalog_path]
   end
 
   defp map_check(%{
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -77,4 +77,12 @@ if config_env() == :prod do
     publisher: [
       connection: amqp_url
     ]
+
+  # Update catalog path to the current application dir during runtime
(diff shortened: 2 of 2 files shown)


fabriziosestito (approved):
LGTM

nelsonkopliku (approved):

lib/wanda/catalog.ex
```diff
@@ -9,13 +9,19 @@ defmodule Wanda.Catalog do
     Fact
   }
 
-  @catalog_path Application.compile_env!(:wanda, [__MODULE__, :catalog_path])
+  @doc """
+  Get checks catalog path
+  """
+  def get_catalog_path() do
+    local_path = Application.fetch_env!(:wanda, Wanda.Catalog)[:catalog_path]
+    Application.app_dir(:wanda, local_path)
```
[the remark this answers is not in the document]
arbulu89: Exactly, it hardcodes to `/build/_build/blablabla` and for example in our dockerfile we move the binary from one place to another breaking this balue

rtorrero, line 17: Why are we now getting the full application path?

arbulu89: Read PR description please XD

rtorrero: Sorry, it wasn't immediately obvious to me. I'm guessing it has to do with the path changing for a realease build thus requiring an absolute path? Nevermind, fine by me if it works :ok_hand:

arbulu89:
Hey @fabriziosestito ,
I have update the code to make it work in production. Check: https://github.com/trento-project/wanda/pull/38/commits/c1395a37e12b723fd0bdda05637755cf07214188

fabriziosestito:
> Hey @fabriziosestito , I have update the code to make it work in production. Check: [92e9766](https://github.com/trento-project/wanda/commit/92e9766d3a6d2f9e83828858fa9813cacef8b1a9)

LGTM

fabriziosestito (approved):

PR #40, opened 2022-10-03
Serve the execution results through an endpoint

dottorblaster:
As above, we're exposing an API to look into historical execution results.


arbulu89 (commented):
So far so good! I have noted some comments.
Could we add controller tests for the new endpoint?

lib/wanda_web/api_spec.ex
```diff
@@ -0,0 +1,24 @@
...
+  def spec do
+    # Discover request/response schemas from path specs
+    OpenApiSpex.resolve_schema_modules(%OpenApi{
+      servers: [
+        # Populate the Server info from a phoenix endpoint
+        Server.from_endpoint(Endpoint)
+      ],
+      info: %Info{
+        title: "Wanda",
+        version: "1.0"
```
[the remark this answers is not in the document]
fabriziosestito: I think the api should go with it's own version, this is version 1 of the API

lib/wanda/results.ex
```diff
@@ -22,4 +25,24 @@ defmodule Wanda.Results do
     }
     |> Wanda.Repo.insert!()
   end
+
+  @doc """
+  Gets execution results from the database.
+  """
+  @spec get_execution_results(String.t(), number(), number()) :: [ExecutionResult.t()]
+  def get_execution_results(group_id, page, items_per_page \\ 10) do
```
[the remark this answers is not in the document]
fabriziosestito: +1 for list_execution_results, as it is the common Phoenix name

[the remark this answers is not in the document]
fabriziosestito: > I would change the page and items_per_page terminology. They are really frontend oriented namings.
Why not limit and offset (and I would calculate the offset in the controller code).

Not sure about this, limit and offset are really database related terms.
I like page and items per page because they are user facing names, the one you expect in a context.

> I liked this example: https://github.com/gothinkster/elixir-phoenix-realworld-example-app/blob/master/lib/real_world/blog/blog.ex#L22

we could use this idiom, I like it

lib/wanda_web/controllers/executions_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+    responses: [
+      ok: {"Check executions response", "application/json", CheckExecutionResponse}
+    ]
+
+  def list_checks_executions(conn, params) do
+    page = Map.get(params, :page, 1)
+    items_per_page = Map.get(params, :items_per_page, 10)
+    group_id = Map.get(params, :group_id)
+
+    results = Results.get_execution_results(group_id, page, items_per_page)
```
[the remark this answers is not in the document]
fabriziosestito: we have views for that, i think we can iterate adding the views in a new PR. 

https://hexdocs.pm/phoenix/views.html

lib/wanda_web/controllers/executions_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+      items_per_page: [in: :query, description: "Items per page", type: :integer, example: 20]
+    ],
+    responses: [
+      ok: {"Check executions response", "application/json", CheckExecutionResponse}
+    ]
+
+  def list_checks_executions(conn, params) do
+    page = Map.get(params, :page, 1)
+    items_per_page = Map.get(params, :items_per_page, 10)
+    group_id = Map.get(params, :group_id)
```
fabriziosestito, line 29: what happens if the user does not send any group_id?

We should straight match the group_id in the function header and then use this to generate an error in case the controller does not match:
https://hexdocs.pm/phoenix/controllers.html#action-fallback

fabriziosestito (changes requested):
Some comments to be addressed and controller tests are missing, we need to test that the return json is in the format we specified in the open API spex.

Refer to this: https://github.com/open-api-spex/open_api_spex#validate-responses

We could use factory to generate a list of results.

lib/wanda/results.ex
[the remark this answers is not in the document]
nelsonkopliku: I agree with the guys, we might find a different idiom, how would the reference provided look like?
Something like this?
- `Results.list_execution_results()` -> full list 
- `Results.list_execution_results(filter_by: [group_id: some_id])` -> filtering
- `Results.list_execution_results(page: page, items_per_page: items_per_page)` -> paginating
- `Results.list_execution_results(filter_by: [group_id: some_id], page: page, items_per_page: items_per_page)` -> filtering + paginating

lib/wanda_web/controllers/executions_controller.ex
dottorblaster: I acted under the assumption that if the user doesn't filter by group_id, the backend should just return unfiltered results. Do we wanna treat group_id as mandatory?

lib/wanda/results.ex
```diff
@@ -22,4 +25,24 @@ defmodule Wanda.Results do
     }
     |> Wanda.Repo.insert!()
   end
+
+  @doc """
+  Gets execution results from the database.
+  """
+  @spec get_execution_results(String.t(), number(), number()) :: [ExecutionResult.t()]
```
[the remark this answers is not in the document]
dottorblaster: Aye

lib/wanda_web/controllers/executions_controller.ex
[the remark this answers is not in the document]
dottorblaster: @arbulu89 I'm super glad you noticed that, yeah it's something we have to do but I'd like to do that in a follow-up PR because I was afraid this PR was getting really huge

lib/wanda_web/controllers/executions_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+
+    results = Results.get_execution_results(group_id, page, items_per_page)
+    json(conn, %{page: page, items_per_page: items_per_page, data: results})
```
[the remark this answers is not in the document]
rtorrero: Can't we take inspiration from HTTP 206 (RFC 9110)?

I like how it proposed to send data on the HTTP header about the position of the served content, see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206#examples)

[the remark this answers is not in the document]
rtorrero: Not suggesting do it on the header, just that I like the way it describes the range and total count

lib/wanda_web/controllers/executions_controller.ex
arbulu89: No, it shouldn't be mandatory. It is a query param. If it doesn't come, the `group_id` we return all the executions. Actually, I like the current implementation on this part.

lib/wanda_web/controllers/executions_controller.ex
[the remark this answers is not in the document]
arbulu89: @dottorblaster Fine for me them! We can follow up.

lib/wanda_web/controllers/executions_controller.ex
fabriziosestito: ah damn sorry, my bad, it is not mandatory

lib/wanda_web/controllers/executions_controller.ex
```diff
@@ -0,0 +1,34 @@
+defmodule WandaWeb.ExecutionsController do
+  use WandaWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias Wanda.Results
+  alias WandaWeb.Schemas.CheckExecutionResponse
+
+  plug OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true
+
+  operation :list_checks_executions,
```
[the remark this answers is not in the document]
dottorblaster: YES

lib/wanda_web/api_spec.ex
```diff
@@ -0,0 +1,24 @@
+defmodule WandaWeb.ApiSpec do
```
nelsonkopliku, line 1: In trento web we faced a similar situation in relation to apis/openapi/openapispex, see [here](https://github.com/trento-project/web/tree/main/lib/trento_web/openapi)

Let's consider If there's some valuable learning we can bring here - even on what not to do :sweat_smile:

lib/wanda_web/schemas/check_execution.ex
```diff
@@ -0,0 +1,20 @@
...
+  require OpenApiSpex
+
+  alias OpenApiSpex.Schema
+
+  OpenApiSpex.schema(%{
+    title: "Check execution",
+    description: "A single execution of checks",
+    type: :object,
+    properties: %{
+      execution_id: %Schema{type: :string, description: "Execution ID"},
```
nelsonkopliku, line 15: We could add a `format: :uuid` to `execution_id` and `group_id`, for a friendlier representation on the api doc.

lib/wanda_web/schemas/check_execution.ex
```diff
@@ -0,0 +1,20 @@
...
+      group_id: %Schema{type: :string, description: "Group ID"},
+      payload: %Schema{type: :object, description: "Payload of the check execution"}
```
nelsonkopliku, line 17: Probably you already thought of this.

What is the plan about the shape of the payload `object`?
As an API consumer, even if at the moment is just us :smile:, I would find it useful to know that shape.

lib/wanda_web/schemas/check_execution.ex
dottorblaster: I tried but I got an error, i googled a lot but I didn't find a solution that wasn't "create your own format with a huge regex"

lib/wanda_web/schemas/check_execution.ex
dottorblaster: Does it have a fixed shape? I thought it was something that can really vary

lib/wanda_web/schemas/check_execution.ex
nelsonkopliku: Not talking about validation, no need for regex :smile: 

Just really a silly thing like we do here for example https://github.com/trento-project/web/blob/main/lib/trento_web/openapi/schema/host.ex#L37

```elixir
%Schema{type: :string, description: "Execution ID", format: :uuid},
```

It would take care of giving that information in the swagger ui and it generate uuid samples.
If it is causing troubles feel free to skip.

lib/wanda_web/schemas/check_execution.ex
nelsonkopliku: Well, it does have a shape, and that shape is `Wanda.Execution.Result` without `:execution_id` and `:group_id`.

> Does it have a fixed shape? I thought it was something that can really vary

Until we change it :smile: 

The point to me is more 
_how stable do we think it is in relation to the needs?_
_What **significant** changes do we foresee to that representation in the upcoming iterations?_

However, I understand it might be a tedious job, moreover if things change so fast that updating the documentation becomes a burden, yet if that happens there might be another kind of issue to tackle (coupling for instance, but let's leave it out)

Bottom line: if not planned to add that missing piece in this or following PRs, then I'd find it wise to track debt.
How does it sound?

test/wanda_web/controllers/executions_controller_test.exs
```diff
@@ -0,0 +1,28 @@
...
+        %ExecutionResult{}
+      ] = insert_list(5, :execution_result)
+
+      json =
+        conn
+        |> get("/api/checks/executions")
+        |> json_response(200)
+
+      api_spec = ApiSpec.spec()
+      assert_schema(json, "CheckExecutionResponse", api_spec)
```
arbulu89, line 25: I think you could assert the returned json value with an expected one

lib/wanda_web/controllers/executions_controller.ex
```diff
@@ -0,0 +1,30 @@
...
+      page: [in: :query, description: "Page", type: :integer, example: 3],
+      items_per_page: [in: :query, description: "Items per page", type: :integer, example: 20]
+    ],
+    responses: [
+      ok: {"Check executions response", "application/json", CheckExecutionResponse}
+    ]
+
+  def list_checks_executions(conn, params) do
+    results = Results.list_execution_results(params)
+    json(conn, %{data: results})
```
arbulu89, line 28: Cannot we send simply the results? `json(conn, results)`

lib/wanda_web/schemas/check_execution_response.ex
```diff
@@ -0,0 +1,21 @@
...
+
+  alias OpenApiSpex.Schema
+  alias WandaWeb.Schemas.CheckExecution
+
+  OpenApiSpex.schema(%{
+    title: "CheckExecutionResponse",
+    description: "Check execution listing response",
+    type: :object,
+    properties: %{
+      page: %Schema{type: :string, description: "Current page"},
```
arbulu89, line 16: It looks like this got outdated

arbulu89 (approved):
Hey @dottorblaster ,
It looks good to me now. I have added some other minor comments. But feel free to omit them.

lib/wanda_web/schemas/check_execution_response.ex
dottorblaster: Damn you're right, thanks

lib/wanda_web/controllers/executions_controller.ex
dottorblaster: I wanted to wrap this into a map to make our life easier when we switch to views (Phoenix' DTOs)

test/wanda_web/controllers/executions_controller_test.exs
dottorblaster: Maybe some punctual assertions, the expected one I think it would be very big and with little value because the meaningful test is the one we do on the context. Here's we're testing more that the response is compliant with the spec. But valid point, I'll think about it

fabriziosestito (approved):
LGTM, please track the debt and the leftovers

lib/wanda_web/schemas/check_execution.ex
dottorblaster: Sounds good, I collected a followup task to better define that shape

PR #42, opened 2022-10-05
Use env values

arbulu89:
Load the new `env` field from the contract and expand it to the execution dynamic server.

### DIFF
--- a/lib/wanda/execution/behaviour.ex
+++ b/lib/wanda/execution/behaviour.ex
@@ -9,14 +9,14 @@ defmodule Wanda.Execution.Behaviour do
               execution_id :: String.t(),
               group_id :: String.t(),
               targets :: [Target.t()],
-              env :: %{String.t() => String.t()}
+              env :: map()
             ) :: :ok | {:error, any}
 
   @callback start_execution(
               execution_id :: String.t(),
               group_id :: String.t(),
               targets :: [Target.t()],
-              env :: %{String.t() => String.t()},
+              env :: map(),
               config :: Keyword.t()
             ) :: :ok | {:error, any}
 
--- a/lib/wanda/execution/behaviour.ex
+++ b/lib/wanda/execution/behaviour.ex
@@ -9,14 +9,14 @@ defmodule Wanda.Execution.Behaviour do
               execution_id :: String.t(),
               group_id :: String.t(),
               targets :: [Target.t()],
-              env :: map()
+              env :: %{String.t() => boolean() | number() | String.t()}
             ) :: :ok | {:error, any}
 
   @callback start_execution(
               execution_id :: String.t(),
               group_id :: String.t(),
               targets :: [Target.t()],
-              env :: map(),
+              env :: %{String.t() => boolean() | number() | String.t()},
               config :: Keyword.t()
             ) :: :ok | {:error, any}
 
--- a/lib/wanda/execution/behaviour.ex
+++ b/lib/wanda/execution/behaviour.ex
@@ -8,13 +8,15 @@ defmodule Wanda.Execution.Behaviour do
   @callback start_execution(
               execution_id :: String.t(),
               group_id :: String.t(),
-              targets :: [Target.t()]
+              targets :: [Target.t()],
+              env :: %{String.t() => String.t()}
             ) :: :ok | {:error, any}
 
   @callback start_execution(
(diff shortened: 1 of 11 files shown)


test/support/factory.ex
```diff
@@ -19,6 +19,14 @@ defmodule Wanda.Factory do
 
   alias Wanda.Results.ExecutionResult
 
+  def map_factory(attrs) do
```
arbulu89, line 22: I could very well name this `env_factory` to be more specific, as it is only used to build the `env` data

lib/wanda/policy.ex
```diff
@@ -26,14 +26,16 @@ defmodule Wanda.Policy do
...
+         env: env
        }) do
     execution_impl().start_execution(
       execution_id,
       agent_id,
       Enum.map(targets, fn %{agent_id: agent_id, checks: checks} ->
         %Target{agent_id: agent_id, checks: checks}
-      end)
+      end),
+      Map.new(env, fn {key, %{kind: {_, value}}} -> {key, value} end)
```
nelsonkopliku, line 38: Where does this `kind` come from? :thinking: 
Is it something related to the `random_env_kind` factory?

test/support/factory.ex
```diff
@@ -51,6 +51,33 @@ defmodule Wanda.Factory do
     }
   end
 
+  def env_factory(attrs) do
```
nelsonkopliku, line 54: If I am understanding it correctly `env_event_factory` is used in tests _closer_ to the integration layer where `ExecutionRequested` event is involved.
While `env_factory` is used in more inner layers.

What other differences are there between `env_factory` `env_event_factory`?
Would a single factory be enough?

nelsonkopliku (commented):
Overall looks good and goes in a reasonable direction.
Commented on a couple of doubts.

test/support/factory.ex
arbulu89: `env_factory` creates data based in our domain (where the env is simply a map).
`env_event_factory` creates the env dataset coming in the `ExecutionRequested` env field.
They are different, as the 2nd is strictly attached to protobuf and it is only used to send and receive messages, not internally.

Maybe I could rename to `execution_requested_env_factory`

nelsonkopliku (approved):
LGTM

lib/wanda/execution/behaviour.ex
```diff
@@ -9,14 +9,14 @@ defmodule Wanda.Execution.Behaviour do
               execution_id :: String.t(),
               group_id :: String.t(),
               targets :: [Target.t()],
-              env :: %{String.t() => String.t()}
+              env :: map()
```
fabriziosestito, line 12: We could be more precise in the spec, `%{String.t() => boolean() | number() | String.t()}`

fabriziosestito (changes requested):

test/policy_test.exs
```diff
@@ -19,10 +19,7 @@ defmodule Wanda.PolicyTest do
     agent_id = UUID.uuid4()
     env = build(:execution_requested_env)
 
-    expected_env =
-      Enum.reduce(env, %{}, fn {key, %{kind: {_, value}}}, acc ->
-        Map.put(acc, key, value)
-      end)
+    expected_env = Map.new(env, fn {key, %{kind: {_, value}}} -> {key, value} end)
```
fabriziosestito, line 22: we are repeating the implementation in the test, which makes the test of little value.

arbulu89:
@fabriziosestito Thanks! All the feedback implemented

test/policy_test.exs
nelsonkopliku: Ah yeah! Glad you spotted! I overlooked that. My bad :sweat_smile:

test/support/factory.ex
```diff
@@ -51,6 +51,22 @@ defmodule Wanda.Factory do
...
+    count = Map.get(attrs, :count, 5)
+
+    Enum.reduce(0..count, %{}, fn _, acc ->
+      Map.put(acc, Faker.Pokemon.name(), random_env_value())
+    end)
+  end
+
+  defp random_env_value do
```
fabriziosestito, line 62: defp should always be at the end, after the def functions

fabriziosestito (approved):
LGTM, just a minor comment

PR #43, opened 2022-10-05
Load check values from yaml

arbulu89:
Load the new `values` field from the yaml.

PD: I will update the tests using the current check test fixtures in the evaluation UT in a next PR. I want to get rid off them in the evaluation tests and test them properly with factories

### DIFF
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -80,18 +80,16 @@ defmodule Wanda.Catalog do
   defp map_values(_), do: []
 
   defp map_value(%{"name" => name, "default" => default} = value) do
-    conditions = Map.get(value, "conditions", [])
+    conditions =
+      value
+      |> Map.get("conditions", [])
+      |> Enum.map(fn %{"value" => condition_value, "when" => expression} ->
+        %Condition{
+          value: condition_value,
+          expression: expression
+        }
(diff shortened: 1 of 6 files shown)


nelsonkopliku (approved):

lib/wanda/catalog.ex
```diff
@@ -69,4 +72,26 @@ defmodule Wanda.Catalog do
...
   end
+
+  defp map_values(%{"values" => values}) do
+    Enum.map(values, &map_value/1)
+  end
+
+  defp map_values(_), do: []
+
+  defp map_value(%{"name" => name, "default" => default} = value) do
+    conditions = Map.get(value, "conditions", [])
```
dottorblaster, line 83:
```elixir
value
|> Map.get("conditions", [])
|> Enum.map(fn %{"value" => value, "when" => expression} ->
        %Condition{
          value: value,
          expression: expression
        }
      end)
```

And then you can actually build the Value as

```elixir
%Value{name: name, default: default, conditions: conditions}
```

dottorblaster (changes requested):
Just a tiny comment

lib/wanda/catalog.ex
arbulu89: Done

PR #44, opened 2022-10-05
Add mix deps.unlock --check-unused to CI

jamie-suse:
Related to https://github.com/trento-project/web/pull/872


fabriziosestito:
@jamie-suse since we are using dependency cache, you'll need to `mix deps.unlock --check-unused`, and add the resulting diffs to a commit in this PR

.github/workflows/ci.yaml
```diff
@@ -50,6 +50,7 @@ jobs:
           mix local.rebar --force
           mix local.hex --force
           mix deps.get
+          mix deps.unlock --check-unused
```
fabriziosestito, line 53: this should be a step of "Static Code Analysis"

fabriziosestito (changes requested):

.github/workflows/ci.yaml
jamie-suse: Moved to a step under Static Analysis, after `Retrieve Elixir Cached Dependencies`.

dottorblaster (approved):
LGTM

PR #45, opened 2022-10-06
Switch to Views for JSON rendering

dottorblaster:
We just need some views in place to avoid JSON raw rendering

### DIFF
--- a/lib/wanda_web/controllers/executions_controller.ex
+++ b/lib/wanda_web/controllers/executions_controller.ex
@@ -26,6 +26,6 @@ defmodule WandaWeb.ExecutionsController do
   def list_checks_executions(conn, params) do
     results = Results.list_execution_results(params)
     total_count = Results.count_execution_results(params)
-    json(conn, %{data: results, total_count: total_count})
+    render(conn, data: results, total_count: total_count)
   end
 end
--- a/lib/wanda_web/views/executions_view.ex
+++ b/lib/wanda_web/views/executions_view.ex
@@ -1,0 +1,7 @@
+defmodule WandaWeb.ExecutionsView do
+  use WandaWeb, :view
+
+  def render("list_checks_executions.json", %{data: data, total_count: total_count}) do
+    %{data: data, total_count: total_count}
+  end
+end


lib/wanda_web/views/executions_view.ex
```diff
@@ -0,0 +1,7 @@
+defmodule WandaWeb.ExecutionsView do
+  use WandaWeb, :view
+
+  def render("list_checks_executions.json", %{data: data, total_count: total_count}) do
```
arbulu89, line 4: Following `phoenix` convention, this function should be named `index`, but well, this name works as well.

arbulu89, line 4: I don't see the need to bringing the `total_count` here, we could simply use `length`. I have seen it widely this way. And we need to use `executions` in the 2nd param to follow the convention and make phoenix work properly
```
def render("list_checks_executions.json", %{executions: executions}) do
    %{
      data: render_many(executions, ExecutionView, "execution.json"),
      total_count: length(executions)
    }
  end
```

lib/wanda_web/views/executions_view.ex
```diff
@@ -0,0 +1,7 @@
...
+    %{data: data, total_count: total_count}
```
arbulu89, line 5: We could have the single execution render as well:
```
def render("execution.json", %{execution: execution}) do
    execution
  end
```

fabriziosestito (approved):

lib/wanda_web/controllers/executions_controller.ex
```diff
@@ -26,6 +26,6 @@ defmodule WandaWeb.ExecutionsController do
   def list_checks_executions(conn, params) do
     results = Results.list_execution_results(params)
     total_count = Results.count_execution_results(params)
-    json(conn, %{data: results, total_count: total_count})
+    render(conn, data: results, total_count: total_count)
```
arbulu89, line 29: This should be changed to (at least, the parameters usage):
```
def list_checks_executions(conn, params) do
    executions = Results.list_execution_results(params)
    render(conn, "list_checks_executions.json", executions: executions)
  end
```

arbulu89 (changes requested):
Hey @dottorblaster ,
Now that we have started following phoenix conventions. Here some comments:

As far as I have seen, phoenix works with some implicit assumptions.
- The controller and view names should be singular. In our case `execution_controller.ex` and `execution_view.ex`. Otherwise, phoenix fails to run some operations.
- The used naming in the `render` functions must use the some specific names

References:
https://github.com/gothinkster/elixir-phoenix-realworld-example-app
https://dev.to/bluzky/a-better-way-to-render-json-response-in-elixir-phoenix-41a3

lib/wanda_web/views/executions_view.ex
dottorblaster: We need the total count because if we paginate, say, 100 entries, with 10 items per page, the total count would still be a hundred while the length of the results we're rendering is 10

dottorblaster: Anyway I think you're right about the `render_many`, I'll try that :+1:

arbulu89: I meant the second variable hehe `index.json` hehe
And it should be named in the same name in the controller. But phoenix accepts any other naming in this case

arbulu89: Yes, feel free to leave the current name if you prefer so

lib/wanda_web/controllers/executions_controller.ex
dottorblaster: Looking at the GitHub API they just call an array of results `items`, I'd feel like having a more generic name could be something valuable for users that interact with the API. @fabriziosestito what do you think?

fabriziosestito: `items` is good

arbulu89: About the naming, I'm not sure if you could have this:
`render(conn, "list_checks_executions.json", items: items)`
I think it would fail, as phoenix expects here the same name in plural of the view (`ExecutionView` -> `executions`).
`render(conn, "list_checks_executions.json", executions: items)`, would work

fabriziosestito: I was about to rename everything to result, we are using the result context and we have so many different name here. I will add a cleanup PR to show you guys if we merge this.

test/wanda_web/views/execution_view_test.exs
```diff
@@ -0,0 +1,31 @@
...
+  describe "ExecutionView" do
+    test "renders list_checks_executions.json" do
+      [
+        %ExecutionResult{execution_id: execution_1, group_id: group_1},
+        %ExecutionResult{execution_id: execution_2, group_id: group_2},
+        %ExecutionResult{execution_id: execution_3, group_id: group_3}
+      ] = data = build_list(3, :execution_result)
+
+      assert %{
+               data: [
```
arbulu89, line 18: I don't know if in this assert `data: ^data` would work (I think it should). It is certainly shorter hehe

arbulu89 (approved):
Thanks dude!
Green green light!

test/wanda_web/views/execution_view_test.exs
dottorblaster: I think you're right, I'll try in another PR so that @fabriziosestito doesn't get blocked too much by this :smile:

PR #46, opened 2022-10-06
Add support for values computation based on environment

nelsonkopliku:
# Description

This PR allows the engine to determine, based on the environment, the expected value for a check execution/evaluation.

```yaml
id: 156F64
name: Corosync configuration file
...
facts:
  - name: corosync_token_timeout
    gatherer: corosync.conf
    argument: totem.token
values:
  - name: expected_token_timeout
    default: 5000
    conditions:
      - value: 30000
        when: env.provider == "azure" || env.provider == "aws"
      - value: 20000
        when: env.provider == "gcp"
expectations:
  - name: timeout
    expect: facts.corosync_token_timeout == values.expected_token_timeout
```

- env is provided by the dashboard when requesting an execution
- values are computed when all the facts are available
- computed values for an execution are provided back as part of the execution result
- using `facts.` and `values.` as namespaces for respective items

## How was this tested?

Test about expecting different values for different environments were added here `test/execution/evaluation_test.exs`.
They test against check **156F64**


test/fixtures/catalog/156F64.yaml
```diff
@@ -0,0 +1,25 @@
+id: 156F64
```
arbulu89, line 1: I wouldn't have the same real check that we have in the catalog here. In fact, I would try not to add a new test fixture, as we are trying to remove them for the evaluation test cases in other PR. It will be possible to use factories to create this use cases

lib/wanda/execution/evaluation.ex
```diff
@@ -18,44 +18,52 @@ defmodule Wanda.Execution.Evaluation do
...
-  @dialyzer {:nowarn_function, eval_expectation: 2}
+  @dialyzer {:nowarn_function, eval_expectation: 3, value_from_conditions: 3}
 
-  @spec execute(String.t(), String.t(), [Check.t()], map()) :: Result.t()
-  def execute(execution_id, group_id, checks, gathered_facts, timeouts \\ []) do
+  @spec execute(
+          String.t(),
+          String.t(),
+          [Check.t()],
+          %{String.t() => boolean() | number() | String.t()},
```
arbulu89, line 27: The signature looks wrong. You have changed the order between `gathered_facts` (which was a map) with the new `env`.

lib/wanda/execution/evaluation.ex
```diff
@@ -82,10 +90,14 @@ defmodule Wanda.Execution.Evaluation do
         message: "Fact gathering ocurred during the execution"
       }
     else
+      computed_values = compute_values(values, facts, env)
```
arbulu89, line 93: I don't think we need this `compute_values` function. It only has one `Enum` usage. I would directly put here:
```
Enum.map(values, &value_from_conditions(&1, facts, env))
```
And I would name the next function `eval_values`, as we already have a `eval_expectations`

lib/wanda/execution/evaluation.ex
```diff
@@ -224,4 +237,51 @@ defmodule Wanda.Execution.Evaluation do
   defp result_weight(:critical), do: 2
   defp result_weight(:warning), do: 1
   defp result_weight(:passing), do: 0
+
+  defp compute_values(values, facts, env) do
+    Enum.map(values, &value_from_conditions(&1, facts, env))
+  end
+
+  defp value_from_conditions(
```
arbulu89, line 245: Let's put this function just after the `has_some_fact_gathering_error` function. Usually, we put the elixir function as they appear and are used in the code

lib/wanda/execution/evaluation.ex
```diff
@@ -224,4 +237,51 @@ defmodule Wanda.Execution.Evaluation do
...
+        case Abacus.eval(expression, %{"facts" => scoped_facts, "env" => env}) do
+          {:ok, true} ->
+            value
+
+          _ ->
+            false
+        end
+      end)
+
+    %Wanda.Execution.Value{
```
arbulu89, line 270: Alias this `Value` at the top

lib/wanda/execution/evaluation.ex
```diff
@@ -224,4 +237,51 @@ defmodule Wanda.Execution.Evaluation do
...
+           default: default,
+           conditions: conditions
+         },
+         facts,
+         env
+       ) do
+    scoped_facts = map_facts_to_evaluation_scope(facts)
+
+    computed_value =
+      Enum.find_value(conditions, default, fn %Wanda.Catalog.Condition{
```
arbulu89, line 257: Alias `Condition` at the top. This way, you can put the `Condition` declaration in one line even, which looks easier to

lib/wanda/execution/evaluation.ex
```diff
@@ -224,4 +237,51 @@ defmodule Wanda.Execution.Evaluation do
...
+           name: name,
+           default: default,
+           conditions: conditions
+         },
+         facts,
+         env
+       ) do
+    scoped_facts = map_facts_to_evaluation_scope(facts)
+
+    computed_value =
```
arbulu89, line 256: Maybe better `evaluated_value`. But I would directly put the output of the `Enum.find_value` when we create the `Value` struct:
```
 %Value{
      name: name,
      value:
        Enum.find_value(conditions, default, fn %Condition{value: value, expression: expression} ->
...
...
```

lib/wanda/execution/evaluation.ex
```diff
@@ -99,14 +111,15 @@ defmodule Wanda.Execution.Evaluation do
...
          %Expectation{name: name, type: type, expression: expression},
-         facts
+         facts,
+         values
        ) do
-    scoped_facts =
-      Enum.reduce(facts, %{}, fn %Fact{name: name, value: value}, acc ->
-        put_in(acc, [name], value)
-      end)
+    evaluation_scope = %{
```
arbulu89, line 117: In order to have some consistency with the evaluation scope, I would create a generic function like
`namespace_scope`. Something like:
```
def namespace_scope(facts, values, env) do
    scoped_facts =
      Enum.reduce(facts, %{}, fn %Fact{name: name, value: value}, acc ->
        put_in(acc, [name], value)
      end)

    scoped_values =
      Enum.reduce(values, %{}, fn %Value{name: name, value: value}, acc ->
        put_in(acc, [name], value)
      end)

    %{"facts" => scoped_facts, "values" => scoped_values, "env" => env}
  end
```

This way, we can create the evaluation scope more or less the same way in all the evaluations.

For example, above, we could have something like: 
```
values = Enum.map(values, &eval_value(&1, namespace_scope(facts, [], env)))
# and for the expectations
Enum.map(expectations, &eval_expectation(&1, namespace_scope(facts, values, env)))
```

lib/wanda/execution/evaluation.ex
fabriziosestito, line 257: you might want to the lambda this in a separate function, since the level of nesting is very hard to read

arbulu89 (changes requested):
The functionality looks good.
I have suggested some changes.
About the tests, I would try to not use test fixtures and use a check created by factories (already working on this)

lib/wanda/execution/evaluation.ex
fabriziosestito: +1

lib/wanda/execution/evaluation.ex
arbulu89: We could even have a struct to make it more strict, with some function to convert it to map

lib/wanda/execution/evaluation.ex
```diff
@@ -224,4 +237,51 @@ defmodule Wanda.Execution.Evaluation do
...
+
+  defp map_facts_to_evaluation_scope(facts) do
+    Enum.reduce(facts, %{}, fn %Fact{name: name, value: value}, acc ->
+      put_in(acc, [name], value)
+    end)
+  end
+
+  defp map_values_to_evaluation_scope(values) do
+    Enum.reduce(values, %{}, fn %Wanda.Execution.Value{name: name, value: value}, acc ->
+      put_in(acc, [name], value)
```
fabriziosestito, line 284: not sure you need put_in here, this is used usually for nested maps. 
We are trying to build a map where the name of the execution/fact is the key and value is the value.

```elixir 
Enum.into(%{}, fn %{name: name, value: value} -> {name, value} end)
```

Also, consider that those two functions are the same basically (since both fact and value have name and value fields) so they might be redundant.

fabriziosestito (dismissed):

lib/wanda/execution/evaluation.ex
fabriziosestito: `eval_values` as xabi proposed

lib/wanda/execution/evaluation.ex
fabriziosestito: we could use the builder pattern, something like:
```elixir
%{} |> add_scope("facts", facts) |> add_scope("values", values)
```

where `add_scope/3` is a function that takes a map as first parameter and does the transformation I proposed here: https://github.com/trento-project/wanda/pull/46#discussion_r989742219

arbulu89: Nice @fabriziosestito , I like this approach more. Only 1 new function

lib/wanda/execution/evaluation.ex
nelsonkopliku: Right. I believe I got confused by the missing type for `timeouts`.

nelsonkopliku:
Alright @fabriziosestito @arbulu89, thanks for the reviews!
I addressed all the feedbacks here d659f7e58b1fa5d7550086ee1cb38b7dd80f2d6f.
All but this https://github.com/trento-project/wanda/pull/46#discussion_r989728532 about the tests.
Let's postpone the discussion/decision to monday!

lib/wanda/execution/evaluation.ex
```diff
@@ -82,10 +92,14 @@ defmodule Wanda.Execution.Evaluation do
         message: "Fact gathering ocurred during the execution"
       }
     else
+      evalued_values = Enum.map(values, &eval_value(&1, facts, env))
```
rtorrero, line 95: did you mean here `evaluated_values`?

lib/wanda/execution/evaluation.ex
```diff
@@ -82,10 +92,14 @@ defmodule Wanda.Execution.Evaluation do
         message: "Fact gathering ocurred during the execution"
```
rtorrero, line 92: Not related to this PR but it might be worth a quick fix: is `error` missing in the sentence probably?

rtorrero (approved):
Thanks for the PR! Just some very minor comments :-)

I also don't really like that we add the same check in the fixtures; it's very likely that they are going to diverge at some point and cause a bit of confusion. Approving, just change the `evalued` thingy and merge when you are ready

lib/wanda/execution/evaluation.ex
nelsonkopliku: fixed here ad731d5d300b6400664d930fd4fb83bfcdb5ec63

lib/wanda/execution/evaluation.ex
nelsonkopliku: fixed here ad731d5d300b6400664d930fd4fb83bfcdb5ec63

nelsonkopliku:
@arbulu89 @fabriziosestito @rtorrero missing things addressed here ad731d5d300b6400664d930fd4fb83bfcdb5ec63.
Please take a look :smile:

rtorrero:
> @arbulu89 @fabriziosestito @rtorrero missing things addressed here [ad731d5](https://github.com/trento-project/wanda/commit/ad731d5d300b6400664d930fd4fb83bfcdb5ec63). Please take a look smile

Everything seems fine now @nelsonkopliku, thanks!

arbulu89 (approved):
Thank you @nelsonkopliku 
Green light from my side!!!
I personally would change how the tests look, but well, they test what they should, so in that matter they do their work!

test/execution/evaluation_test.exs
```diff
@@ -622,7 +624,391 @@ defmodule Wanda.Execution.EvaluationTest do
                  }
                ],
                result: :warning
-             } = Evaluation.execute(execution_id, group_id, checks, gathered_facts)
+             } = Evaluation.execute(execution_id, group_id, checks, gathered_facts, %{})
+    end
+  end
+
+  describe "environment based check evaluation" do
+    defp build_corosync_token_timeout_check do
```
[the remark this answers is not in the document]
nelsonkopliku: thanks, changed.

lib/wanda/execution/evaluation.ex
```diff
@@ -67,25 +77,29 @@ defmodule Wanda.Execution.Evaluation do
...
     if has_some_fact_gathering_error?(facts) do
       %AgentCheckError{
         agent_id: agent_id,
         facts: facts,
         type: :fact_gathering_error,
-        message: "Fact gathering ocurred during the execution"
+        message: "Fact gathering error occurred during the execution"
       }
     else
+      evaluated_values = Enum.map(values, &eval_value(&1, facts, env))
```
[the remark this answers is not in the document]
nelsonkopliku: changed like this
```elixir
defp add_agent_check_result_or_error(agent_id, facts, expectations, env, values) do
  if has_some_fact_gathering_error?(facts) do
    ...
  else
    value_evaluation_scope = add_scope(%{"env" => env}, "facts", facts)

    evaluated_values = Enum.map(values, &eval_value(&1, value_evaluation_scope))

    expectation_evaluation_scope =
      %{}
      |> add_scope("facts", facts)
      |> add_scope("values", evaluated_values)

    %AgentCheckResult{
      agent_id: agent_id,
      facts: facts,
      values: evaluated_values,
      expectation_evaluations:
        Enum.map(expectations, &eval_expectation(&1, expectation_evaluation_scope))
    }
  end
end
```

I also tried to do
```elixir
%{}
    |> add_scope("env", env)
    |> add_scope("facts", facts)
```
but it came out that that would require having two versions of `add_scope/3` where one would accept a list as the third argument (when we pass a list of facts/values), while the other would accept a map (when we pass the `env`), and this I understood not being idiomatic :sweat_smile:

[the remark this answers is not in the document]
arbulu89: well, you could have `add_scope(%{"env" => env}, "facts", facts)` this of course XD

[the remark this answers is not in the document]
nelsonkopliku: indeed it is that way :wink:

test/execution/evaluation_test.exs
```diff
@@ -622,7 +624,391 @@ defmodule Wanda.Execution.EvaluationTest do
...
+      build(:check,
+        id: "156F64",
```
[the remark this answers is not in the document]
nelsonkopliku: swapped references to dummy check :wink:

PR #47, opened 2022-10-07
Refactor evaluation tests

arbulu89:
Refactor evaluation unit tests to use factories instead of test fixtures (which depend on other code), fix factories and clean up test fixtures.

All the changes follow more or less the same pattern (replace `Checks.get_check` by the factory based check)


test/execution/server_test.exs
```diff
@@ -72,13 +86,13 @@ defmodule Wanda.Execution.ServerTest do
       assert ExecutionResult |> Repo.all() |> Enum.empty?()
     end
 
-    test "should exit when all facts are sent by all agents" do
+    test "should exit when all facts are sent by all agents", state do
```
fabriziosestito, line 89: the "state" here is usually referred as "context". Also you can already pattern match %{check: %{id: id}}

fabriziosestito (changes requested):
Left a comment regarding the test context usage.

I think the goal here is to remove the catalog fixtures completely.
We could refactor further the factory to use the builder pattern,
see here: https://github.com/thoughtbot/ex_machina#flexible-factories-with-pipes
so we could do something like this:

```elixir
build(:check) |> with_expectation(:expect_same) |> with_severity(:critical) ....
```

test/execution/server_test.exs
arbulu89: @fabriziosestito About removing the catalog fixtures completely: we wouldn't be testing the load from yaml. That's why they are still some there. They are not used for anything else. At the end, even though it is simple, we do some operations there that are good to test.

fabriziosestito: @arbulu89 ah ok that make sense

fabriziosestito (approved):

nelsonkopliku (approved):
As soon as we merge this PR, I am gonna adjust this other https://github.com/trento-project/wanda/pull/46

PR #48, opened 2022-10-07
Cleanup execution controller

fabriziosestito:
Some refactor, cleanup, renaming.
Fixed the and improved controller tests that was always returning green.

Follow up PR will change a bit the database table and add the proper object schema


lib/wanda_web/schemas/list_results_response.ex
```diff
@@ -0,0 +1,20 @@
+defmodule WandaWeb.Schemas.ListResultsResponse do
+  @moduledoc """
+  Check execution response API spec
```
arbulu89, line 3: Old docstring

lib/wanda_web/views/result_view.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  def render("list_results.json", %{results: results, total_count: total_count}) do
+    %{
+      items: render_many(results, WandaWeb.ResultView, "result.json", as: :result),
+      total_count: total_count
+    }
+  end
+
+  def render("result.json", %{result: %ExecutionResult{} = result}) do
+    result
```
arbulu89, line 14: Here we depend on the internal domain. We could maybe at some point create mappers to provide the frontend specific example (even though it is exactly the same). Just as possibility

arbulu89 (commented):
cool! i have more or less the same for the catalog

lib/wanda_web/controllers/result_controller.ex
```diff
@@ -1,31 +1,33 @@
...
         example: "00000000-0000-0000-0000-000000000001"
       ],
       page: [in: :query, description: "Page", type: :integer, example: 3],
       items_per_page: [in: :query, description: "Items per page", type: :integer, example: 20]
     ],
-    responses: [
-      ok: {"Check executions response", "application/json", CheckExecutionResponse}
-    ]
+    responses: %{
+      200 => {"List results response", "application/json", ListResultsResponse},
```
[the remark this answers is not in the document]
fabriziosestito: honestly i just copied it  from the documentation :D

lib/wanda_web/views/result_view.ex
fabriziosestito: well, this is not really a part of the domain since `ExecutionResult` is the persistence, since it is the ecto schema. I think the mapper you are referring to could be the view itself.

dottorblaster (approved):
LGTM

CDimonaco (approved):

lib/wanda_web/views/result_view.ex
arbulu89: Yes, the view is the mapper. Now we just send our persisted data, which by now would be good enough. I meant that we might want to map to a specific struct (not now).

PR #51, opened 2022-10-11
Add initialization tasks for a release

arbulu89:
More goodies for a "prod/released". It adds the auto migration feature. It can be used like:
` bin/wanda eval "Wanda.Release.init"`

Ref: https://hexdocs.pm/ecto_sql/Ecto.Migrator.html#module-example-running-migrations-in-a-release
I did some small changes to make it more simillar to the web, adding the `init` function to just have an unique entry point in the helm chart init container


dottorblaster (approved):
LGTM

nelsonkopliku (approved):

PR #54, opened 2022-10-12
Rename controllers context

fabriziosestito:
Rename controller actions and context to be as close as possible to Phoenix standards, so we can use the `resource` directive in the router.

### DIFF
--- a/lib/wanda/results.ex
+++ b/lib/wanda/results.ex
@@ -3,22 +3,24 @@ defmodule Wanda.Results do
   This module exposes functionalities to interact with the historycal log of ExecutionResults.
   """
 
+  alias Wanda.Repo
+
   alias Wanda.Execution.Result
   alias Wanda.Results.ExecutionResult
 
   import Ecto.Query
 
   @doc """
-  Saves a result to the database.
+  Create a new result.
   """
-  @spec save_result(Result.t()) :: ExecutionResult.t()
-  def save_result(
+  @spec create_execution_result(Result.t()) :: ExecutionResult.t()
+  def create_execution_result(
         %Result{
           execution_id: execution_id,
           group_id: group_id
         } = result
       ) do
-    Wanda.Repo.insert!(%ExecutionResult{
+    Repo.insert!(%ExecutionResult{
       execution_id: execution_id,
       group_id: group_id,
       payload: result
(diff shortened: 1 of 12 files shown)


lib/wanda/results.ex
```diff
@@ -3,33 +3,37 @@ defmodule Wanda.Results do
...
-  def save_result(
+  @spec create_execution_result(Result.t()) :: ExecutionResult.t()
+  def create_execution_result(
         %Result{
           execution_id: execution_id,
           group_id: group_id
         } = result
       ) do
-    Wanda.Repo.insert!(%ExecutionResult{
+    Repo.insert!(%ExecutionResult{
```
dottorblaster, line 23: Shouldn't have Credo warned us about this?

dottorblaster (approved):
LGTM

nelsonkopliku (approved):

lib/wanda/results.ex
fabriziosestito: Credo warns only if it is nested 2 or 3 levels. 
Sometimes it is useful to alias with one level of nesting, for example when you have overlapping module names from different "contexts". This is not the case, tho.

lib/wanda/results.ex
```diff
@@ -3,33 +3,37 @@ defmodule Wanda.Results do
...
 
   import Ecto.Query
 
   @doc """
-  Saves a result to the database.
+  Create a new result.
   """
-  @spec save_result(Result.t()) :: ExecutionResult.t()
-  def save_result(
+  @spec create_execution_result(Result.t()) :: ExecutionResult.t()
```
arbulu89, line 16: Now the strange this is that we sillt use `ExecutionResult`

lib/wanda_web/controllers/catalog_controller.ex
```diff
@@ -5,13 +5,13 @@ defmodule WandaWeb.CatalogController do
   alias Wanda.Catalog
   alias WandaWeb.Schemas.CatalogResponse
 
-  operation :list_catalog,
+  operation :catalog,
```
arbulu89, line 8: Should be name this `index` too?

priv/repo/seeds.exs
```diff
@@ -13,7 +13,7 @@
 execution_id = "00000000-0000-0000-0000-000000000001"
 group_id = "00000000-0000-0000-0000-000000000002"
 
-Wanda.Results.save_result(%Wanda.Execution.Result{
+Wanda.Results.create_execution_result(%Wanda.Execution.Result{
```
arbulu89, line 16: We should start considering at some point to remove this seed

arbulu89 (approved):

lib/wanda_web/controllers/catalog_controller.ex
fabriziosestito: index is usually used for resources, this would work if the catalog path was /checks/

I think the catalog is a special case, so I am following what the real world app is doing with the `feed`:
https://github.com/gothinkster/elixir-phoenix-realworld-example-app/blob/8620fb2829703051cdbd032f93be6cd941c4810e/lib/real_world_web/router.ex#L21

PR #55, opened 2022-10-13
Add get check result

fabriziosestito:
Add `get_check_result to `Results` context and `/api/checks/results/:id` show action to `ResultController`

### DIFF
--- a/lib/wanda/execution/server.ex
+++ b/lib/wanda/execution/server.ex
@@ -148,7 +148,7 @@ defmodule Wanda.Execution.Server do
   end
 
   defp store_and_publish_execution_result(%Result{} = result) do
-    Results.create_execution_result(result)
+    Results.create_execution_result!(result)
 
     execution_completed = Messaging.Mapper.to_execution_completed(result)
     :ok = Messaging.publish("results", execution_completed)
--- a/lib/wanda_web/views/result_view.ex
+++ b/lib/wanda_web/views/result_view.ex
@@ -2,12 +2,17 @@ defmodule WandaWeb.ResultView do
   use WandaWeb, :view
 
   alias Wanda.Results.ExecutionResult
+  alias WandaWeb.ResultView
 
   def render("index.json", %{results: results, total_count: total_count}) do
     %{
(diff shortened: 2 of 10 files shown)


lib/wanda/execution/server.ex
```diff
@@ -148,7 +148,7 @@ defmodule Wanda.Execution.Server do
   end
 
   defp store_and_publish_execution_result(%Result{} = result) do
```
nelsonkopliku, line 150: Should this be banged as well? `store_and_publish_execution_result!`
Or it does not matter because its private? :thinking:

nelsonkopliku (approved):
Looks good

lib/wanda_web/views/result_view.ex
```diff
@@ -2,14 +2,19 @@ defmodule WandaWeb.ResultView do
...
 
   def render("index.json", %{results: results, total_count: total_count}) do
     %{
-      items: render_many(results, WandaWeb.ResultView, "result.json", as: :result),
+      items: render_many(results, ResultView, "result.json"),
       total_count: total_count
     }
   end
 
+  def render("show.json", %{result: result}) do
```
arbulu89, line 14: This looks like a smell. We have the `show.json` which should be used to render 1 result, and after that we have the `result.json` which is used for the same thing with other name.
I think we are missing something there. We should only have the `show` one there and do the mapping (the Map.put) there.

arbulu89 (approved):
It looks good @fabriziosestito 
I think there is something strange in the new `show` and old `result` views. The 2nd looks that is not needed

lib/wanda_web/views/result_view.ex
fabriziosestito: I think this is how it should be, see: https://hexdocs.pm/phoenix/1.4.2/views.html#rendering-json

lib/wanda/execution/server.ex
fabriziosestito: --> https://elixirforum.com/t/when-to-suffix-a-function-by-a-bang/34490

PR #57, opened 2022-10-16
Checks catalog

nelsonkopliku:


stefanotorresi:
as a side note, please bear in mind that we want to get rid of the old x.y.z numbering scheme and leave just the base16 identifier

priv/catalog/1.3.1-0B6DB2.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  GCP:
+
+
+  SUSE / KVM:
+
+    - https://documentation.suse.com/sle-ha/15-SP3/html/SLE-HA-all/cha-ha-storage-protect.html
+    - https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#id-1.11.5.5.4.4
+
+run_when: env.provider != "aws" && env.provider != "gcp"
```
angelabriel, line 49: Is 'run_when:' already implemented and usable or is this WIP?

fabriziosestito: @angelabriel please ignore this PR as it was discovery work and cannot be taken as reference right now

angelabriel: ok, many thanks for the clarification

PR #58, opened 2022-10-18
Use google protobuf value

fabriziosestito:
Adapt policy code to fit with the new format that is using google protobuf Value instead of our own sum type


test/policy_test.exs
```diff
@@ -67,7 +67,11 @@ defmodule Wanda.PolicyTest do
                execution_id: execution_id,
                agent_id: agent_id,
                facts_gathered: [
-                 %{check_id: "check_id", name: "name", value: {:text_value, "value"}}
+                 %{
+                   check_id: "check_id",
+                   name: "name",
+                   fact_value: {:value, %{kind: {:text_value, "value"}}}
```
arbulu89, line 73: `:string_value`

arbulu89 (approved):
good!
fix ci removing uuid deps and the test, and green light

nelsonkopliku (approved):

PR #59, opened 2022-10-18
Chore: remove unused miss dep

fabriziosestito:
We don't use `miss` anymore


nelsonkopliku (approved):

arbulu89 (commented):
How is this possible? The CI should've caught this

fabriziosestito:
> How is this possible? The CI should've caught this

@arbulu89 the CI checks if we have locked deps that are not in the `mix.exs` anymore.

PR #60, opened 2022-10-19
Store execution state

nelsonkopliku:
# Description

This PR mainly adds the capability to track the status (`running`, `completed`) of an Execution.
- a `running` execution would be created when the execution process starts
- trying to create an an already existent execution would fail
- when an execution ends, the previously persisted entity transits to `completed` and the payload of the result gets populated
- trying to complete an already completed execution, ~~would just pass through without changing anything (? see Doubts)~~ would return an error, so we can act accordingly

## How was this tested?

It was tested by 
- adjusting `test/results_test.exs` to consider the new shape of the state
- extending `test/results_test.exs` to cover execution completion scenario
- adjusting `test/execution/server_test.exs` to reflect the new requirements

### Naming
At the moment we are relying on `Wanda.Results.ExecutionResult`, yet it seems to not properly represent the actual thing anymore.
We are indeed talking about an `Execution` that is either `running` or `completed`, thus it might make little sense for an `ExecutionResult` to be `running` or `completed`.

However for the sake of simplicity we're sticking to the improper `ExecutionResult` for the moment, deferring a bit that refactor which might require to revisit also other stuff (will track it down as tech debt) 

### Doubts

Would it help to have also `timed_out` (next to `running` and `completed`) as a state for an execution? Otherwise the execution would still be `completed` but the internal information would tell that it is actually timed out

```elixir
# lib/wanda/results/execution_result.ex

field :status, Ecto.Enum, values: [:running, :completed], default: :running
```

---

What happens if message publishing fails after a successful tracking of the execuiton?

```elixir
# lib/wanda/execution/server.ex

def handle_continue(:start_execution, ...) do
  ...
  Results.create_execution_result!(execution_id, group_id, targets)
  # what if creation succeeds but publishing fails?
  :ok = Messaging.publish("agents", facts_gathering_requested)
  ...
end

...

defp store_and_publish_execution_result(%Result{} = result) do
  Results.complete_execution_result!(result)
  ...
  # what if completion succeeds but publishing fails?
  :ok = Messaging.publish("results", execution_completed)
end
```

---

~~As mentioned, trying to complete an already completed execution, would just pass through without changing anything
https://github.com/trento-project/wanda/pull/60/files#diff-79feda40eb263347d33b6e0b840efee3b28823990ab1a0f71a503407d24b7e78R72
Now that I am thinking of it, probably it would be better to raise an error if trying to complete an already completed execution as it would avoid sending out an already sent message about check execution being completed :thinking:~~

**Addressed differently**: we return an error when we try to complete an already completed execution so we can act accordingly https://github.com/trento-project/wanda/pull/60/commits/a272e9ce8df9bd3dcb4de5e0ee618ffa3e67472e

### What's next
- Adjust API layer to properly consider the `running`/`completed` differentiation


lib/wanda/results/execution_result.ex
```diff
@@ -5,15 +5,48 @@ defmodule Wanda.Results.ExecutionResult do
...
   @type t :: %__MODULE__{}
 
   @derive {Jason.Encoder, [except: [:__meta__]]}
   @primary_key false
   schema "execution_results" do
     field :execution_id, Ecto.UUID, primary_key: true
     field :group_id, Ecto.UUID
-    field :payload, :map
+    field :payload, :map, default: %{}
+
```
arbulu89, line 20: Why this white space here? XD

priv/repo/migrations/20220928131105_add_execution_result.exs
```diff
@@ -6,8 +6,12 @@ defmodule Wanda.Repo.Migrations.AddExecutionResult do
       add :execution_id, :uuid, primary_key: true
       add :group_id, :uuid, null: false
       add :payload, :map, null: false
+      add :status, :string, null: false
+      add :targets, :map, null: false, default: "[]"
```
arbulu89, line 10: Do we need this default value here for some reason?
It could as well be `{:array, :map}` as well, to enforce being a list.
I don't know if it works though

test/execution/server_test.exs
```diff
@@ -83,7 +82,8 @@ defmodule Wanda.Execution.ServerTest do
       )
 
       assert_receive :wandalorian
-      assert ExecutionResult |> Repo.all() |> Enum.empty?()
+
+      refute ExecutionResult |> Repo.all() |> Enum.empty?()
```
arbulu89, line 86: This assertion looks pretty weak. Wouldn't be more explicit to check that we have a new execution on running state with that execution_id? This way we test that the execution is in running state as well

test/results_test.exs
```diff
@@ -4,35 +4,41 @@ defmodule Wanda.ResultsTest do
...
       assert %ExecutionResult{
                execution_id: ^execution_id,
                group_id: ^group_id,
-               payload: %{
-                 "result" => "passing",
-                 "check_results" => [_ | _]
-               }
+               status: :running,
+               payload: %{},
+               targets: [
```
arbulu89, line 32: I think we could simply `targets: ^targets` here, to not have this explicit pins

test/results_test.exs
```diff
@@ -58,13 +68,118 @@ defmodule Wanda.ResultsTest do
...
+                   }
+                 ]
                }
              ] = Repo.all(ExecutionResult)
     end
+
+    test "should raise an error when trying to create an already existing execution" do
+      [
+        %ExecutionResult{execution_id: execution_id, group_id: group_id},
+        %ExecutionResult{}
```
arbulu89, line 90: What is this empty execution result for?

test/support/factory.ex
```diff
@@ -150,15 +150,31 @@ defmodule Wanda.Factory do
...
+      targets: Map.get(attrs, :targets, []),
+      payload: payload,
       inserted_at: Map.get(attrs, :inserted_at, DateTime.utc_now()),
-      updated_at: Map.get(attrs, :updated_at, DateTime.utc_now())
+      updated_at: Map.get(attrs, :updated_at, DateTime.utc_now()),
+      completed_at: Map.get(attrs, :completed_at, nil)
     }
   end
 
+  def running_execution_result_factory(_attrs) do
```
arbulu89, line 168: ExMachina provides better ways to do this.
Check this: https://github.com/thoughtbot/ex_machina#flexible-factories-with-pipes

We could have `with_running_state` function

test/results_test.exs
```diff
@@ -58,13 +68,118 @@ defmodule Wanda.ResultsTest do
...
+               |> length()
+    end
+  end
+
+  describe "Completing an Execution" do
+    test "should correctly complete a running execution" do
+      %ExecutionResult{
+        execution_id: execution_id,
+        group_id: group_id,
+        status: :running,
```
arbulu89, line 109: We can remove this `status` and `payload` here I guess. They are not used

test/support/factory.ex
```diff
@@ -150,15 +150,31 @@ defmodule Wanda.Factory do
     execution_id = Map.get(attrs, :execution_id, Faker.UUID.v4())
     group_id = Map.get(attrs, :group_id, Faker.UUID.v4())
 
+    payload =
```
arbulu89, line 153: We could use an empty payload by default and add a `with_payload` helper function to the factory: 
https://github.com/thoughtbot/ex_machina#flexible-factories-with-pipes

test/results_test.exs
```diff
@@ -58,13 +68,118 @@ defmodule Wanda.ResultsTest do
...
+                 build(
+                   :result,
+                   execution_id: execution_id,
+                   group_id: group_id,
+                   result: :passing
+                 )
+               )
+    end
+
+    test "should pass through an already completed execution" do
```
arbulu89, line 131: What is this test for? To check if already completed executions are not changed?
It looks way to complicated, doesn't it?
If we want to check that the payload is changed, I would do it in the previous test, which is the happy path

arbulu89 (changes requested):
Hey @nelsonkopliku ,
I like the general implementation.
Some nitpickts, and I think other needed improvements (specially on the factories) and in the enums

lib/wanda/results/execution_result.ex
```diff
@@ -5,15 +5,48 @@ defmodule Wanda.Results.ExecutionResult do
...
+    field :status, Ecto.Enum, values: [:running, :completed], default: :running
```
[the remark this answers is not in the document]
fabriziosestito: I propose to track this as debt and maybe create a library around that macro

lib/wanda/results/execution_result.ex
```diff
@@ -5,15 +5,48 @@ defmodule Wanda.Results.ExecutionResult do
...
+
+    embeds_many :targets, Target do
+      field :agent_id, Ecto.UUID
+      field :checks, {:array, :string}
+    end
 
     timestamps(type: :utc_datetime_usec)
+
+    field :completed_at, :utc_datetime_usec
```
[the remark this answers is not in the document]
fabriziosestito: Maybe we could use automatic timestamps without the ecto `timestamp`, so we can name it what we want.

lib/wanda/results.ex
```diff
@@ -5,25 +5,21 @@ defmodule Wanda.Results do
...
+  Create a new execution as soon as it starts.
   """
-  @spec create_execution_result!(Result.t()) :: ExecutionResult.t()
-  def create_execution_result!(
-        %Result{
-          execution_id: execution_id,
-          group_id: group_id
-        } = result
-      ) do
+  @spec create_execution_result!(String.t(), String.t(), [Target.t()]) :: ExecutionResult.t()
```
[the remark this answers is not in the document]
fabriziosestito: I would leave it as it is, one can infer the initial state of the function by reading to lines below. No need to specify it.

lib/wanda/results.ex
```diff
@@ -69,6 +65,26 @@ defmodule Wanda.Results do
     |> Repo.one()
   end
 
+  @doc """
+  Marks a previously started execution as completed
+  """
+  @spec complete_execution_result!(Result.t()) :: ExecutionResult.t()
+  def complete_execution_result!(%Result{execution_id: execution_id} = result) do
+    case get_execution_result!(execution_id) do
+      %ExecutionResult{
```
fabriziosestito, line 74: Do we need the `get_execution_result` and the case here? Isn't the update idempotent if we try to change the same field with `complete` several times? Also making a query before updating gives room for [time to check time to use bugs](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)

lib/wanda/results/execution_result.ex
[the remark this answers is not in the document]
fabriziosestito: +1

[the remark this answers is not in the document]
nelsonkopliku: Fair, gonna change.

lib/wanda/results/execution_result.ex
```diff
@@ -5,15 +5,48 @@ defmodule Wanda.Results.ExecutionResult do
...
+
+  @spec changeset(t() | Ecto.Changeset.t(), map) :: Ecto.Changeset.t()
+  def changeset(execution, params) do
+    execution
+    |> cast(params, __MODULE__.__schema__(:fields) -- [:targets])
+    |> cast_embed(:targets, with: &target_changeset/2)
+  end
+
+  @spec complete(t(), Result.t()) :: Ecto.Changeset.t()
+  def complete(execution, %Result{} = result) do
```
fabriziosestito, line 41: I am not sure about this. I have never found a schema that needs this kind of function since it already has the changeset. You can easily build a changeset in the context without having to rely to this one.

lib/wanda/results/execution_result.ex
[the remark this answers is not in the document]
nelsonkopliku: Do you mean just not using `timestamps(type: :utc_datetime_usec)` and define our own
```elixir
field :started_at, :utc_datetime_usec
field :completed_at, :utc_datetime_usec
```
I'd be ok with that, what about the `:updated_at`, would that still be valuable?

priv/repo/migrations/20220928131105_add_execution_result.exs
fabriziosestito: please doublecheck, but i think that what we are searching for is `add :targets, :jsonb, default: "[]"`

lib/wanda/results.ex
```diff
@@ -5,25 +5,21 @@ defmodule Wanda.Results do
...
-        } = result
-      ) do
+  @spec create_execution_result!(String.t(), String.t(), [Target.t()]) :: ExecutionResult.t()
+  def create_execution_result!(execution_id, group_id, targets) do
     Repo.insert!(%ExecutionResult{
       execution_id: execution_id,
       group_id: group_id,
-      payload: result
+      status: :running,
+      targets: Enum.map(targets, &Map.from_struct(&1))
```
fabriziosestito, line 22: I believe you don't need to map those if they have `Jason.encoder` protocol derive

lib/wanda/results.ex
fabriziosestito:
```elixir
%ExecutionResult{id: id}
|> ExecutionResult.changeset(%{
      payload: result,
      status: :completed,
      completed_at: DateTime.utc_now()
    })
|> Repo.update!()
```

fabriziosestito (changes requested):

test/execution/server_test.exs
fabriziosestito: +1

test/results_test.exs
```diff
@@ -4,35 +4,41 @@ defmodule Wanda.ResultsTest do
...
+                 %{
+                   agent_id: ^agent_id,
+                   checks: ^checks
+                 }
+               ]
              } = Repo.one!(ExecutionResult)
     end
 
-    test "should correctly add an item to a non empty log" do
+    test "should correctly create a running execution in a non empty system" do
```
fabriziosestito, line 41: i think this tests are going to far aka testing that the ecto create functionality work. We can remove this test, and rename the one above "should create an execution in a starting running state"

test/results_test.exs
fabriziosestito: + 1 on removing this test, we already know that if we update a row in the database with the same data, the operation is idempotent.

lib/wanda/results.ex
```diff
@@ -5,25 +5,21 @@ defmodule Wanda.Results do
...
 
-  alias Wanda.Execution.Result
+  alias Wanda.Execution.{Result, Target}
   alias Wanda.Results.ExecutionResult
 
   import Ecto.Query
 
   @doc """
-  Create a new result.
+  Create a new execution as soon as it starts.
```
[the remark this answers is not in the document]
nelsonkopliku: Alright, would just `Create a new execution.` be ok?

lib/wanda/results.ex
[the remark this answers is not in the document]
nelsonkopliku: At this point, if you're ok, I'd keep it as it is and defer to the refactor that we know will happen here.

lib/wanda/results/execution_result.ex
fabriziosestito: agreed on renaming this to `complete_changeset`

lib/wanda/results.ex
```diff
@@ -69,6 +65,26 @@ defmodule Wanda.Results do
...
+        status: :running
+      } = execution ->
+        execution
+        |> ExecutionResult.complete(result)
+        |> Repo.update!()
+
+      %ExecutionResult{
+        status: :completed
+      } = execution ->
+        execution
```
fabriziosestito, line 84: agreed on return ok or error if it was already completed

priv/repo/migrations/20220928131105_add_execution_result.exs
nelsonkopliku: I followed ecto doc here https://hexdocs.pm/ecto/Ecto.Schema.html#embeds_many/3

> It is recommended to declare your [embeds_many/3](https://hexdocs.pm/ecto/Ecto.Schema.html#embeds_many/3) field with type :map in your migrations, instead of using {:array, :map}. Ecto can work with both maps and arrays as the container for embeds (and in most databases maps are represented as JSON which allows Ecto to choose what works best).

I haven't tried the `add :targets, :jsonb, default: "[]"` gonna try it

test/results_test.exs
nelsonkopliku: It is not really empty, it is the second element of what we get from `insert_list(2, :execution_result)`, I am interested that there is a second element, yet I don't mind what it carries with it.

I can completely ignore it this way
```elixir
[
  %ExecutionResult{execution_id: execution_id, group_id: group_id},
  _
] = insert_list(2, :execution_result)
```

test/results_test.exs
nelsonkopliku: What I want to express here by pattern matching `status: :running` and `payload: %{}` is that we are starting with a running execution (status running and empty payload) and ending up with that same execution being completed and with some expected payload (later in the assertion).

lib/wanda/results.ex
nelsonkopliku: Adding `@derive Jason.Encoder` to `Wanda.Execution.Target` does not seem to be enough.

If I do that and then I pass `targets` here directly
```elixir
    Repo.insert!(%ExecutionResult{
      execution_id: execution_id,
      group_id: group_id,
      status: :running,
      targets: targets
    })
```
I get this error 
```
** (UndefinedFunctionError) function Wanda.Execution.Target.__changeset__/0 is undefined or private
```
I believe because `Wanda.Execution.Target` is not an ecto struct.

Alternatively I can use `ExecutionResult.changeset` and pass `targets` to it
```elixir
    %ExecutionResult{}
    |> ExecutionResult.changeset(%{
      execution_id: execution_id,
      group_id: group_id,
      status: :running,
      targets: targets
    })
    |> Repo.insert!()
```
yet it is not enough anyway because now ecto's changeset expects a map when casting `:targets` inside an `ExecutionResult` (received error `Ecto.CastError (expected params to be a :map, got: %Wanda.Execution.Target`)

At the end, to make it work I still needed a `Map.from_struct`, but in another place

```elixir
defmodule Wanda.Results.ExecutionResult do
  ...
  defp target_changeset(target, params) do
    cast(target, Map.from_struct(params), [:agent_id, :checks])
  end
  ...
end
```

So, for what I could experiment we either have `targets: Enum.map(targets, &Map.from_struct(&1))` in the context or `cast(target, Map.from_struct(params), [:agent_id, :checks])` in the changeset.

I don't see much value in creating another specific schema atm, let alone make `Wanda.Execution.Target` an ecto schema itself.

test/execution/server_test.exs
nelsonkopliku: Right, changing to
```elixir
assert %ExecutionResult{execution_id: ^execution_id, status: :running} =
               Repo.one!(ExecutionResult)
```

test/results_test.exs
nelsonkopliku: I am afraid that won't match as we have a list of `Wanda.Execution.Target` on one side and a list of simpler structs `%{}` on the other side.

priv/repo/migrations/20220928131105_add_execution_result.exs
fabriziosestito: I see, I believe this is not an `embeds_many` tho since we are not embedding schemas but we have a "free" json field

fabriziosestito: Forgive me, i totally miss the point the you were using an embeds_many, ignore my comment above

test/execution/server_test.exs
nelsonkopliku: here https://github.com/trento-project/wanda/pull/60/commits/9ec12528f08b35acadd9b31c91f865877e111630

lib/wanda/results.ex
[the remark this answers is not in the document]
nelsonkopliku: here https://github.com/trento-project/wanda/pull/60/commits/9ec12528f08b35acadd9b31c91f865877e111630

lib/wanda/results/execution_result.ex
nelsonkopliku: here https://github.com/trento-project/wanda/pull/60/commits/9ec12528f08b35acadd9b31c91f865877e111630

test/results_test.exs
nelsonkopliku: Good point, collapsed the two tests to just one https://github.com/trento-project/wanda/pull/60/commits/0587367f3124fd67218c92a5498440dcdf03c304

test/results_test.exs
nelsonkopliku: Adjusted here https://github.com/trento-project/wanda/pull/60/commits/a272e9ce8df9bd3dcb4de5e0ee618ffa3e67472e

lib/wanda/results.ex
nelsonkopliku: adjusted here https://github.com/trento-project/wanda/pull/60/commits/a272e9ce8df9bd3dcb4de5e0ee618ffa3e67472e

test/results_test.exs
arbulu89: The question was more about why we need it hehe
As far as I see it is only used to check later on that we have 2 rows stored, which is not really needed. You could simply check that you have one, and forget about this one, as it doesn't really add any value.
Anyway, it is a trivial thing, so don't bother that much

test/results_test.exs
arbulu89: But you already know this when you add `make_running`. You are simply testing your own helper methods and factories.

test/support/factory.ex
arbulu89: At the end these 2 factories are not needed, right?

arbulu89 (commented):
I kind of find myself unable to re-review, as I see that many of the decisions were taken offline, and I don't understand some of the reasons.
So I'm letting this to you two guys XD

lib/wanda_web/views/result_view.ex
```diff
@@ -15,7 +15,27 @@ defmodule WandaWeb.ResultView do
...
+    result
+    |> Map.put("started_at", started_at)
+    |> Map.put("execution_id", execution_id)
+    |> Map.put("group_id", group_id)
+    |> Map.put("result", extract_result(status, result))
+    |> Map.put("timeout", extract_timeout(status, result))
   end
+
+  # not really
+  defp extract_result(:running, _), do: :passing
```
arbulu89, line 36: Doesn't this look strange? We are having a false passing here

test/support/factory.ex
nelsonkopliku: No indeed, removed.

lib/wanda_web/views/result_view.ex
nelsonkopliku: Yes it does :smile: 
And I am not really proud of the previous chain of `Map.put` either :sweat_smile: 

The fact is that we need to consider the running/completed variants of the executions also on the API layer.

I didn't go too far in that direction 'cause it might grow the PR, so I was thinking of just deferring to the next PRs to keep the scope narrow.

Honestly, I would completely remove the last commit https://github.com/trento-project/wanda/pull/60/commits/6656527aa23dbe55a5463071b0249aef04f44da4 for now

test/results_test.exs
nelsonkopliku: Alright changed here https://github.com/trento-project/wanda/pull/60/commits/bde3ce30348124a8ad986edf41320737046bc695

lib/wanda_web/views/result_view.ex
nelsonkopliku: At the end, as agreed, I simplified the factories here https://github.com/trento-project/wanda/pull/60/commits/7838b7c00059585a32112f02f5225e9a8d8aa4ad, yet this file will stay like this at the moment and be refactored in following PRs when we address the API layer

nelsonkopliku:
@arbulu89 @fabriziosestito addressed all the addressable feedbacks.

lib/wanda/results.ex
```diff
@@ -69,6 +66,32 @@ defmodule Wanda.Results do
...
+  @doc """
+  Marks a previously started execution as completed
+  """
+  @spec complete_execution_result!(String.t(), Result.t()) ::
+          ExecutionResult.t()
+  def complete_execution_result!(execution_id, %Result{} = result) do
+    # from(e in ExecutionResult, where: e.execution_id == ^execution_id, select: e)
+    # |> Repo.update!()
+
+    # ExecutionResult.changeset(%ExecutionResult{execution_id: execution_id}, %{
```
rtorrero: I'm guessing this is a leftover :-)

lib/wanda/results/execution_result.ex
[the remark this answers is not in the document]
rtorrero: I'm 99% sure I'm missing something, but shouldn't `completed_at` be the same as `updated_at` when status is `completed`? Why do we need it then?

rtorrero (commented):
Good stuff @nelsonkopliku, thanks for the detailed description

test/support/factory.ex
```diff
@@ -146,16 +146,24 @@ defmodule Wanda.Factory do
...
+    %ExecutionResult{
+      execution_id: Faker.UUID.v4(),
+      group_id: Faker.UUID.v4(),
+      status: :running,
+      targets: [],
+      started_at: DateTime.utc_now()
+    }
+  end
 
+  def with_completed_status(
```
arbulu89, line 159: Nice man!
But it is not used anywhere XD

arbulu89 (commented):
Some comments yet

lib/wanda/results/execution_result.ex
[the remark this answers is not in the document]
nelsonkopliku: `updated_at` is not used anymore

lib/wanda/results/execution_result.ex
```diff
@@ -5,15 +5,48 @@ defmodule Wanda.Results.ExecutionResult do
...
+  @fields ~w(execution_id group_id payload status)a
+
+  @spec changeset(t() | Ecto.Changeset.t(), map) :: Ecto.Changeset.t()
+  def changeset(execution, params) do
+    execution
+    |> cast(params, @fields)
+    |> cast_embed(:targets, with: &target_changeset/2)
+  end
+
+  @spec complete_changeset(t(), Result.t()) :: Ecto.Changeset.t()
```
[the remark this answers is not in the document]
nelsonkopliku: removed

test/support/factory.ex
nelsonkopliku: lol, if it's not too harm let's keep it, as a sample and for probable future usage

lib/wanda/results/execution_result.ex
```diff
@@ -5,15 +5,48 @@ defmodule Wanda.Results.ExecutionResult do
...
+    embeds_many :targets, Target do
+      field :agent_id, Ecto.UUID
+      field :checks, {:array, :string}
+    end
+
+    timestamps(type: :utc_datetime_usec, inserted_at: :started_at, updated_at: false)
+    field :completed_at, :utc_datetime_usec
+  end
+
+  @fields ~w(execution_id group_id payload status)a
```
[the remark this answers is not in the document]
nelsonkopliku: moved on top

lib/wanda/results.ex
```diff
@@ -69,6 +66,22 @@ defmodule Wanda.Results do
...
   end
 
+  @doc """
+  Marks a previously started execution as completed
+  """
+  @spec complete_execution_result!(String.t(), Result.t()) ::
+          ExecutionResult.t()
+  def complete_execution_result!(execution_id, %Result{} = result) do
+    ExecutionResult
+    |> Repo.get!(execution_id)
```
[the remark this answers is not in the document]
nelsonkopliku: There was some issues in returning the actual updated ExecutionResult, it didn't have all the fields properly populated if not getting before :shrug:

fabriziosestito (approved):
LGTM

arbulu89 (approved):
Thanks!

rtorrero (commented):
Forgot to say: my 2 cents about the naming to throw some ideas for a future PR; maybe `ExecutionStatus` (though it can be confusing having already a `ExecutionState`) or maybe `ExecutionReport`. In any case, let's defer this.

PR #61, opened 2022-10-19
Load checks properly for the execution

arbulu89:
Small issue I have found in an strange scenario (when no checks are selected for the execution).
The checks were not being loaded properly, as the `uniq` function was executed still on target lists. We simply missed the `concat` function. It involved cases where 2 targets have different checks selected (we don't have this scenario now, but who knows XD)

PD: i have extracted the function to make it testable. Testing gen servers start with wanted variables is pretty much a difficult


lib/wanda/execution.ex
```diff
@@ -35,6 +29,14 @@ defmodule Wanda.Execution do
     end
   end
 
+  def get_checks_from_targets(targets) do
```
fabriziosestito, line 32: The fact that this one needs to be public to be testable, and that it needs an integration test, it is a smell that maybe does not belong to this module. Perhaps we could move the function, without the `Catalog.get_check/1` call to the `Target` module and just test the pure function till uniq without the catalog side effects. Consider also adding `Catalog.get_checks/1` if needed.

fabriziosestito (changes requested):

lib/wanda/execution.ex
```diff
@@ -35,6 +29,14 @@ defmodule Wanda.Execution do
...
+    targets
+    |> Enum.map(& &1.checks)
```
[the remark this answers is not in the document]
arbulu89: Damn, you are right. I would bet that I already tried this...
Most probably I let the file unsaved or similar...

lib/wanda/execution.ex
CDimonaco: I agree that we can move the code to the Target module in order to segregate the responsability better, and we can test it better and faster :D

test/execution/target_test.exs
```diff
@@ -0,0 +1,34 @@
+defmodule Wanda.Execution.TargetTest do
+  use ExUnit.Case
+
+  import Wanda.Factory
+
+  alias Wanda.Execution.Target
+
+  describe "target" do
+    test "should load checks properly when the targets have the same checks" do
```
fabriziosestito, line 9: super-mega-nitpick: i would rename this to "return", since there is not really a "load" side effect.

lib/wanda/execution.ex
```diff
@@ -7,16 +7,26 @@ defmodule Wanda.Execution do
...
+      |> Target.get_checks_from_targets()
+      |> Catalog.get_checks()
+
+    maybe_start_execution(execution_id, group_id, targets, checks, env, config)
+  end
+
+  @impl true
+  defdelegate receive_facts(execution_id, agent_id, facts), to: Server
+
+  defp maybe_start_execution(_, _, _, [], _, _), do: {:error, :no_checks_selected}
```
fabriziosestito, line 27: nice, in the future we could also check if `checks` exist (sorry for the pun)

lib/wanda/execution/target.ex
```diff
@@ -12,4 +12,11 @@ defmodule Wanda.Execution.Target do
           agent_id: String.t(),
           checks: [String.t()]
         }
+
+  @spec get_checks_from_targets([__MODULE__.t()]) :: [String.t()]
```
fabriziosestito, line 16: this should be `[t()]` without `__MODULE__`

fabriziosestito (changes requested):
Some nitpicks and green to go

test/execution/target_test.exs
arbulu89: too much copy/paste XD

lib/wanda/execution.ex
arbulu89: yes, you are right. Now, if a non existing check is given, the app fails, and the consumer process is restarted...
I will open a ticket to track it

PR #63, opened 2022-10-24
Detect already running execution for group_id

arbulu89:
Detect if an execution with a specific group id is already being executed. Using the `:global` registry should make this distributed as well. Raise an error in this case by now.
We can do different things with this error later on: stop current execution, requeue message...

Some refs on the name registration: https://hexdocs.pm/elixir/1.12/GenServer.html#module-name-registration
PD: Maybe there better solutions for this, but I had some fun learning the :global registry usage hehe

### DIFF
--- a/lib/wanda/execution.ex
+++ b/lib/wanda/execution.ex
@@ -27,21 +27,43 @@ defmodule Wanda.Execution do
   defp maybe_start_execution(_, _, _, [], _, _), do: {:error, :no_checks_selected}
 
   defp maybe_start_execution(execution_id, group_id, targets, checks, env, config) do
-    case DynamicSupervisor.start_child(
-           Supervisor,
-           {Server,
-            execution_id: execution_id,
-            group_id: group_id,
-            targets: targets,
-            checks: checks,
-            env: env,
-            config: config}
-         ) do
-      {:ok, _} ->
-        :ok
+    with false <- execution_for_group_id_already_running(group_id),
+         {:ok, _} <-
+           DynamicSupervisor.start_child(
+             Supervisor,
+             {Server,
+              execution_id: execution_id,
+              group_id: group_id,
+              targets: targets,
+              checks: checks,
+              env: env,
+              config: config}
+           ) do
+      :ok
+    else
+      true ->
+        {:error, :group_already_running}
 
       {:error, _} = error ->
         error
     end
+  end
+
+  defp execution_for_group_id_already_running(group_id) do
+    :global.registered_names()
+    |> Enum.flat_map(fn entry ->
+      case entry do
+        {Server, execution_id} -> [{Server, execution_id}]
+        _ -> []
+      end
+    end)
+    |> Enum.any?(fn name ->
+      %{group_id: g} =
+        name
+        |> :global.whereis_name()
+        |> :sys.get_state()
+
+      g == group_id
+    end)
   end
 end
--- a/lib/wanda/execution.ex
+++ b/lib/wanda/execution.ex
@@ -58,12 +58,14 @@ defmodule Wanda.Execution do
       end
     end)
     |> Enum.any?(fn name ->
-      %{group_id: g} =
+      group_id ==
         name
         |> :global.whereis_name()
-        |> :sys.get_state()
-
-      g == group_id
+        |> get_group_id()
     end)
+  end
+
+  defp get_group_id(name) do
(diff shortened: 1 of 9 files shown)


dottorblaster (dismissed):
Looks good, just a couple things then green light from me

lib/wanda/execution.ex
```diff
@@ -27,21 +27,43 @@ defmodule Wanda.Execution do
...
-            targets: targets,
-            checks: checks,
-            env: env,
-            config: config}
-         ) do
-      {:ok, _} ->
-        :ok
+    with false <- execution_for_group_id_already_running(group_id),
+         {:ok, _} <-
+           DynamicSupervisor.start_child(
```
fabriziosestito, line 32: Hey @arbulu89 good try, but this is actually not needed. 
By simply changing the execution name to the group_id in the execution server `via_tuple` function, the dynamic supervisor will automatically not start an existing child returning `{:error, {:already_started, pid}`.
So by matching that error in the old code case you could return :group_already_running or whatever we want.

https://github.com/trento-project/wanda/blob/b3dde2dd2da96c33ecd0d514aaa88a8b62b53252/lib/wanda/execution/server.ex#L40

you'll need to change `receive_facts` as it will take both group_id and execution_id and in the handler of the `receive_facts` you could discard execution_ids that are different from the running genserver.

fabriziosestito (changes requested):
A proposal on the name registration

lib/wanda/execution.ex
arbulu89: @fabriziosestito I thought of that beforehand and I don't really like it. The execution id is the unique identifier of the execution, and we use it to run other functions, like getting the facts from this specific id. If I replace it by the group_id, we could have 2 executions with the same executions ids. And if I combine them, we could still have strange scenarios like accepting the same execution_id with a different group_id as valid. Strange things.

I personally wouldn't go that path (and it requires as well, to be passing the group_id all around the execution messages, etc)

fabriziosestito: > we could have 2 executions with the same executions ids
we know that executions_id are uuid, so they are unique. when would 2 executions with the same execution_id happen?

lib/wanda/execution.ex
```diff
@@ -27,21 +27,43 @@ defmodule Wanda.Execution do
...
+      case entry do
+        {Server, execution_id} -> [{Server, execution_id}]
+        _ -> []
+      end
+    end)
+    |> Enum.any?(fn name ->
+      %{group_id: g} =
+        name
+        |> :global.whereis_name()
+        |> :sys.get_state()
```
fabriziosestito, line 64: We should never introspect the state  of an actor: https://www.erlang.org/doc/man/sys.html#get_state-2

See the green note, this function is there just for debugging purpose

arbulu89: Yes, I had already changed this locally

lib/wanda/execution.ex
```diff
@@ -58,12 +58,14 @@ defmodule Wanda.Execution do
...
     end)
     |> Enum.any?(fn name ->
-      %{group_id: g} =
+      group_id ==
         name
         |> :global.whereis_name()
-        |> :sys.get_state()
-
-      g == group_id
+        |> get_group_id()
```
fabriziosestito: This is unfortunately a bit "unsafe" since the GenServer could be down (e.g. it concurrently processed a message that causes its graceful stop or termination, like the latest receive facts message) and this could cause an exception.
Also, the server could be finding a slow database causing this call to hang, as it processes one message at a time.
That's why, even if this is better than the state introspection, I wouldn't go with this solution.

arbulu89: Ok man! Even though the risk to happen such a scenario is small, it is there, so let's avoid it. I already implemented the other version. I will update the contracts first.

lib/wanda/execution.ex
```diff
@@ -27,21 +27,43 @@ defmodule Wanda.Execution do
...
+              group_id: group_id,
+              targets: targets,
+              checks: checks,
+              env: env,
+              config: config}
+           ) do
+      :ok
+    else
+      true ->
+        {:error, :group_already_running}
```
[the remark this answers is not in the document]
fabriziosestito: @arbulu89 + 1 for this

fabriziosestito (approved):
nitpick on error atom, but LGTM
happy when CI is

PR #64, opened 2022-10-24
Abstract RabbitMQ processing logic

jamie-suse:
Use GenRMQ processor to abstract the processing logic in Wanda AMQP consumer, this allow us to write better tests and separate concerns.

### DIFF
--- a/lib/wanda/messaging/adapters/amqp/consumer.ex
+++ b/lib/wanda/messaging/adapters/amqp/consumer.ex
@@ -3,9 +3,6 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
   AMQP consumer.
   """
 
-  alias Trento.Contracts
-  alias Wanda.Policy
-
   @behaviour GenRMQ.Consumer
 
   require Logger
@@ -24,13 +21,10 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
   def consumer_tag, do: "wanda"
 
   @impl GenRMQ.Consumer
-  def handle_message(%GenRMQ.Message{payload: payload} = message) do
-    with {:ok, event} <- Contracts.from_event(payload),
-         :ok <- Policy.handle_event(event) do
-      GenRMQ.Consumer.ack(message)
-    else
-      {:error, reason} ->
-        handle_error(message, reason)
+  def handle_message(%GenRMQ.Message{payload: _payload} = message) do
+    case processor().process(message) do
+      :ok -> GenRMQ.Consumer.ack(message)
+      {:error, reason} -> handle_error(message, reason)
     end
   end
 
@@ -50,4 +44,7 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
       shutdown: 500
     }
   end
+
+  defp processor,
+    do: Application.fetch_env!(:wanda, Wanda.Messaging.Adapters.AMQP)[:processor]
 end
--- a/lib/wanda/messaging/adapters/amqp/consumer.ex
+++ b/lib/wanda/messaging/adapters/amqp/consumer.ex
@@ -21,7 +21,7 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
   def consumer_tag, do: "wanda"
 
   @impl GenRMQ.Consumer
-  def handle_message(%GenRMQ.Message{payload: _payload} = message) do
+  def handle_message(%GenRMQ.Message{} = message) do
     case processor().process(message) do
       :ok -> GenRMQ.Consumer.ack(message)
       {:error, reason} -> handle_error(message, reason)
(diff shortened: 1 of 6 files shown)


lib/wanda/messaging/adapters/amqp/consumer.ex
```diff
@@ -24,13 +21,10 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
...
 
   @impl GenRMQ.Consumer
-  def handle_message(%GenRMQ.Message{payload: payload} = message) do
-    with {:ok, event} <- Contracts.from_event(payload),
-         :ok <- Policy.handle_event(event) do
-      GenRMQ.Consumer.ack(message)
-    else
-      {:error, reason} ->
-        handle_error(message, reason)
+  def handle_message(%GenRMQ.Message{payload: _payload} = message) do
```
arbulu89, line 24: If we don't use the payload, we could have this simply as:
```
def handle_message(%GenRMQ.Message{} = message) do
```

lib/wanda/messaging/adapters/amqp/consumer.ex
```diff
@@ -24,13 +21,10 @@ defmodule Wanda.Messaging.Adapters.AMQP.Consumer do
...
+    case processor().process(message) do
+      :ok -> GenRMQ.Consumer.ack(message)
```
arbulu89, line 26: Usually the mix formatter puts the case implementation in a new line. Did you execute `mix format`?

lib/wanda/messaging/adapters/amqp/processor.ex
```diff
@@ -0,0 +1,19 @@
...
+  """
+
+  alias Trento.Contracts
+  alias Wanda.Policy
+
+  @behaviour GenRMQ.Processor
+
+  require Logger
+
+  def process(%GenRMQ.Message{payload: payload}) do
```
arbulu89, line 13: Maybe we want to print the whole message as debug here. 
```
def process(%GenRMQ.Message{payload: payload} = message) do
    Logger.debug("Received message: #{inspect(message)}")
...
```

lib/wanda/messaging/adapters/amqp/processor.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Wanda.Messaging.Adapters.AMQP.Processor do
+  @moduledoc """
+  AMQP processor.
+  """
+
+  alias Trento.Contracts
+  alias Wanda.Policy
+
+  @behaviour GenRMQ.Processor
```
arbulu89, line 9: I would put this `behaviour` in top of all `alias`

arbulu89 (changes requested):
It looks good in general.
I have noted some comments to improve how the code looks

lib/wanda/messaging/adapters/amqp/processor.ex
fabriziosestito: https://github.com/christopheradams/elixir_style_guide#module-attribute-ordering

lib/wanda/messaging/adapters/amqp/consumer.ex
jamie-suse: Interesting. I did run it before committing and ran it just now and it stayed the same 🤔 I guess this is a corner-case when the `case` is the outermost block in a function?

lib/wanda/messaging/adapters/amqp/processor.ex
jamie-suse: Changed as per style guide. Is it possible to configure the linter to check for this? I will look into it.

lib/wanda/messaging/adapters/amqp/consumer.ex
fabriziosestito: both are valid. If you add a new line just after the arrow it will format the second part of the blocks in the same way

test/messaging/adapters/amqp/consumer_test.exs
```diff
@@ -3,74 +3,40 @@ defmodule Wanda.Messaging.Adapters.AMQP.ConsumerTest do
...
-               |> Publisher.publish_message("executions")
+      assert :ok = Publisher.publish_message(message, "executions")
 
       assert_receive :consumed, 1_000
     end
+  end
 
-    test "should consume FactsGathered" do
+  describe "handle_error/1" do
+    test "should reject unknown events and move them to the dead letter queue" do
```
fabriziosestito, line 30: nitpick: we could write "messages" as the events is the domain word

lib/wanda/messaging/adapters/amqp/processor.ex
fabriziosestito: @jamie-suse it is possible, but doing that has some downside. Sometimes there are macros that need to have a different order, so in my experience is a bit hard to enforce.
You could try to activate the credo check  (it's in the "controversial" section) and see what happens.

fabriziosestito (approved):
LGTM

lib/wanda/messaging/adapters/amqp/consumer.ex
arbulu89: Okey!!

arbulu89 (approved):
Great Jamie!
Ready to go from my side

PR #65, opened 2022-11-02
Refactor context

fabriziosestito:
Refactor to remove the clash between result/execution/execution result

It moves the execution server API inside the server module, so we can expose the context and refactor the schema with a proper name.


lib/wanda_web/controllers/result_controller.ex
```diff
@@ -4,7 +4,7 @@ defmodule WandaWeb.ResultController do
 
   alias OpenApiSpex.Schema
```
arbulu89, line 5: Should we still continue calling this `result_controller`? Everything was moved to `Executions` so maybe `executions_controller`. And together with this the view code, and most probably the routes (`/api/checks/executions`)

lib/wanda_web/views/result_view.ex
```diff
@@ -16,11 +16,11 @@ defmodule WandaWeb.ResultView do
   end
 
   def render("result.json", %{
```
arbulu89, line 18: `execution.json` instead of `result.json`

arbulu89 (approved):
If the idea is to remove completely the `result` concept, we still have it in the code in some few places. I have noted some of them.

The worst thing is that it will open many conflicts with this hehe: https://github.com/trento-project/wanda/pull/63

lib/wanda/executions.ex
```diff
@@ -1,22 +1,25 @@
...
+    Target
+  }
 
   import Ecto.Query
 
   @doc """
   Create a new execution.
   """
-  @spec create_execution_result!(String.t(), String.t(), [Target.t()]) :: ExecutionResult.t()
+  @spec create_execution_result!(String.t(), String.t(), [Target.t()]) :: Execution.t()
```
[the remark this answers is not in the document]
fabriziosestito: grrr i missed it 
thanks

lib/wanda_web/controllers/result_controller.ex
fabriziosestito: @nelsonkopliku is taking care of this in another PR

nelsonkopliku (approved):
I saw this coming to life :smile: Looks good!

fabriziosestito:
let's wait to merge this once we have the API pr merged

arbulu89 (approved):

PR #66, opened 2022-11-02
Refactor executions api

nelsonkopliku:
# Description

This PR follows up #60 and adjusts the API layer now serving `Executions` rather than `Results`

- refactored `Results` references to `Executions`
- streamlined the execution schema to cover both `running` and `completed` executions
- polished the `ExecutionView` to map from an internal representation to a public one in a cleaner way

Extra
- added missing fields for changeset to consider (`started_at`, `completed_at`)

Complements #65, and there might be some conflicts in Controller/View components.

## How was this tested?

Current tests were updated accordingly.

## Samples
Following some samples of the API results.

Keep in mind that:
- we have `running` and `completed` executions
- these two variants come with the same basic shape
- the execution status, might give different meaning to properties of the structure

### Basic shape
```json
{
    "check_results": [], <- list, empty when running
    "completed_at": null, <- date of completion, null when running
    "execution_id": "da4468a3-1b2e-40f4-afb8-bedbef1d6b52", <- uuid, always available
    "group_id": "27c3b3e0-987f-45a4-881f-d07ad405359c", <- uuid, always available
    "result": "unknown", <- unknown when running, otherwise canonical `passing`, `warning`, `critical`
    "started_at": "2022-11-02T16:11:57.085165Z", <- starting time, always available
    "status": "running", <- running|completed
    "timeout": [] <- list, empty when running
}
```

### A running execution
```json
{
    "check_results": [], <- still not available
    "completed_at": null, <- still not completed
    "execution_id": "da4468a3-1b2e-40f4-afb8-bedbef1d6b52",
    "group_id": "27c3b3e0-987f-45a4-881f-d07ad405359c",
    "result": "unknown", <- still not available
    "started_at": "2022-11-02T16:11:57.085165Z",
    "status": "running",
    "timeout": [] <- still not available
}
```

### A completed execution
```json
{
    "check_results": [ <- set on completion
        {
            "agents_check_results": [
                {...}
            ],
            "check_id": "1AD4FR",
            "expectation_results": [
                {...}
            ],
            "result": "passing"
        }
    ],
    "completed_at": "2022-11-02T16:39:26.654000Z", <- set on completion
    "execution_id": "c143764a-2b3a-4e5e-b1e1-8bb776beb642",
    "group_id": "47906b47-0820-4772-844e-2c84136b89bb",
    "result": "passing", <- or `warning` or `critical`
    "started_at": "2022-11-02T16:39:26.654150Z",
    "status": "completed", <- new status set on completion
    "timeout": [] <- empty in case of no agent timing out, populated otherwise
}
```

### DIFF
--- a/lib/wanda_web/schemas/execution_response.ex
+++ b/lib/wanda_web/schemas/execution_response.ex
@@ -1,14 +1,14 @@
-defmodule WandaWeb.Schemas.ResultResponse do
+defmodule WandaWeb.Schemas.ExecutionResponse do
   @moduledoc nil
 
   require OpenApiSpex
 
   alias OpenApiSpex.Schema
 
-  alias WandaWeb.Schemas.ResultResponse.CheckResult
+  alias WandaWeb.Schemas.ExecutionResponse.CheckResult
 
   OpenApiSpex.schema(%{
-    title: "ResultResponse",
+    title: "ExecutionResponse",
     description: "The result of an execution",
     type: :object,
     properties: %{
--- a/lib/wanda_web/schemas/execution_response.ex
+++ b/lib/wanda_web/schemas/execution_response.ex
@@ -34,15 +34,17 @@ defmodule WandaWeb.Schemas.ExecutionResponse do
       },
       result: %Schema{
         type: :string,
-        enum: ["passing", "warning", "critical", "unknown"],
+        nullable: true,
+        enum: ["passing", "warning", "critical"],
         description: "Aggregated result of the execution, unknown for running ones"
       },
       timeout: %Schema{
         type: :array,
+        nullable: true,
         items: %Schema{type: :string, format: :uuid, description: "Agent ID"},
         description: "Timed out agents"
       },
-      check_results: %Schema{type: :array, items: CheckResult}
+      check_results: %Schema{type: :array, nullable: true, items: CheckResult}
     },
     required: [
       :execution_id,
--- a/lib/wanda_web/schemas/execution_response.ex
+++ b/lib/wanda_web/schemas/execution_response.ex
@@ -1,5 +1,7 @@
 defmodule WandaWeb.Schemas.ExecutionResponse do
-  @moduledoc nil
+  @moduledoc """
+  Execution item response API spec
+  """
 
   require OpenApiSpex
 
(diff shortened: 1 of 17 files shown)


fabriziosestito (changes requested):

lib/wanda_web/schemas/execution_response.ex
```diff
@@ -0,0 +1,58 @@
...
+        description: "Timed out agents"
+      },
+      check_results: %Schema{type: :array, items: CheckResult}
+    },
+    required: [
+      :execution_id,
+      :group_id,
+      :status,
+      :started_at,
+      :completed_at,
```
arbulu89, line 52: I'm not sure on open api `required` usage. Does this require to have a populdated field? I mean, for example `completed_at` and `timeout` can be empty values

lib/wanda_web/views/execution_view.ex
```diff
@@ -0,0 +1,48 @@
...
+      started_at: started_at,
+      completed_at: completed_at,
+      execution_id: execution_id,
+      group_id: group_id,
+      result: extract_result(status, result),
+      timeout: extract_timeout(status, result)
+    }
+  end
+
+  defp extract_result(:running, _), do: :unknown
```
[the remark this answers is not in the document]
arbulu89: By the way, I wouldn't use the `unknown` term, because it is not an unknown value. We don't have the result because we are still running, so I would put this value as nil maybe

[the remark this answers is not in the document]
arbulu89: And in other term, in many other places we have named these functions as `map_*` because at the end we are mapping to a new output

lib/wanda_web/schemas/execution_response.ex
```diff
@@ -0,0 +1,58 @@
...
+      },
+      completed_at: %Schema{
+        type: :string,
+        nullable: true,
+        format: :"date-time",
+        description: "Execution completion time"
+      },
+      result: %Schema{
+        type: :string,
+        enum: ["passing", "warning", "critical", "unknown"],
```
arbulu89, line 37: As put in other comment, I'm not convinced with the `unknown` value here

arbulu89 (changes requested):
I think we shouldn't use the `unknown` term if possible, because it is not a real unknown in fact. I would use `nil` values as suggested by Fabrizio in this cases.

lib/wanda_web/schemas/execution_response.ex
nelsonkopliku: Can be either empty lists or null, but they will always be there, for simplicity. Both in running and completed executions.
I can also remove something from the required list, what about the actual api then? would it return `completed_at: null` for a running execution or would it not return it at all?

fabriziosestito: return null for every not yet populated field works for me

lib/wanda_web/views/execution_view.ex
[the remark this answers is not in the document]
fabriziosestito: @arbulu89 @nelsonkopliku +1 for having null and not unknown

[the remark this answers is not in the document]
nelsonkopliku: - made fields nullable
- refactored functions to `map_*`

@fabriziosestito ok with revisiting as you're suggesting, let's keep it in the radar.

lib/wanda_web/schemas/execution_response.ex
nelsonkopliku: removed and made nullable

test/wanda_web/views/execution_view_test.exs
```diff
@@ -1,20 +1,20 @@
...
+defmodule WandaWeb.ExecutionViewTest do
   use WandaWeb.ConnCase, async: true
 
   import Phoenix.View
   import Wanda.Factory
 
   alias Wanda.Results.ExecutionResult
 
-  describe "ResultView" do
+  describe "ExecutionView" do
```
arbulu89, line 9: We could test this new logic about sending nil values in this conditions

arbulu89 (approved):
It look nice now!
PD: I don't know if the new logic of this nillable things is being tested

test/wanda_web/views/execution_view_test.exs
nelsonkopliku: Added in the last commit :wink:

fabriziosestito (approved):
LGTM

PR #68, opened 2022-11-07
Handle CORS in dev environment

arbulu89:
Enable CORS usage in dev mode so wanda can be queried from the browser.

PD: I'm not 100% if we can use mix env in the endpoint code. We use in the router in the web projcet, that's for sure


CDimonaco:
Maybe it's more safe to have some config value and retrieve them with get_env? So we can be sure that works.

Anyway another possibility is to configure the cors plug accordingly with the origins in the config file, so that the application can be configured for cors operations if needed. (If we are on the same domain and we don't want to enable cors we just don't configure it, or configure with the same domain as origin, this is a no-op, the browser on the same domain will not instantiate a cors preflight on the same domain) I feel like it's a zero cost feature, what do you think?

arbulu89:
> Maybe it's more safe to have some config value and retrieve them with get_env? So we can be sure that works.
> 
> Anyway another possibility is to configure the cors plug accordingly with the origins in the config file, so that the application can be configured for cors operations if needed. (If we are on the same domain and we don't want to enable cors we just don't configure it, or configure with the same domain as origin, this is a no-op, the browser on the same domain will not instantiate a cors preflight on the same domain) I feel like it's a zero cost feature, what do you think?

Hey, I have moved the cors plug to the router so it is only applied in the `api` endpoint. I think we are safe having it there and using the env variable, as we do it in other places.

About configuring, if you need so for your usecase, you can always add the value in the config file, but I'm not sure if need some hardcoded/default configuration at this point

nelsonkopliku (approved):
LGTM

PR #69, opened 2022-11-07
Change Abacus to Rhai 

fabriziosestito:
Refactored the evaluation functional core to use Rhai instead of abacus.
Added a couple of tests to show arrays and maps capabilities.
NOTE:
Once we move the repo to trento-project, we will probably rename `rhai_rustler` in something like `eval`, since we have already customized the rhai language's default behavior.


test/executions/evaluation_test.exs
```diff
@@ -628,6 +628,98 @@ defmodule Wanda.Executions.EvaluationTest do
     end
   end
 
+  describe "expressions with arrays" do
+    test "should return a passing result" do
+      check = Catalog.get_check("array_check")
```
arbulu89, line 633: We should't use this `Catalog.get_check` here. Use the factory instead.

arbulu89 (changes requested):
Hey @fabriziosestito ,
The usage looks great.
I wouldn't create new test fixtures to test the evaluation. Let's use the factories. We did an effort in the past to follow that path, so let's try to be consistent

arbulu89 (approved):

CDimonaco (approved):
lgtm

PR #74, opened 2022-11-14
Add cargo to dockerfile and build rustler

nelsonkopliku:
As per title adds cargo to the container and builds rustler.

@fabriziosestito remember me please what the issue was :sweat_smile:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 RUN zypper -n addrepo https://download.opensuse.org/repositories/devel:/languages:/erlang/SLE_15_SP3/devel:languages:erlang.repo
 RUN zypper -n --gpg-auto-import-keys ref -s
-RUN zypper -n in elixir
+RUN zypper -n in elixir rust cargo
 COPY . /build
 WORKDIR /build
 ARG MIX_ENV=prod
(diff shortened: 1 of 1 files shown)


fabriziosestito:
Related to: https://github.com/philss/rustler_precompiled/pull/40

Dockerfile
```diff
@@ -4,7 +4,7 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 RUN zypper -n addrepo https://download.opensuse.org/repositories/devel:/languages:/erlang/SLE_15_SP3/devel:languages:erlang.repo
 RUN zypper -n --gpg-auto-import-keys ref -s
-RUN zypper -n in elixir
+RUN zypper -n in elixir cargo
```
fabriziosestito, line 7: NITPICK: to be more explicit we are installing bot rust and cargo so: `rust cargo`

fabriziosestito (approved):

Dockerfile
nelsonkopliku: ok :+1:

Dockerfile
```diff
@@ -18,7 +18,7 @@ COPY --from=elixir-build /build /build
 WORKDIR /build
 ARG MIX_ENV=prod
 ENV MIX_ENV=$MIX_ENV
-RUN mix phx.digest
+RUN RHAI_RUSTLER_FORCE_BUILD=true mix phx.digest
```
fabriziosestito, line 21: note to self: we might want to consider making this part of application env in the future

PR #76, opened 2022-11-14
Integrate TLint into CI

dottorblaster:
Integrating TLint into the CI so no new checks hit `main` if not compliant

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,6 +50,21 @@ jobs:
           mix deps.compile --warnings-as-errors
           mix dialyzer --plt
 
+  tlint:
+    name: Run TLint over checks
+    runs-on: ubuntu-20.04
+    container:
+      image: ghcr.io/trento-project/tlint:latest
+      volumes:
+        - ${{ github.workspace }}:/data
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Run TLint
+        run: "/home/tlint/tlint lint -f /data/priv/catalog"
+
   static-code-analysis:
     name: Static Code Analysis
     needs: elixir-deps
--- a/priv/catalog/156F64.yaml
+++ b/priv/catalog/156F64.yaml
@@ -51,7 +51,7 @@ values:
       - value: 30000
         when: env.provider == "azure" || env.provider == "aws"
       - value: 20000
-        whens: env.provider == "gcp"
+        when: env.provider == "gcp"
 
 expectations:
   - name: timeout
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -50,6 +50,24 @@ jobs:
           mix deps.compile --warnings-as-errors
           mix dialyzer --plt
 
+  tlint:
+    name: Run TLint over checks
+    needs: elixir-deps
```
fabriziosestito, line 55: why does it need elixir-deps?

fabriziosestito (changes requested):
Mega LGTM in general, maybe a dependency too much in the CI step.

.github/workflows/ci.yaml
dottorblaster: You're right :+1:

fabriziosestito (approved):
green from my side

nelsonkopliku (approved):

arbulu89 (approved):
Nice!
Could we use external schemas to do the checking/
Wanda already has its own [schema](https://github.com/trento-project/wanda/blob/main/guides/check_definition.schema.json) in the guides folder.
This way, we don't depend on updating the linter project if we change the checks structure and the schema.

We could add an optional flag for that, and provider the file.
What do you think?

.github/workflows/ci.yaml
```diff
@@ -50,6 +50,23 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Run TLint
+        run: "/home/tlint/tlint lint -f /data/priv/catalog"
+      
+
```
arbulu89, line 69: I would remove one of the newlines by the way. 1 is enough XD

.github/workflows/ci.yaml
```diff
@@ -50,6 +50,23 @@ jobs:
...
+    container:
+      image: ghcr.io/trento-project/tlint:latest
+      volumes:
+        - ${{ github.workspace }}:/data
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
```
arbulu89, line 65: Remove newline here to be consistent with the rest of the file

.github/workflows/ci.yaml
fabriziosestito: ah yes the infamous newline

I've removed them from the whole file since the CI was really hard to read by using it in every step

dottorblaster:
@arbulu89 you are actually right, we have to do that using a build script, I tracked that as a ticket this morning :smile:

PR #77, opened 2022-11-15
Use strict module ordering

fabriziosestito:
Enables a credo check to enforce the ordering of the module parts and reactors the relevant modules.


arbulu89 (approved):

dottorblaster (commented):
Late to the party, LGTM, are we using this on web?

fabriziosestito:
> Late to the party, LGTM, are we using this on web?

@dottorblaster not yet but we will after tackling a couple of debts

PR #78, opened 2022-11-15
Enforce @spec usage for public functions

fabriziosestito:
Moar credo checks


arbulu89 (commented):
Well, we don't have that many failures XD

fabriziosestito:
> Well, we don't have that many failures XD

@arbulu89 I am having second thoughts about it, the check is very strict and often we don't really need a `@spec` (controllers, modules that implement a behaviour, web aggregates, test factories, generated files, etc..) so I am thinking to close this but I will suggest to use specs in the best practices of the docs repo

PR #83, opened 2022-11-15
Add coveralls

fabriziosestito:
Add `excoveralls` and coverage gh step


arbulu89 (approved):
Nice!
It's a pity that we don't see the coverage output in the action it self. 
Does it create some link to the execution in coveralls page?

dottorblaster (approved):
LGTM

@arbulu89 you're right, we might put the coveralls badge on the README

fabriziosestito:
> Nice! It's a pity that we don't see the coverage output in the action it self. Does it create some link to the execution in coveralls page?

@arbulu89 it should but I am struggling to make this work, I moved the PR from draft to ready to see if something was changing

PR #86, opened 2022-11-16
Check A1244C spec: Corosync consensus timeout is set to expected value

nelsonkopliku:
Adds Check A1244C Spec.


pirat013:
@nelsonkopliku I saw your PR today and figured out that we have modified the same check. I've added some more details in the remediation section and I'd suggest going forward with my  https://github.com/trento-project/wanda/pull/115 

What do you think?

fabriziosestito:
@pirat013 let's close this and keep the #115

PR #90, opened 2022-11-17
Add documentation for the package_version gatherer

rtorrero:
This PR simply adds a bit of documentation to the newly merged package_version gatherer.
- I've followed @arbulu89 suggestions on remove some unnecessary stuff (see pr #89)
- I've tweaked a bit the other corosync table to improve readability when not rendered somewhere without markdown support

### DIFF
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -105,7 +105,7 @@ For more information refer to [trento-project/agent/../gatherers/hostsfile_test.
 
 ### package_version
 
-This gatherer allows checking the installation of a certain package in the system
+This gatherer returns the version as a string of the specified package
 
 Sample arguments
 | Name                   | Return value  
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -110,17 +110,17 @@ This gatherer returns the version as a string of the specified package
 Sample arguments
 | Name                   | Return value  
 | ---------------------- | -------------------------------------------------------------------------------------
-| `package_name`         | a string containing the installed version for the package `package_name`, e.g "2.4.5"
+| `corosync`             | a string containing the installed version for the package `package_name`, e.g "2.4.5"
 
 Specification examples:
 
 ```yaml
 facts:
-  - name: package_version-corosync
+  - name: corosync_version
     gatherer: package_version
     argument: corosync
 
-  - name: package_version-pacemaker
+  - name: pacemaker_version
     gatherer: package_version
     argument: pacemaker
 
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -18,12 +18,13 @@ fact = gatherer(argument)
 
 Here's a collection of build-in gatherers, with information about how to use them.
 
-| name                             | implementation                                                                                                                                        |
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):

guides/gatherers.md
```diff
@@ -66,3 +67,30 @@ facts:
...
+Sample arguments
+| Name                   | Return value  
+| ---------------------- | -------------------------------------------------------------------------------------
+| `package_name`         | a string containing the installed version for the package `package_name`, e.g "2.4.5"
+
+Specification examples:
+
+```yaml
+facts:
+  - name: package_version-corosync
```
arbulu89, line 85: Come on man! Let's avoid having this `_-` thins things in the examples! XD
Put underscore only
They are ugly as hell and people take reference on them

guides/gatherers.md
```diff
@@ -66,3 +67,30 @@ facts:
 ```
 
 For extra information refer to [trento-project/agent/../gatherers/corosyncconf_test.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/corosyncconf_test.go)
+
+
+### package_version
+
+This gatherer allows checking the installation of a certain package in the system
```
arbulu89, line 74: I would rephrase to a more specific description.
`This gatherer returns the version as a string of the specified package` or similar.

guides/gatherers.md
```diff
@@ -66,3 +67,30 @@ facts:
...
+
+Sample arguments
```
arbulu89, line 76: I would sincerely remove this `sample arguments` entry. It is quite obvious how to use it in the further examples.

arbulu89 (changes requested):

guides/gatherers.md
rtorrero: It's not random! probably has a bit of OCD though :smile: .I want to make a clear difference between the name of the gatherer vs the argument we are giving

guides/gatherers.md
rtorrero: I agree that for this gatherer, it's quite obvious, but having consistency between all gatherers sections is important IMO. It barely makes a difference in space, and consistent sections generally allow for faster reading and finding what you want.

guides/gatherers.md
arbulu89: I would still change it. It is obvous that we are talking about this gatherer as the `gatherer` field already says so.
I would be fine if you change to `corosync_version`.

guides/gatherers.md
arbulu89: So put some real package name, if its an example.
I don't know, I find it so redundant...

guides/gatherers.md
fabriziosestito: +1 to remove the dash and just use the lowdash for consistency. `corosync_version` or `corosync_package_version` will do

fabriziosestito: @rtorrero I just realized it breaks the evaluation language as `-`  cannot appear in the variables: https://rhai.rs/book/ref/variables.html#valid-names

PR #94, opened 2022-11-18
Fix protobuf message mapping

arbulu89:
Implement the protobug complex types (list/structs) mapping.
I have doutbts whether the `from_` functions should return defined modules types instead of plain maps

### DIFF
--- a/lib/wanda/messaging/mapper.ex
+++ b/lib/wanda/messaging/mapper.ex
@@ -41,6 +41,12 @@ defmodule Wanda.Messaging.Mapper do
     )
   end
 
+  @spec from_execution_requested(ExecutionRequested.t()) :: %{
+          execution_id: String.t(),
+          group_id: String.t(),
+          targets: [Target.t()],
+          env: %{String.t() => boolean() | number() | String.t()}
+        }
   def from_execution_requested(%ExecutionRequested{
         execution_id: execution_id,
         group_id: group_id,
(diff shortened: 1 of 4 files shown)


lib/wanda/messaging/mapper.ex
```diff
@@ -33,13 +37,52 @@ defmodule Wanda.Messaging.Mapper do
...
+          targets: [Target.t()],
+          env: %{String.t() => boolean() | number() | String.t()}
+        }
+  def from_execution_requested(%ExecutionRequested{
+        execution_id: execution_id,
+        group_id: group_id,
+        targets: targets,
+        env: env
+      }) do
+    %{
```
arbulu89, line 56: I have doubts whether we should create a specific module type instead of returning plain map

test/messaging/mapper_test.exs
```diff
@@ -112,4 +114,151 @@ defmodule Wanda.Messaging.MapperTest do
...
+          },
+          %{
+            check_id: "check3",
+            name: "boolean_value",
+            fact_value: {:value, %{kind: {:boolean_value, true}}}
+          },
+          %{
+            check_id: "check4",
+            name: "list_value",
+            fact_value:
```
arbulu89, line 199: I didn't create any factory here, as these are protobuf related types, and are not part of our domain. They only have sense in this particular tests

fabriziosestito: super!

dottorblaster:
`protobug` :joy:

test/messaging/mapper_test.exs
dottorblaster: Fair enough

lib/wanda/messaging/mapper.ex
fabriziosestito: it's ok to use a map for this I wouldn't go as far as creating a struct for this since it's used internally and it doesn't go in or out our domain

fabriziosestito (approved):

nelsonkopliku (approved):
Top!

PR #96, opened 2022-11-18
Fix flaky evaluation test

arbulu89:
Fix a flaky evaluation test. It happens super rarely, but...
This is the error: https://github.com/trento-project/wanda/actions/runs/3496341979/jobs/5854146511
It happens because the list can be populated with duped start wars characters (too small dataset XD)
Adding the `Uniq` fixes the thing


dottorblaster (approved):
LGTM

PR #73, opened 2022-11-11
Start execution api

nelsonkopliku:

PR #101, opened 2022-11-22
Init before start

nelsonkopliku:
This is a leftover from https://github.com/trento-project/wanda/pull/73 and adds a call to `Wanda.Release.init()` before starting Wanda, allowing to have everything set up (e.g. migrations).

I noticed that even if we have wanda container depending on postgres one, at a certain point during startup an error occurs
```
Postgrex.Protocol (#PID<0.135.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (postgres:5432): connection refused - :econnrefused
```
However, wanda retries and it connects correctly immediately after, so I thought this being good enough.

If you think it is not enough we could leverage something like https://github.com/vishnubob/wait-for-it as mentioned in https://docs.docker.com/compose/startup-order/

### DIFF
--- a/docker-compose.checks.yaml
+++ b/docker-compose.checks.yaml
@@ -15,7 +15,7 @@ services:
       - rabbitmq
     ports:
       - 4000:4000
-    command: start
+    entrypoint: /bin/sh -c "/app/bin/wanda eval \"Wanda.Release.init()\" && /app/bin/wanda start"
     volumes:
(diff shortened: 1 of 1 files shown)


fabriziosestito (approved):
LGTM

> However, wanda retries and it connects correctly immediately after, so I thought this being good enough.

This is acceptable since it is expected that the application deals with the connection

PR #103, opened 2022-11-25
Add ability to handle non-existent & malformed Checks supplied to catalog

jamie-suse:
Currently, when the user provides a Check that does not exist, _or_ a malformed Check file,  when starting execution via the CLI, the server crashes with an exception.

Current example scenario:

```
iex(1)> execution_id = UUID.uuid4()
"60172c74-26b3-4e54-9870-232668abfb1a"
iex(2)> targets = [%Wanda.Executions.Target{agent_id: "99cf8a3a-48d6-57a4-b302-6e4482227ab6", checks: ["non_existent_check", "156F64"]}]
[
  %Wanda.Executions.Target{
    agent_id: "99cf8a3a-48d6-57a4-b302-6e4482227ab6",
    checks: ["non_existent_check", "156F64"]
  }
]
iex(3)> Wanda.Executions.Server.start_execution(execution_id, UUID.uuid4(), targets, %{})
** (YamlElixir.FileNotFoundError) Failed to open file "priv/catalog/non_existent_check.yaml": no such file or directory
    (yaml_elixir 2.9.0) lib/yaml_elixir.ex:22: YamlElixir.read_from_file!/2
    (wanda 0.1.0) lib/wanda/catalog.ex:35: Wanda.Catalog.get_check/1
    (elixir 1.13.4) lib/enum.ex:1593: Enum."-map/2-lists^map/1-0-"/2
    (wanda 0.1.0) lib/wanda/executions/server.ex:35: Wanda.Executions.Server.start_execution/5
```

And similarly for a malformed Check file:

```
iex(1)> execution_id = UUID.uuid4()
"180577b8-fe6c-4f05-be8e-3431e3a28e5f"
iex(2)> targets = [%Wanda.Executions.Target{agent_id: "99cf8a3a-48d6-57a4-b302-6e4482227ab6", checks: ["malformed_check", "non_existent_check", "156F64"]}]
[
  %Wanda.Executions.Target{
    agent_id: "99cf8a3a-48d6-57a4-b302-6e4482227ab6",
    checks: ["malformed_check", "non_existent_check", "156F64"]
  }
]
iex(3)> Wanda.Executions.Server.start_execution(execution_id, UUID.uuid4(), targets, %{})
** (FunctionClauseError) no function clause matching in Wanda.Catalog.map_check/1    
    
    The following arguments were given to Wanda.Catalog.map_check/1:
    
        # 1
        %{
          "description" => "A malformed check\n",
          "expectations" => [
            %{"expect_same" => "facts.jedi", "name" => "some_expectation"}
          ],
          "group" => "Test",
          "id" => "malformed_check",
          "name" => "Malformed check",
          "remediation" => "## Remediation\nRemediation text\n"
        }
    
    Attempted function clauses (showing 1 out of 1):
    
        defp map_check(%{
      "id" => id,
      "name" => name,
      "group" => group,
      "description" => description,
      "remediation" => remediation,
      "facts" => facts,
      "expectations" => expectations
    } = check)
    
    (wanda 0.1.0) lib/wanda/catalog.ex:51: Wanda.Catalog.map_check/1
    (elixir 1.13.4) lib/enum.ex:1593: Enum."-map/2-lists^map/1-0-"/2
    (wanda 0.1.0) lib/wanda/executions/server.ex:35: Wanda.Executions.Server.start_execution/5
```

This PR changes the behaviour so that when the catalogue is given a non-existent or malformed Check, it logs an error and continues processing the other checks in the list.

```
iex(1)> execution_id = UUID.uuid4()
"710b8bfd-f2b3-4eb2-b390-b6a3f56e0cad"
iex(2)> targets = [%Wanda.Executions.Target{agent_id: "99cf8a3a-48d6-57a4-b302-6e4482227ab6", checks: ["malformed_check", "non_existent_check", "156F64"]}]
[
  %Wanda.Executions.Target{
    agent_id: "99cf8a3a-48d6-57a4-b302-6e4482227ab6",
    checks: ["malformed_check", "non_existent_check", "156F64"]
  }
]
iex(3)> Wanda.Executions.Server.start_execution(execution_id, UUID.uuid4(), targets, %{})
[error] Check with ID malformed_check is malformed. Check if all the required fields are present.
[error] Error getting Check with ID non_existent_check: %YamlElixir.FileNotFoundError{message: "Failed to open file \"priv/catalog/non_existent_check.yaml\": no such file or directory"}
[debug] Starting execution: 710b8bfd-f2b3-4eb2-b390-b6a3f56e0cad
:ok
```

### DIFF
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -11,6 +11,8 @@ defmodule Wanda.Catalog do
     Value
   }
 
+  require Logger
+
   @default_severity :critical
 
   @doc """
@@ -22,18 +24,24 @@ defmodule Wanda.Catalog do
     |> Path.join("/*")
     |> Path.wildcard()
     |> Enum.map(&Path.basename(&1, ".yaml"))
-    |> Enum.map(&get_check(&1))
+    |> get_checks()
   end
 
   @doc """
   Get a check from the catalog.
   """
-  @spec get_check(String.t()) :: Check.t()
+  @spec get_check(String.t()) :: {:ok, Check.t()} | {:error, any}
   def get_check(check_id) do
-    get_catalog_path()
-    |> Path.join("#{check_id}.yaml")
-    |> YamlElixir.read_from_file!()
-    |> map_check()
+    case get_catalog_path()
+         |> Path.join("#{check_id}.yaml")
+         |> YamlElixir.read_from_file() do
+      {:ok, file_content} ->
+        {:ok, map_check(file_content)}
+
+      {:error, reason} = error ->
+        Logger.error("Error getting file with ID #{check_id}: #{inspect(reason)}")
+        error
+    end
   end
 
   @doc """
@@ -41,7 +49,12 @@ defmodule Wanda.Catalog do
   """
   @spec get_checks([String.t()]) :: [Check.t()]
   def get_checks(checks_id) do
-    Enum.map(checks_id, &get_check/1)
+    Enum.flat_map(checks_id, fn check_id ->
+      case get_check(check_id) do
+        {:ok, check} -> [check]
+        {:error, _} -> []
+      end
+    end)
   end
 
   defp get_catalog_path do
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -34,8 +34,7 @@ defmodule Wanda.Catalog do
   def get_check(check_id) do
     with path <- Path.join(get_catalog_path(), "#{check_id}.yaml"),
          {:ok, file_content} <- YamlElixir.read_from_file(path),
-         {:ok, check} <-
-           map_check(file_content) do
+         {:ok, check} <- map_check(file_content) do
       {:ok, check}
     else
       {:error, :malformed_check} = error ->
@@ -93,9 +92,7 @@ defmodule Wanda.Catalog do
      }}
   end
 
-  defp map_check(_) do
-    {:error, :malformed_check}
-  end
+  defp map_check(_), do: {:error, :malformed_check}
 
   defp map_severity(%{"severity" => "critical"}), do: :critical
   defp map_severity(%{"severity" => "warning"}), do: :warning
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -32,14 +32,21 @@ defmodule Wanda.Catalog do
   """
   @spec get_check(String.t()) :: {:ok, Check.t()} | {:error, any}
   def get_check(check_id) do
-    case get_catalog_path()
-         |> Path.join("#{check_id}.yaml")
-         |> YamlElixir.read_from_file() do
-      {:ok, file_content} ->
-        {:ok, map_check(file_content)}
+    with path <- Path.join(get_catalog_path(), "#{check_id}.yaml"),
+         {:ok, file_content} <- YamlElixir.read_from_file(path),
+         {:ok, check} <-
+           map_check(file_content) do
+      {:ok, check}
+    else
+      {:error, :malformed_check} = error ->
+        Logger.error(
+          "Check with ID #{check_id} is malformed. Check if all the required fields are present."
+        )
+
+        error
 
       {:error, reason} = error ->
-        Logger.error("Error getting file with ID #{check_id}: #{inspect(reason)}")
+        Logger.error("Error getting Check with ID #{check_id}: #{inspect(reason)}")
         error
     end
   end
@@ -72,17 +79,22 @@ defmodule Wanda.Catalog do
            "expectations" => expectations
          } = check
        ) do
-    %Check{
-      id: id,
-      name: name,
-      group: group,
-      description: description,
-      remediation: remediation,
-      severity: map_severity(check),
-      facts: Enum.map(facts, &map_fact/1),
-      values: map_values(check),
-      expectations: Enum.map(expectations, &map_expectation/1)
-    }
+    {:ok,
+     %Check{
+       id: id,
+       name: name,
+       group: group,
+       description: description,
+       remediation: remediation,
+       severity: map_severity(check),
+       facts: Enum.map(facts, &map_fact/1),
+       values: map_values(check),
+       expectations: Enum.map(expectations, &map_expectation/1)
+     }}
+  end
+
+  defp map_check(_) do
+    {:error, :malformed_check}
   end
 
   defp map_severity(%{"severity" => "critical"}), do: :critical
--- a/lib/wanda/executions/server.ex
+++ b/lib/wanda/executions/server.ex
@@ -38,7 +38,7 @@ defmodule Wanda.Executions.Server do
 
     targets =
       Enum.map(targets, fn %{checks: target_checks} = target ->
-        %Target{target | checks: intersection(target_checks, checks_ids)}
+        %Target{target | checks: Enum.filter(target_checks, fn check -> check in checks_ids end)}
       end)
 
     maybe_start_execution(execution_id, group_id, targets, checks, env, config)
@@ -151,8 +151,6 @@ defmodule Wanda.Executions.Server do
 
     {:stop, :normal, state}
   end
-
-  defp intersection(list_a, list_b), do: list_a -- list_a -- list_b
 
   defp continue_or_complete_execution(
          %State{
--- a/lib/wanda/executions/server.ex
+++ b/lib/wanda/executions/server.ex
@@ -34,6 +34,13 @@ defmodule Wanda.Executions.Server do
       |> Target.get_checks_from_targets()
       |> Catalog.get_checks()
 
+    checks_ids = Enum.map(checks, & &1.id)
+
+    targets =
+      Enum.map(targets, fn %{checks: target_checks} = target ->
+        %Target{target | checks: intersection(target_checks, checks_ids)}
+      end)
+
     maybe_start_execution(execution_id, group_id, targets, checks, env, config)
   end
 
@@ -144,6 +151,8 @@ defmodule Wanda.Executions.Server do
 
     {:stop, :normal, state}
   end
+
+  defp intersection(list_a, list_b), do: list_a -- list_a -- list_b
 
   defp continue_or_complete_execution(
          %State{
--- a/test/catalog_test.exs
+++ b/test/catalog_test.exs
@@ -34,77 +34,83 @@ defmodule Wanda.CatalogTest do
     end
 
     test "should load a check from a yaml file properly" do
-      assert %Check{
-               id: "expect_check",
-               name: "Test check",
-               group: "Test",
-               description: "Just a check\n",
-               remediation: "## Remediation\nRemediation text\n",
-               severity: :critical,
-               facts: [
-                 %Fact{
-                   name: "jedi",
-                   gatherer: "wandalorian",
-                   argument: "-o"
-                 },
-                 %Fact{
-                   name: "other_fact",
-                   gatherer: "no_args_gatherer",
-                   argument: ""
-                 }
-               ],
-               values: [
-                 %Value{
-                   conditions: [
-                     %Condition{expression: "some_expression", value: 10},
-                     %Condition{expression: "some_other_expression", value: 15}
-                   ],
-                   default: 5,
-                   name: "expected_value"
-                 },
-                 %Value{
-                   conditions: [
-                     %Condition{expression: "some_third_expression", value: 5}
-                   ],
-                   default: 10,
-                   name: "expected_higher_value"
-                 }
-               ],
-               expectations: [
-                 %Expectation{
-                   name: "some_expectation",
-                   type: :expect,
-                   expression: "facts.jedi == values.expected_value"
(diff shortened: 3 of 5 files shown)


lib/wanda/catalog.ex
```diff
@@ -22,26 +24,37 @@ defmodule Wanda.Catalog do
...
-    |> YamlElixir.read_from_file!()
-    |> map_check()
+    case get_catalog_path()
+         |> Path.join("#{check_id}.yaml")
+         |> YamlElixir.read_from_file() do
+      {:ok, file_content} ->
+        {:ok, map_check(file_content)}
+
+      {:error, reason} = error ->
+        Logger.error("Error getting file with ID #{check_id}: #{inspect(reason)}")
```
arbulu89, line 42: Let's update the message to `Error getting check with...`

lib/wanda/catalog.ex
```diff
@@ -22,26 +24,37 @@ defmodule Wanda.Catalog do
...
   def get_check(check_id) do
-    get_catalog_path()
-    |> Path.join("#{check_id}.yaml")
-    |> YamlElixir.read_from_file!()
-    |> map_check()
+    case get_catalog_path()
+         |> Path.join("#{check_id}.yaml")
+         |> YamlElixir.read_from_file() do
+      {:ok, file_content} ->
+        {:ok, map_check(file_content)}
```
arbulu89, line 39: Now I'm wondering if we should do some sanitiy check with `map_check`. I mean, this function will fail if it doesn't receive some expected fields (like `name` missing in check).
Maybe we should be resilient as well here.

It would be as simple as adding other case as `map_check(_)` which returns a `{:error, :malformed_check}` or similar, and tune `map_check` to return `{:ok, check}` otherwise.

arbulu89 (commented):
@jamie-suse The implementation and tests look great.
I have added a side note on making `map_check` more resilient against malformed check files.
I think it is pretty simple. Do you think we could include it in this PR?
Otherwise, I'm fine on merging this and creating a next PR with the update.

lib/wanda/catalog.ex
```diff
@@ -22,54 +24,70 @@ defmodule Wanda.Catalog do
...
-      name: name,
-      group: group,
-      description: description,
-      remediation: remediation,
-      severity: map_severity(check),
-      facts: Enum.map(facts, &map_fact/1),
-      values: map_values(check),
-      expectations: Enum.map(expectations, &map_expectation/1)
-    }
+  defp map_check(check) do
```
arbulu89, line 64: Let's use the power of pattern matching here!
So, restore back the `map_check` to the way it was, and simply create a new line after the original (this is important, to keep proper precendence of functions):
```
def map_check(_)  do
  Logger.error("Check with ID #{check_id} is malformed. Check if all the needed fields are present.")
  {:error, :malformed_check}
end
```

This way, if the 1st function doesn't have the expected arguments, it will go to the next.

test/catalog_test.exs
```diff
@@ -15,17 +15,18 @@ defmodule Wanda.CatalogTest do
     test "should return the whole catalog" do
       catalog_path = Application.fetch_env!(:wanda, Wanda.Catalog)[:catalog_path]
 
-      files =
+      valid_files =
         catalog_path
         |> File.ls!()
         |> Enum.sort()
+        |> Enum.filter(fn file -> file != "malformed_check.yaml" end)
```
arbulu89, line 22: You don't really need to do this. Actually, is nice to see that the catalog doesn't return malformed checks

test/catalog_test.exs
```diff
@@ -34,77 +35,87 @@ defmodule Wanda.CatalogTest do
...
-      assert %Check{severity: :warning} = Catalog.get_check("warning_severity_check")
+      assert {:ok, %Check{severity: :warning}} = Catalog.get_check("warning_severity_check")
+    end
+
+    test "should return an error for non-existent check" do
+      assert {:error, _} = Catalog.get_check("non_existent_check")
+    end
+
+    test "should return an error for malformed check" do
+      assert {:error, _} = Catalog.get_check("malformed_check")
```
arbulu89, line 113: Let's pattern match here the expected error, which is `{:error, :malformed_check}`

arbulu89 (changes requested):

test/catalog_test.exs
jamie-suse: We need to do this in order for the assertions in this test to be correct:

```Elixir
assert length(valid_files) == length(catalog)
```

and

```Elixir
Enum.with_index(catalog, fn check, index ->
        file_name =
          valid_files
          |> Enum.at(index)
          |> Path.basename(".yaml")

        assert %Check{id: ^file_name} = check
end)
```

lib/wanda/catalog.ex
```diff
@@ -22,26 +24,44 @@ defmodule Wanda.Catalog do
...
-  @spec get_check(String.t()) :: Check.t()
+  @spec get_check(String.t()) :: {:ok, Check.t()} | {:error, any}
   def get_check(check_id) do
-    get_catalog_path()
-    |> Path.join("#{check_id}.yaml")
-    |> YamlElixir.read_from_file!()
-    |> map_check()
+    with path <- Path.join(get_catalog_path(), "#{check_id}.yaml"),
+         {:ok, file_content} <- YamlElixir.read_from_file(path),
+         {:ok, check} <-
```
arbulu89, line 37: Such an ugly formatting hehe
Cannot we have `map_check` in the same line than `{:ok, check}`?

nelsonkopliku (changes requested):
I tried the timeout scenario and I got an error while the server was handling this.
```
** (MatchError) no match of right hand side value: nil
    (wanda 0.1.0) lib/wanda/executions/evaluation.ex:42
```
Thing is that `non_existent_check` remains in the `targets` piece of state of the execution and if that target timeouts, `non_existent_check` would still be taken into account for evaluation execution causing an error because it won't be found among the available checks, when queried for it. See [server.ex#L31](https://github.com/trento-project/wanda/blob/main/lib/wanda/executions/server.ex#L31), [server.ex#L140](https://github.com/trento-project/wanda/blob/main/lib/wanda/executions/server.ex#L140) and [evaluation.ex#L42](https://github.com/trento-project/wanda/blob/main/lib/wanda/executions/evaluation.ex#L42)

At execution startup we could simply _clean up_ targets state from non existing checks, if we do not use them.

test/catalog_test.exs
arbulu89: I see. Ok them. It looks ugly to have the name of the malformed check there hardcoded, but it is what it is hehe

arbulu89 (approved):
Ok @jamie-suse ,
Green light from my side.
Let's give some time to others to review if they want. We can merge next week otherwise

arbulu89:
> I tried the timeout scenario and I got an error while the server was handling this.
> 
> ```
> ** (MatchError) no match of right hand side value: nil
>     (wanda 0.1.0) lib/wanda/executions/evaluation.ex:42
> ```
> 
> Thing is that `non_existent_check` remains in the `targets` piece of state of the execution and if that target timeouts, `non_existent_check` would still be taken into account for evaluation execution causing an error because it won't be found among the available checks, when queried for it. See [server.ex#L31](https://github.com/trento-project/wanda/blob/main/lib/wanda/executions/server.ex#L31), [server.ex#L140](https://github.com/trento-project/wanda/blob/main/lib/wanda/executions/server.ex#L140) and [evaluation.ex#L42](https://github.com/trento-project/wanda/blob/main/lib/wanda/executions/evaluation.ex#L42)
> 
> At execution startup we could simply _clean up_ targets state from non existing checks, if we do not use them.

Really good catch @nelsonkopliku ,
We need to exclude the unknown check ids from the rest of the execution. 
Work for monday hehe

lib/wanda/catalog.ex
```diff
@@ -22,26 +24,44 @@ defmodule Wanda.Catalog do
...
+           map_check(file_content) do
+      {:ok, check}
+    else
+      {:error, :malformed_check} = error ->
+        Logger.error(
+          "Check with ID #{check_id} is malformed. Check if all the required fields are present."
+        )
+
+        error
+
```
fabriziosestito, line 47: we could be more precise about and bubble up a specific error when a check is not found.
By matching:

```elixir 
{:error,
 %YamlElixir.FileNotFoundError{ }}
```
you could return `{:error, :not_found}`

lib/wanda/catalog.ex
```diff
@@ -22,26 +24,44 @@ defmodule Wanda.Catalog do
...
+    end
   end
 
   @doc """
   Get specific checks from the catalog.
   """
   @spec get_checks([String.t()]) :: [Check.t()]
   def get_checks(checks_id) do
-    Enum.map(checks_id, &get_check/1)
+    Enum.flat_map(checks_id, fn check_id ->
```
fabriziosestito, line 59: try to rewrite this using a reduce, flat mapping is a bit hacky as we are not really returning an array with `get_check` function

fabriziosestito (changes requested):

lib/wanda/catalog.ex
fabriziosestito: on second thoughts, since the `get_checks` function is best-effort, couldn't we just return `nil` here if no check is found? @arbulu89 thoughts?

lib/wanda/catalog.ex
fabriziosestito: If we agree on this: https://github.com/trento-project/wanda/pull/103#discussion_r1032629906 we could also map and then reject nil

arbulu89: Well, this is Enum's documented way of mapping and filtering together.
```
Keep in mind that filter is not capable of filtering and transforming an element at the same time. If you would like to do so, consider using [flat_map/2](https://hexdocs.pm/elixir/1.12/Enum.html#flat_map/2).
```

lib/wanda/catalog.ex
arbulu89: What would be `something goes really wrong`? Can be have something more critical that the check doesn't exist hehe
I don't think this matches that match on Ecto's `get` function, as the `Repo.get` returning `nil` is a normal scenario (you don't match any row with the provided clause), but here we are in an exception. That's why I prefer the `ok, error` option.
Anyway, `get_check` is being used almost as a private function, so we could use either ways.

About being more precise in the error capturing and bubbling up, the same as before. We are using this function almost as a private, so nobody is using the error response for anything more than knowing that it is an error. But well, we can improve it of course for sake of completeness

CDimonaco: I'm with @arbulu89 , I think we need something specific to match in order to distinguish the "real" error

CDimonaco (approved):

fabriziosestito (approved):
Agreed with @arbulu89 to make this pass and work on improved error handing in another PR

lib/wanda/catalog.ex
arbulu89: As a comment, I had a discussion with @fabriziosestito , and he explained me some differences about recoverable and non recoverable errors, so we could actually make this function better.
But let's postpone it for future improvements.

nelsonkopliku (approved):
Looks good, thanks for taking care also of the timeout scenario!. 

Left just a couple of non blocking comments. 
One extra thing coming to my mind is if it would be valuable enhancing the tests for the timeout scenario to cover this case.

Up to you.

lib/wanda/executions/server.ex
```diff
@@ -34,6 +34,14 @@ defmodule Wanda.Executions.Server do
       |> Target.get_checks_from_targets()
       |> Catalog.get_checks()
 
+    checks_ids = Enum.map(checks, & &1.id)
+
+    # Remove non-existent checks from targets
+    targets =
+      Enum.map(targets, fn %{checks: target_checks} = target ->
+        %Target{target | checks: intersection(target_checks, checks_ids)}
```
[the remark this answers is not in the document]
jamie-suse: It looks a bit strange because the linter removes the brackets because of the right-associativity. Maybe it's clearer as
```
A -- (A -- B)
```

So you are first getting the difference between the two sets/lists, and then removing the difference from one of the sets to get the intersection of the two.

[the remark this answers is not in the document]
fabriziosestito: @jamie-suse in my opinion the idiomatic way to do this is to use `Enum.filter`, even if using `--` looks more elegant it looks a little bit more opaque and hard to read.
Also, it's a bit weird that this module contains a function that does generic list operations, while the `Enum` standard library module already provides a way to do this.

I would do what @nelsonkopliku  suggested, or if you prefer:
 ```elixir
Enum.filter(target_checks, fn check -> check in check_ids end)
```

fabriziosestito (changes requested):

lib/wanda/executions/server.ex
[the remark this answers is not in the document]
fabriziosestito: That said, @jamie-suse function is faster but I just find it hard to read, so I would vouch for clarity in this case as we have very small lists.

fabriziosestito (approved):

PR #104, opened 2022-11-25
Add critical, warning and passing counts to the execution view

fabriziosestito:
This PR adds critical, warning, and passing counts to the execution view.
These fields are needed to show aggregated counters in the FE.

Note:
I found the factory to be a bit "static". 
We need to extend the factory, adding checks results factory to use them in the test.

Since I am afraid that I will need to change most of the tests I am deferring it to a follow-up PR


lib/wanda_web/views/execution_view.ex
```diff
@@ -58,4 +61,19 @@ defmodule WandaWeb.ExecutionView do
 
   defp map_checks_results(:running, _), do: nil
   defp map_checks_results(:completed, %{"check_results" => check_results}), do: check_results
+
+  defp count_critical(:running, _), do: nil
+
+  defp count_critical(:completed, %{"check_results" => check_results}),
+    do: Enum.count(check_results, &(&1["result"] == "critical"))
```
dottorblaster, line 68: Could we generalize this to

```elixir
defp count(%{"checks_results" => check_results}, result),
  do: Enum.count(check_results, &(&1["result"] == result))
```

?

dottorblaster (approved):
Just a minor comment then LGTM I think

lib/wanda_web/views/execution_view.ex
arbulu89: It looks a good idea to me. It is an internal function, so we shouldn't expect incorrect usage, and in the worst case, we would get a 0 count value

arbulu89 (approved):

CDimonaco (approved):

lib/wanda_web/views/execution_view.ex
fabriziosestito: yes I agree, I totally missed this!

PR #105, opened 2022-11-28
Use specific compose ports for wanda dev/test docker-compose

fabriziosestito:
This PR assigns specific docker-compose forwarded ports to services so they don't clash with the web compose.

It also moves the development phoenix endpoint from 4000 to 4001.

This way, both web and Wanda can be started at the same time.

The checks dev docker-compose remain unchanged.

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -17,6 +17,15 @@ config :wanda, WandaWeb.Endpoint,
 # Do not print debug messages in production
 config :logger, level: :info
 
+config :wanda, Wanda.Messaging.Adapters.AMQP,
+  consumer: [
+    connection: "amqp://wanda:wanda@localhost:5672"
+  ],
+  publisher: [
+    connection: "amqp://wanda:wanda@localhost:5672"
+  ],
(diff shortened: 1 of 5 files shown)


CDimonaco (approved):

config/prod.exs
```diff
@@ -17,6 +17,15 @@ config :wanda, WandaWeb.Endpoint,
 # Do not print debug messages in production
 config :logger, level: :info
 
+config :wanda, Wanda.Messaging.Adapters.AMQP,
```
arbulu89, line 20: We don't need to change the prod env, right?
It is already inheriting this exact values from config

arbulu89 (approved):
Good. I think we don't have to add any change is prod, as it uses the same values as in config

config/prod.exs
fabriziosestito: yep exactly. I like to keep the config with "standard" values and the other env as an override of them.

arbulu89: But it is simply overriding the same values XD
So it doesn't do anything at the end.

PR #106, opened 2022-11-28
Add targets execution view

arbulu89:
As we were already storing the targets for each execution, this change exposes them in the executions view.
This is needed in order to know which are the targets and their checks on the FE execution view, when the execution is on running state.

image

### DIFF
--- a/lib/wanda/executions/execution.ex
+++ b/lib/wanda/executions/execution.ex
@@ -10,7 +10,6 @@ defmodule Wanda.Executions.Execution do
   @type t :: %__MODULE__{}
 
   @fields ~w(execution_id group_id result status started_at completed_at)a
-  @target_fields ~w(agent_id checks)a
 
   @derive {Jason.Encoder, [except: [:__meta__]]}
   @primary_key false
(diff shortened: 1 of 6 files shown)


lib/wanda/executions/execution.ex
```diff
@@ -20,10 +19,7 @@ defmodule Wanda.Executions.Execution do
     field :result, :map, default: %{}
     field :status, Ecto.Enum, values: [:running, :completed]
 
-    embeds_many :targets, Target do
-      field :agent_id, Ecto.UUID
-      field :checks, {:array, :string}
-    end
+    embeds_many :targets, Wanda.Executions.Execution.Target
```
arbulu89, line 22: The inline `embeds_many` doesn't allow to add the json derive..
And i couldn't make it work without extracting it to a independent module

dottorblaster (approved):
LGTM

lib/wanda/executions/execution.ex
```diff
@@ -21,6 +21,8 @@ defmodule Wanda.Executions.Execution do
     field :status, Ecto.Enum, values: [:running, :completed]
 
     embeds_many :targets, Target do
+      @derive {Jason.Encoder, [except: [:id]]}
```
fabriziosestito, line 24: TIL

fabriziosestito (approved):
LGTM :shipit:

nelsonkopliku (approved):

rtorrero (approved):
Thanks @arbulu89 , LGTM

CDimonaco (approved):

PR #108, opened 2022-11-28
Add last execution by group

fabriziosestito:
This PR adds an API endpoint to retrieve the last execution by `group_id`.
We need this in the FE to get the last executions in the components instead of fetching a list with an artificial pagination.

### DIFF
--- a/lib/wanda_web/router.ex
+++ b/lib/wanda_web/router.ex
@@ -12,6 +12,7 @@ defmodule WandaWeb.Router do
     pipe_through :api
 
     resources "/executions", ExecutionController, only: [:index, :show]
+    get "/groups/:id/executions/last", ExecutionController, :last
     post "/executions/start", ExecutionController, :start
     get "/catalog", CatalogController, :catalog
   end
--- a/test/wanda_web/controllers/execution_controller_test.exs
+++ b/test/wanda_web/controllers/execution_controller_test.exs
@@ -67,6 +67,26 @@ defmodule WandaWeb.ExecutionControllerTest do
     end
   end
 
+  describe "get last execution by group id" do
+    test "should return the last execution", %{conn: conn} do
+      %{group_id: group_id} = 10 |> insert_list(:execution) |> List.last()
+
+      json =
+        conn
+        |> get("/api/checks/groups/#{group_id}/executions/last")
+        |> json_response(200)
+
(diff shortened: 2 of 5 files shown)


CDimonaco (approved):

test/wanda_web/controllers/execution_controller_test.exs
```diff
@@ -67,6 +67,26 @@ defmodule WandaWeb.ExecutionControllerTest do
...
+        conn
+        |> get("/api/checks/groups/#{group_id}/executions/last")
+        |> json_response(200)
+
+      api_spec = ApiSpec.spec()
+      assert_schema(json, "ExecutionResponse", api_spec)
+    end
+
+    test "should return a 404", %{conn: conn} do
+      assert_error_sent 404, fn ->
```
arbulu89, line 84: Question. How does this even know that it should return 404?
Does elixir capture the specific error from Ecto and is able to know that is not found?

lib/wanda_web/router.ex
```diff
@@ -12,6 +12,7 @@ defmodule WandaWeb.Router do
     pipe_through :api
 
     resources "/executions", ExecutionController, only: [:index, :show]
+    get "/groups/:id/executions/last", ExecutionController, :last
```
arbulu89, line 15: I guess this removes the need to have the `group_id` filter in the `/executions` endpoint, as we would maybe better have `/groups/:id/executions`.
Anyway, let's not bother now. Both would do the same and 1st is already implemented

arbulu89 (approved):

nelsonkopliku (approved):

dottorblaster (approved):
LGTM

test/wanda_web/controllers/execution_controller_test.exs
fabriziosestito: `Ecto.NoResultsError` exception is caught by a Plug that displays a 404, this is provided by Phoenix.
In development you'll see the phoenix debug page for every exception including the 404,  in production you'll just get a 404 error view.

lib/wanda_web/router.ex
fabriziosestito: the group_id filter could be still  helpfpul when we'll do the historical result view

PR #109, opened 2022-12-01
Improve Execution open api doc

nelsonkopliku:
This PR aims to improve the api doc by adjusting `ChecksResult` to support `AgentCheckError`

### DIFF
--- a/lib/wanda_web/schemas/execution/agent_check_error.ex
+++ b/lib/wanda_web/schemas/execution/agent_check_error.ex
@@ -22,11 +22,12 @@ defmodule WandaWeb.Schemas.ExecutionResponse.AgentCheckError do
             Fact,
             FactError
           ]
-        }
+        },
+        nullable: true
       },
       type: %Schema{type: :string, description: "Error type"},
       message: %Schema{type: :string, description: "Error message"}
     },
-    required: [:agent_id, :facts, :type, :message]
+    required: [:agent_id, :type, :message]
   })
 end
--- a/lib/wanda_web/schemas/execution/agent_check_error.ex
+++ b/lib/wanda_web/schemas/execution/agent_check_error.ex
@@ -3,23 +3,30 @@ defmodule WandaWeb.Schemas.ExecutionResponse.AgentCheckError do
 
   alias OpenApiSpex.Schema
 
-  alias WandaWeb.Schemas.ExecutionResponse.{ExpectationEvaluation, Fact}
+  alias WandaWeb.Schemas.ExecutionResponse.{Fact, FactError}
 
   require OpenApiSpex
 
   OpenApiSpex.schema(%{
     title: "AgentCheckError",
-    description: "The result of check on a specific agent",
+    description:
+      "An error describing that some of the facts could not be gathered on a specific agent eg. gathering failure or timeout",
     type: :object,
     properties: %{
       agent_id: %Schema{type: :string, format: :uuid, description: "Agent ID"},
-      facts: %Schema{type: :array, items: Fact, description: "Facts gathered from the targets"},
-      expectation_evaluations: %Schema{
+      facts: %Schema{
+        description: "Facts gathered, possibly with errors, from the target",
         type: :array,
-        items: ExpectationEvaluation,
-        description: "Expectation evaluated during the check execution"
-      }
+        items: %Schema{
+          oneOf: [
+            Fact,
+            FactError
+          ]
+        }
+      },
+      type: %Schema{type: :string, description: "Error type"},
+      message: %Schema{type: :string, description: "Error message"}
     },
-    required: [:agent_id, :facts, :expectation_evaluations]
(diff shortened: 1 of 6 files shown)


test/support/factory.ex
```diff
@@ -166,6 +167,72 @@ defmodule Wanda.Factory do
     }
   end
 
+  def completed_with_errors(
```
fabriziosestito, line 170: @arbulu89 since you are taking care of the execution factory you might want to look at this

fabriziosestito (approved):

lib/wanda_web/schemas/execution/agent_check_error.ex
```diff
@@ -3,23 +3,30 @@ defmodule WandaWeb.Schemas.ExecutionResponse.AgentCheckError do
...
-        description: "Expectation evaluated during the check execution"
-      }
+        items: %Schema{
+          oneOf: [
+            Fact,
+            FactError
+          ]
+        }
+      },
+      type: %Schema{type: :string, description: "Error Type"},
```
arbulu89, line 27: I don't see this `title case` used anywhere. I would go with `Error type` in any case.

lib/wanda_web/schemas/execution/agent_check_error.ex
```diff
@@ -3,23 +3,30 @@ defmodule WandaWeb.Schemas.ExecutionResponse.AgentCheckError do
...
-    description: "The result of check on a specific agent",
+    description:
+      "An error describing that some of the facts could not be gathered on a specific agent",
     type: :object,
     properties: %{
       agent_id: %Schema{type: :string, format: :uuid, description: "Agent ID"},
-      facts: %Schema{type: :array, items: Fact, description: "Facts gathered from the targets"},
-      expectation_evaluations: %Schema{
+      facts: %Schema{
+        description: "Facts gathered, possibly with errors, from the target",
```
arbulu89, line 18: This facts list is nullable in our code

lib/wanda_web/schemas/execution/agent_check_error.ex
```diff
@@ -3,23 +3,30 @@ defmodule WandaWeb.Schemas.ExecutionResponse.AgentCheckError do
...
-  alias WandaWeb.Schemas.ExecutionResponse.{ExpectationEvaluation, Fact}
+  alias WandaWeb.Schemas.ExecutionResponse.{Fact, FactError}
 
   require OpenApiSpex
 
   OpenApiSpex.schema(%{
     title: "AgentCheckError",
-    description: "The result of check on a specific agent",
+    description:
+      "An error describing that some of the facts could not be gathered on a specific agent",
```
arbulu89, line 13: we could maybe extend the description to include the timeout scenario

lib/wanda_web/schemas/execution/fact_error.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  require OpenApiSpex
+
+  OpenApiSpex.schema(%{
+    title: "FactError",
+    description: "An error describing that a fact could not be gathered",
+    type: :object,
+    properties: %{
+      check_id: %Schema{type: :string, description: "Check ID"},
+      name: %Schema{type: :string, description: "Fact Name"},
```
arbulu89, line 14: The same with the title case

test/support/factory.ex
arbulu89: @nelsonkopliku I don't see this function as `completed_with_errors` as a transforming funciton.
Basically, it creates a totally new one from scratch, using simply the execution_id and group_id. 
I would simply have normal factory function for that.

test/support/factory.ex
```diff
@@ -166,6 +167,72 @@ defmodule Wanda.Factory do
...
+        %Execution{execution_id: execution_id, group_id: group_id} = execution
+      ) do
+    result = %Result{
+      execution_id: execution_id,
+      group_id: group_id,
+      check_results: [
+        %CheckResult{
+          check_id: "some_check_id",
+          expectation_results: [
+            %ExpectationResult{
```
arbulu89, line 180: Do we need to hardcode all of these entries. Why don't we use other factories to populate them?

test/support/factory.ex
```diff
@@ -166,6 +167,72 @@ defmodule Wanda.Factory do
...
+                  type: :expect
+                },
+                %ExpectationEvaluationError{
+                  name: "some_expectation_name",
+                  message: Faker.StarWars.quote(),
+                  type: :fact_missing_error
+                }
+              ]
+            },
+            %AgentCheckError{
```
arbulu89, line 202: It would be nice to have a new factory for `AgentCheckError` so we can reuse it here

lib/wanda_web/schemas/execution/agent_check_error.ex
nelsonkopliku: Interesting. Can you point out where this happens? I am not finding it :sweat_smile:

PR #111, opened 2022-12-02
Use Kernel.-- instead of Enum.filter

fabriziosestito:
Based on this: https://github.com/prodis/miss-elixir/blob/8b3a0a24114877aad4fe191fb0c60a0788827cb5/lib/miss/list.ex#L9

using @jamie-suse proposed solution to intersect the existing checks


arbulu89 (approved):
@fabriziosestito 's inner self having a battle during the night sleep XD
Thanks man!

PR #112, opened 2022-12-02
Refactor factories

fabriziosestito:
This PR refactor factories to be more composable in tests and adds few helpers to remove repeated code.

### DIFF
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -18,130 +18,77 @@ defmodule Wanda.Factory do
     Target
   }
 
-  def check_factory(attrs) do
+  def check_factory do
     %Catalog.Check{
-      id: Map.get(attrs, :id, UUID.uuid4()),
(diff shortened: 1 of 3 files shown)


test/support/factory.ex
```diff
@@ -150,19 +97,64 @@ defmodule Wanda.Factory do
       execution_id: Faker.UUID.v4(),
       group_id: Faker.UUID.v4(),
       status: :running,
-      targets: Enum.map(build_list(2, :target), &Map.from_struct(&1)),
+      targets: 1..5 |> Enum.random() |> build_list(:execution_target),
```
dottorblaster, line 100: Are we sure we wanna have this inline?

fabriziosestito: it seems readable enough to me

dottorblaster: then go :metal:

PR #133, opened 2022-12-15
Add a ref to target to README

dottorblaster:
Adding some targets info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,12 @@ In order to get detailed information for an execution, see [Getting Execution de
 > Please note that execution is _eventually started_, meaning that a successful response to the previous API call does not guarantee that the execution is running, but that it has been accepted by the system to start.
 
 #### Execution Targets
-When starting an execution a list of targets has to be specified. Each item in this list is an execution target. An execution target is a target host that will receive the payload, gather the facts informations about the requested checks, and send them back.
 
-Each target _must_ specify an `agent_id`. That can be obtained just issuing `trento-agent id` in a host's command line.
+An execution target is a target host where the checks are executed. This requires to have the `trento-agent` executable running in the host. In order to specify an execution order, its `agent_id` and a list of checks to be executed are provided. Once the execution is started, a facts gathering request is sent to these targets, facts are gathered and sent back to Wanda, where the checks result is evaluated using the gathered facts.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,6 +88,75 @@ In order to get detailed information for an execution, see [Getting Execution de
 
 > Please note that execution is _eventually started_, meaning that a successful response to the previous API call does not guarantee that the execution is running, but that it has been accepted by the system to start.
 
+#### Execution Targets
+When starting an execution a list of targets has to be specified. Each item in this list is an execution target. An execution target is a target host that will receive the payload, gather the facts informations about the requested checks, and send them back.
```
arbulu89, line 92: I find the order of the text a bit unsorted. I would start with the `An execution target is a target host...` phrase first, to describe what it is. And later on continue with the `When starting an execution..`

In fact I would rephrase to something like:
```
An execution target is a target host where the checks are executed. This requires to have the `trento-agent` executable running in the host. In order to specify an execution order, its `agent_id` and a list of checks to be executed are provided. Once the execution is started, a facts gathering request is sent to these targets, facts are gathered and sent back to Wanda, where the checks result is evaluated using the gathered facts.

The `agent_id` can be obtained just issuing `trento-agent id` command in the target host.

```

After this, you could explain the usage of agent_id and selected_checks

README.md
```diff
@@ -88,6 +88,75 @@ In order to get detailed information for an execution, see [Getting Execution de
...
+
+Each target _must_ specify an `agent_id`. That can be obtained just issuing `trento-agent id` in a host's command line.
+
+Each target _must_ specify a list of checks, that can be empty. These are the selected checks for each agent, that will be executed.
```
arbulu89, line 96: Let's use the present term.
`..that will be executed` -> `..that are executed`

README.md
```diff
@@ -88,6 +88,75 @@ In order to get detailed information for an execution, see [Getting Execution de
 
 > Please note that execution is _eventually started_, meaning that a successful response to the previous API call does not guarantee that the execution is running, but that it has been accepted by the system to start.
 
+#### Execution Targets
```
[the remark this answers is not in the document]
fabriziosestito: you might want to format this with prettier

arbulu89 (approved):
Thanks @dottorblaster ,
Ready to merge from my side

PR #136, opened 2022-12-16
Document cibadmin gatherer

arbulu89:
Document `cibadmin` gatherer.
I'm afraid that the cibadmin gatherer usage will bring many questions, but the cib itself is not an easy thing.
At the end, the gatherer simply returns a `map` version of the cibadmin xml format, that's all.

Depends on: https://github.com/trento-project/agent/pull/149

### DIFF
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -20,12 +20,46 @@ Here's a collection of build-in gatherers, with information about how to use the
 
 | name                             | implementation                                                                                                                                                 |
 | -------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| [`cibadmin`](#cibadmin)          | [trento-project/agent/../gatherers/cibadmin.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/cibadmin.go)    
 | [`corosync.conf`](#corosyncconf) | [trento-project/agent/../gatherers/corosyncconf.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/corosyncconf.go)          |
 | [`corosync-cmapctl`](#corosync-cmapctl) | [trento-project/agent/../gatherers/corosynccmapctl.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/corosynccmapctl.go)          |
 | [`hosts`](#hosts-etchosts)       | [trento-project/agent/../gatherers/hostsfile.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/hostsfile.go)                |
 | [`package_version`](#package_version) | [trento-project/agent/../gatherers/packageversion.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/packageversion.go) |
 | [`sbd_config`](#sbd_config)      | [trento-project/agent/../gatherers/sbd.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/sbd.go)                            |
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):
:eyes:

guides/gatherers.md
```diff
@@ -20,13 +20,47 @@ Here's a collection of build-in gatherers, with information about how to use the
...
+  - name: first_primitive
+    gatherer: cibadmin
+    argument: cib.configuration.resources.primitive.0
+
+  - name: second_cluster_property
+    gatherer: cibadmin
+    argument: cib.configuration.crm_config.cluster_property_set.nvpair.1
+```
+
+For extra information refer to [trento-project/agent/../gatherers/corosyncconf_test.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/corosyncconf_test.go)
```
rtorrero, line 62: I think you forgot to update the link here

guides/gatherers.md
```diff
@@ -20,13 +20,47 @@ Here's a collection of build-in gatherers, with information about how to use the
...
 | [`package_version`](#package_version) | [trento-project/agent/../gatherers/packageversion.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/packageversion.go) |
 | [`sbd_config`](#sbd_config)      | [trento-project/agent/../gatherers/sbd.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/sbd.go)                            |
 | [`systemd`](#systemd)            | [trento-project/agent/../gatherers/systemd.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/systemd.go)                    |
 
+### cibadmin
+
+This gatherer allows accessing Pacemaker's CIB information, the output of the `cibadmin` command more precisely.
+As the `cibadmin` command output is in XML format, the gatherer converts it to map/dictionary type of format, so the fields are available with the normal dot access way.
+Some specific fields, such as `primitive`, `clone`, `master`, etc (find the complete list [here](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/cibadmin.go#L48)) are converted to lists, in order to avoid differences when the field appears one or multiple times.
```
rtorrero, line 35: Just a suggestion, but consider using the footnote markdown syntax for this, might look better

guides/gatherers.md
arbulu89: Thanks!

guides/gatherers.md
arbulu89: Not really sure about this...
The footnote goes to the end of the page, and it looks a bit disconnected. If we started using this, I would use it consistenly in other places, not only here
I would go with the simple/straightforward solution.
What do you think?

rtorrero: ah... true, I looked at the unrendered markdown and it seemed it was possible to do it per section, but when you render it, it goes indeed to the bottom. Nevermind, if it's not possible then this is good enough

rtorrero (approved):
LGTM

guides/gatherers.md
rtorrero: Ah, I see you added it. It doesn't look bad, if you like it and want to keep it, maybe add  something like:
```[^1]: Find the complete list of supported cibadmin fields [here](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/cibadmin.go#L48)```

So that it also keeps it's context when read separately.

In any case, just a suggestion :+1:

arbulu89: At the end I have reverted the change @rtorrero , I don't really like the big disconnection between the sections

PR #143, opened 2022-12-22
Add a cheat sheet for Rhai

dottorblaster:
As above, just an attempt to use new ExDocs cheatsheet feature to generate the one we need.

### DIFF
--- a/guides/rhai_expressions_cheat_sheet.cheatmd
+++ b/guides/rhai_expressions_cheat_sheet.cheatmd
@@ -1,0 +1,81 @@
+# Rhai expressions cheatsheet
+
+In this cheatsheet are grouped the most frequent manipulations that can be done through Rhai, for convenience.
+
+## Cheatsheet
+{: .col-2}
+
+### Arrays
+
+#### Filtering an array
+
+```rust
+[1, 2, 3].filter(|value| value % 2 == 0)
+=> [2]
+```
+
+#### Finding the index of an element inside an array
+
+```rust
+["wow", "check"].index_of(|value| value == "check")
+=> 1
+```
+
+#### Finding an element inside an array
+
+```rust
+let nodelist = [
+  #{resource_id: 5, name: "foo"},
+  #{resource_id: 12, name: "bar"}
+];
+
+nodelist.find(|value| value.resource_id == 5)
+=> #{"name": "foo", "resource_id": 5}
+```
+
+#### Checking if an expression is true for every element of an array 
+
+```rust
+[2, 4, 6].all(|value| value % 2 == 0)
+=> true
+```
+
+### Maps
+
+#### Get only keys (returns an array)
+
+```rust
+#{a: 1, b: 2}.keys()
+=> ["a", "b"]
+```
+
+#### Get only values (returns an array)
+
+```rust
+#{a: 1, b: 2}.values()
+=> [1, 2]
+```
+
+#### Check if a map has more than 5 keys which name starts with "ring"
+
+```rust
+let map = #{
+  ring_one: 12,
+  ring_two: 34,
+  ring_three: 90,
+  ring_four: 234,
+  ring_five: 908
+};
+
+map.keys().filter(|prop| prop.starts_with("ring")).len() >= 5
+=> true
+```
+
+### Strings
+
+#### Splitting a string
+
+```rust
+"a;b;c".split(";")
+=> ["a", "b", "c"]
+```
--- a/mix.exs
+++ b/mix.exs
@@ -58,7 +58,8 @@ defmodule Wanda.MixProject do
       "CONTRIBUTING.md",
       "guides/specification.md",
       "guides/expression_language.md",
-      "guides/gatherers.md"
+      "guides/gatherers.md",
+      "guides/rhai_expressions_cheat_sheet.cheatmd"
     ]
   end
 


guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
+# Rhai expressions cheatsheet
+
+We are grouping into this cheatsheet the most frequent manipulations that can be done through Rhai, for convenience.
+
+## Cheatsheet
+{: .col-2}
```
rtorrero, line 6: I think this is not rendering properly in my markdown renderer, is this expected?

rtorrero (approved):
@dottorblaster Looks good to me! the only comment I left is also related to the question I have: where are we going to place this and in which form?

dottorblaster:
@rtorrero the syntax is related to the `cheatmd` format.

This is the final result:

image

guides/rhai_expressions_cheat_sheet.cheatmd
dottorblaster: Yes :smile:

rtorrero:
@dottorblaster that looks perfect, thanks!

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+#### Checking if an expression is true for every element of a list
+
+```rust
+[2, 4, 6].all(|value| value % 2 == 0)
+=> true
+```
+
+### Maps
+
+#### Get only keys (in a list)
```
gereonvey, line 33: This should probably be "Get only keys (in a map)" not "in a list"

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+
+```rust
+#{a: 1, b: 2}.keys()
+=> ["a", "b"]
+```
+
+#### Get only values (in a list)
```
gereonvey, line 40: This should probably be "Get only keys (in a map)" not "in a list"

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+
+### Lists
+
+#### Filtering a list
+
+```rust
+[1, 2, 3].filter(|value| value % 2 == 0)
+=> [2]
+```
+
```
gereonvey, line 16: Can we add links to the Rhai documentation for each of these? Would be handy in case you run into problems. For an experienced programmer it's probably fine to guess the structure/syntax of these, but for someone not that experienced a pointer to the documentation would be great.

guides/rhai_expressions_cheat_sheet.cheatmd
dottorblaster: "in a list" means that it returns a list, maybe I should be more clear about that like `returns a list`

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
+# Rhai expressions cheatsheet
+
+We are grouping into this cheatsheet the most frequent manipulations that can be done through Rhai, for convenience.
```
fabriziosestito, line 3: Don't use the "we" as per the SUSE documentation style guide

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+
+## Cheatsheet
+{: .col-2}
+
+### Lists
```
fabriziosestito, line 8: We should use the name Array to be consistent with rhai documentation

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+
+#### Filtering a list
```
fabriziosestito, line 10: array

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+
+```rust
+[1, 2, 3].filter(|value| value % 2 == 0)
+=> [2]
+```
+
+#### Finding an element inside a list
```
fabriziosestito, line 17: array

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+
+```rust
+repl> ["wow", "check"].index_of(|value| value == "check")
+=> 1
+```
+
+#### Checking if an expression is true for every element of a list
+
```
fabriziosestito, line 25: array

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+```
+
+#### Get only values (in a list)
+
+```rust
+#{a: 1, b: 2}.values()
+=> [1, 2]
+```
+
+### Generic operations
```
fabriziosestito, line 47: Why generics? Those are "Strings"

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,54 @@
...
+
+```rust
+[1, 2, 3].filter(|value| value % 2 == 0)
+=> [2]
+```
+
+#### Finding an element inside a list
+
+```rust
+repl> ["wow", "check"].index_of(|value| value == "check")
```
fabriziosestito, line 20: This finds the index, not the element.
I would keep this by changing the title and adding another cheat describing how to get an element by a property in a list of maps.

`nodelist[nodelist.index_of(|x| x.nodeId == 1)]`
where `nodelist = [#{nodeId: 1}, ....]`

fabriziosestito (changes requested):

guides/rhai_expressions_cheat_sheet.cheatmd
angelabriel: index_of() returns -1, if the element searched by '==' is not available. And this means, that the 'cheat' will return the last element of the list/array, which is NOT the expected result.

dottorblaster: Luckily `.find()` by @fabriziosestito got merged in the last release

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,81 @@
...
+=> 1
+```
+
+#### Finding an element inside an array
+
+```rust
+let nodelist = [
+  #{resource_id: 5, name: "foo"},
+  #{resource_id: 12, name: "bar"}
+]
```
fabriziosestito, line 30: this is not valid rhai you are missing the ';' to terminate the line

guides/rhai_expressions_cheat_sheet.cheatmd
```diff
@@ -0,0 +1,81 @@
...
+#### Check if a map has more than 5 keys which name starts with "ring"
+
+```rust
+let map = #{
+  ring_one: 12,
+  ring_two: 34,
+  ring_three: 90,
+  ring_four: 234,
+  ring_five: 908
+}
```
fabriziosestito, line 68: missing ';'

fabriziosestito (changes requested):
I found some problems with the rhai syntax, please check if the cheats are working on:
https://rhai.rs/playground/stable/

angelabriel:
> I found some problems with the rhai syntax, please check if the cheats are working on: https://rhai.rs/playground/stable/

at least the examples using `find` will not work in the playground as the playground is compiled with rhai 1.7 and not with 1.12

nelsonkopliku (approved):

fabriziosestito (approved):

PR #147, opened 2023-01-02
Add documentation for verify_password gatherer

rtorrero:
This PR adds an entry for the verify_password gatherer in the gatherers.md file.

### DIFF
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -387,17 +387,20 @@ For extra information refer to [trento-project/agent/../gatherers/systemd_test.g
 
 **Argument required**: yes.
 
-This gatherer uses `getent` to determine whether a password from a certain user needs to be changed or not. It returns `true` if the password should be changed or `false` if not.
+This gatherer determines whether a given user has its password still set to the (unsafe) default one. It returns `true` if it needs to be changed to something or `false` if it has already been changed.
 
 Specification examples:
 ```yaml
 facts:
-  - name: hacluster
+  - name: hacluster_has_default_password
     gatherer: verify_password
     argument: hacluster
 ```
 
 For the argument, only whitelisted user are allowed. Currently whitelisted usernames:
  - `hacluster`
+
+The default passwords that we check against, are:
+ - `linux`
   
 For extra information refer to [trento-project/agent/../gatherers/verifypassword_test.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/verifypassword_test.go)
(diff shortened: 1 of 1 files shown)


guides/gatherers.md
```diff
@@ -381,3 +382,22 @@ facts:
...
+### verify_password
+
+**Argument required**: yes.
+
+This gatherer uses `getent` to determine whether a password from a certain user needs to be changed or not. It returns `true` if the password should be changed or `false` if not.
+
+Specification examples:
+```yaml
+facts:
+  - name: hacluster
```
nelsonkopliku, line 395: What about something like `name: hacluster_well_known_password_detected`? To me, it would better fit with what the gatherer is actually doing.

guides/gatherers.md
```diff
@@ -381,3 +382,22 @@ facts:
...
+    gatherer: verify_password
```
nelsonkopliku, line 396: Given that the responsibility of this gatherer is to tell whether a well known password is detected for a user or not (and it won't change much afaiu), I'd see fit also a more specific name like `verify_known_password` or similar, rather than`verify_password`, which I find a bit distant from what is actually happening.

How would that work?

nelsonkopliku (changes requested):
Nice job.
Just a couple of suggestions about the content and provided example.

guides/gatherers.md
```diff
@@ -381,3 +382,22 @@ facts:
 ```
 
 For extra information refer to [trento-project/agent/../gatherers/systemd_test.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/systemd_test.go)
+
+### verify_password
+
+**Argument required**: yes.
+
+This gatherer uses `getent` to determine whether a password from a certain user needs to be changed or not. It returns `true` if the password should be changed or `false` if not.
```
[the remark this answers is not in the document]
rtorrero: I see you point; I initially wrote it more along those lines but I thought that we could quickly realize that the predefined passwords are not enough and we need to run additional checks. But well, we can always update the documentation later.

guides/gatherers.md
rtorrero: I see your point. I don't exactly like the "well known" term as it sounds like something positive where it's not. I'll change it a bit but I get the point :+1:

guides/gatherers.md
rtorrero: I wouldn't change this. Even if we don't plan to change in it the neart future, this gatherer could be easily improved to add additional verifications for the password (i.e. "is this password based on dictionary words?" or "is this password too short?", ...). It would be a pain in that case to update all checks to match this.

wdyt?

nelsonkopliku: Naming is always a pandora box :smile: no worries, keep it as you wish.

guides/gatherers.md
```diff
@@ -381,3 +382,22 @@ facts:
...
+Specification examples:
+```yaml
+facts:
+  - name: hacluster_has_default_password
+    gatherer: verify_password
+    argument: hacluster
+```
+
+For the argument, only whitelisted user are allowed. Currently whitelisted usernames:
+ - `hacluster`
```
nelsonkopliku, line 401: What about adding also the password against which we are checking?

nelsonkopliku (approved):
LGTM

PR #153, opened 2023-01-05
rewrite trento community check regarding 'hacluster' password change

angelabriel:
fixing jsc#CFSA-1096, check 1.5.2/790926

### DIFF
--- a/priv/catalog/790926.yaml
+++ b/priv/catalog/790926.yaml
@@ -1,0 +1,44 @@
+id: "790926"
+name: hacluster password
+group: Miscellaneous
+description: |
+  The hacluster user password has been changed from the default value
+remediation: |
+  ## Abstract
+  The password of the `hacluster` user should be changed after setting up the cluster
+
+  ## Remediation
+  ```sudo passwd hacluster```
+
+  ## References
+  Azure:
+
+    - https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker#install-the-cluster
+
+  AWS:
+
+    - https://docs.aws.amazon.com/sap/latest/sap-hana/sap-hana-on-aws-cluster-configuration.html
+
+  GCP:
+
+    - https://cloud.google.com/solutions/sap/docs/netweaver-ha-config-sles
+
+  Nutanix:
+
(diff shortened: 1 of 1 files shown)


priv/catalog/790926.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+facts:
+  - name: hacluster_passwd_changed
+    gatherer: verify_password
+    argument: hacluster
+
+expectations:
+  - name: expectations_hacluster_passwd_changed
+    expect: |
+      !facts.hacluster_passwd_changed;
```
nelsonkopliku, line 44: The semantic of the `verify_password` gatherer is slightly different as it returns `true` in case a well known password has been detected for the provided user, `false` otherwise ([doc here](https://www.trento-project.io/wanda/gatherers.html#verify_password)).

`facts.hacluster_passwd_changed` would be `true` for the default password `linux` meaning that it has **not** been changed and expecting `!facts.hacluster_passwd_changed` does not sound like what we want, 'cause afaik we actually want the opposite: _we do want well known passwords to be changed_.

We could do the following:
```yaml
facts:
  - name: hacluster_has_default_password # or whatever you prefer
    gatherer: verify_password
    argument: hacluster

expectations:
  - name: expectations_hacluster_passwd_changed
    expect: !facts.hacluster_has_default_password
```

nelsonkopliku (changes requested):
One more, awesome!
We just need to swap the logic and we're good :wink:

priv/catalog/790926.yaml
angelabriel: ok, I see my chosen names where not really self-explaining :-) My intention was like 'no, pasword was not changed'
But I agree, your names are fitting much better.
Will change soon.
Btw. 
`expect: !facts.hacluster_has_default_password` gives me an error
`Error: Error("expectations[0].expect: invalid type: enum, expected any valid JSON value", line: 43, column: 13)`
from tlint and trying to run the check against wanda is doing simply nothing. wanda is silently skipping the check.

That's the reason I wrote
    ```
expect: |
      !facts.hacluster_passwd_changed;
```
Is there something I can change in my environment? Or is it a problem in wanda?

arbulu89 (approved):

nelsonkopliku (approved):

PR #154, opened 2023-01-05
Implement when condition

dottorblaster:
This PR implements a `when` clause inside the YAML specification. `env` variables can be provided while retrieving the catalog. The checks then are matched against the `when` clause. If a `when` clause doesn't evaluate truthy, the check gets filtered away.

Example:

```yml
when: env.provider == "azure"
```

### DIFF
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -18,13 +18,14 @@ defmodule Wanda.Catalog do
   @doc """
   Get the checks catalog with all checks
   """
-  @spec get_catalog() :: [Check.t()]
-  def get_catalog do
+  @spec get_catalog(map()) :: [Check.t()]
+  def get_catalog(env \\ %{}) do
     get_catalog_path()
     |> Path.join("/*")
     |> Path.wildcard()
     |> Enum.map(&Path.basename(&1, ".yaml"))
     |> get_checks()
+    |> Enum.filter(&when_condition(&1, env))
   end
 
   @doc """
@@ -63,6 +64,17 @@ defmodule Wanda.Catalog do
     end)
   end
 
+  defp when_condition(%Check{when: nil}, _), do: true
+
+  defp when_condition(%Check{when: when_clause}, env) do
+    scope = %{"env" => env}
+
+    case Rhai.eval(when_clause, scope) do
+      {:ok, true} -> true
+      _ -> false
+    end
+  end
+
   defp get_catalog_path do
     Application.fetch_env!(:wanda, Wanda.Catalog)[:catalog_path]
   end
@@ -85,6 +97,7 @@ defmodule Wanda.Catalog do
        group: group,
        description: description,
        remediation: remediation,
+       when: Map.get(check, "when"),
        severity: map_severity(check),
        facts: Enum.map(facts, &map_fact/1),
        values: map_values(check),
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -18,7 +18,7 @@ defmodule Wanda.Catalog do
   @doc """
   Get the checks catalog with all checks
   """
-  @spec get_catalog(map()) :: [Check.t()]
+  @spec get_catalog(%{String.t() => String.t()}) :: [Check.t()]
   def get_catalog(env \\ %{}) do
     get_catalog_path()
     |> Path.join("/*")
@@ -65,12 +65,12 @@ defmodule Wanda.Catalog do
     |> Enum.filter(&when_condition(&1, env))
   end
 
+  defp when_condition(_, env) when env == %{}, do: true
+
   defp when_condition(%Check{when: nil}, _), do: true
 
   defp when_condition(%Check{when: when_clause}, env) do
-    scope = %{"env" => env}
-
-    case Rhai.eval(when_clause, scope) do
+    case Rhai.eval(when_clause, %{"env" => env}) do
       {:ok, true} -> true
       _ -> false
     end
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -24,8 +24,7 @@ defmodule Wanda.Catalog do
     |> Path.join("/*")
     |> Path.wildcard()
     |> Enum.map(&Path.basename(&1, ".yaml"))
-    |> get_checks()
-    |> Enum.filter(&when_condition(&1, env))
+    |> get_checks(env)
   end
 
   @doc """
@@ -54,14 +53,16 @@ defmodule Wanda.Catalog do
   @doc """
   Get specific checks from the catalog.
   """
-  @spec get_checks([String.t()]) :: [Check.t()]
-  def get_checks(checks_id) do
-    Enum.flat_map(checks_id, fn check_id ->
+  @spec get_checks([String.t()], map()) :: [Check.t()]
+  def get_checks(checks_id, env) do
+    checks_id
+    |> Enum.flat_map(fn check_id ->
       case get_check(check_id) do
         {:ok, check} -> [check]
         {:error, _} -> []
       end
     end)
+    |> Enum.filter(&when_condition(&1, env))
   end
 
   defp when_condition(%Check{when: nil}, _), do: true
--- a/lib/wanda/executions/server.ex
+++ b/lib/wanda/executions/server.ex
@@ -27,6 +27,12 @@ defmodule Wanda.Executions.Server do
 
   @default_timeout 5 * 60 * 1_000
 
+  @doc """
+  Starts a check execution.
+
+  The checks are filtered leveraging the `env` and evaluating the `when` condition using a best-effort approach.
+  If non-existing check IDs are provided inside a target, they will get filtered away.
+  """
   @impl true
   def start_execution(execution_id, group_id, targets, env, config \\ []) do
     checks =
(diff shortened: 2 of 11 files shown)


test/catalog_test.exs
```diff
@@ -34,6 +34,21 @@ defmodule Wanda.CatalogTest do
...
 
+    test "should return the whole catalog but the check with the when clause" do
+      catalog_path = Application.fetch_env!(:wanda, Wanda.Catalog)[:catalog_path]
+
+      valid_files =
+        catalog_path
+        |> File.ls!()
+        |> Enum.sort()
+        |> Enum.filter(fn file -> file != "malformed_check.yaml" end)
+
```
fabriziosestito, line 45: maybe we could improve this without using logic but being explicit about which checks we are expecting

lib/wanda_web/schemas/env.ex
```diff
@@ -0,0 +1,23 @@
...
+  OpenApiSpex.schema(
+    %{
+      title: "ExecutionEnv",
+      description: "Contextual Environment for the current execution",
+      type: :object,
+      additionalProperties: %Schema{
+        oneOf: [
+          %Schema{type: :string},
+          %Schema{type: :integer},
+          %Schema{type: :boolean}
```
fabriziosestito, line 17: this could also be an array of itself

fabriziosestito (changes requested):
We could add a test case to the catalog controller test with when condition active based on the length of the items

test/catalog_test.exs
```diff
@@ -34,6 +34,21 @@ defmodule Wanda.CatalogTest do
...
+      catalog = Catalog.get_catalog(%{"provider" => "aws"})
+      assert length(valid_files) - 1 == length(catalog)
+
+      assert Enum.any?(catalog, fn %Check{id: id} -> id == "when_condition_check" end) == false
```
fabriziosestito, line 49: use `refute` instead of `assert`

fabriziosestito (changes requested):

lib/wanda/executions/server.ex
```diff
@@ -32,7 +32,7 @@ defmodule Wanda.Executions.Server do
     checks =
       targets
       |> Target.get_checks_from_targets()
-      |> Catalog.get_checks()
+      |> Catalog.get_checks(env)
```
arbulu89, line 35: As a comment, with this we are not simply filtering the catalog checks, we are filtering the execution as well, and I'm not sure if this was part of the initial idea.

test/fixtures/catalog/when_condition_check.yaml
```diff
@@ -0,0 +1,33 @@
+id: when_condition_check
```
arbulu89, line 1: Do we need a whole new fixture to test this new field?
Couldn't we use an already existing one and add the field there?

test/catalog_test.exs
```diff
@@ -21,7 +21,7 @@ defmodule Wanda.CatalogTest do
         |> Enum.sort()
         |> Enum.filter(fn file -> file != "malformed_check.yaml" end)
 
-      catalog = Catalog.get_catalog()
+      catalog = Catalog.get_catalog(%{"provider" => "azure"})
```
arbulu89, line 24: I think the previous implementation made more sense, because it was testing that the catalog is retrieved entirely if no `env` is given, which is something we should still be testing.
More when the test name is not changed.
I would keep this as it was and create new tests with `azure` as provider

test/catalog_test.exs
```diff
@@ -34,6 +34,21 @@ defmodule Wanda.CatalogTest do
       end)
     end
 
+    test "should return the whole catalog but the check with the when clause" do
+      catalog_path = Application.fetch_env!(:wanda, Wanda.Catalog)[:catalog_path]
+
+      valid_files =
```
arbulu89, line 40: I think we could remove this `valid_files` code together with the `assert length(valid_files) - 1 == length(catalog)` line, this is already tested.
We should aim to see that the when clauses that don't match are not returned, which is already done with the latest assert. I think it should be enough

test/catalog_test.exs
```diff
@@ -34,6 +34,21 @@ defmodule Wanda.CatalogTest do
       end)
     end
 
+    test "should return the whole catalog but the check with the when clause" do
```
arbulu89, line 37: To have a more meaningful test name:
`should filter out checks if the when clause doesn't match` or similar

arbulu89 (commented):
@dottorblaster The code looks great.
I added some test related comments, and with them, I miss a test in the controllers side to see that correct encoded urls still pass the spec.
(`/api/checks/catalog?provider=azure&foo=bar` or something like this)

lib/wanda/catalog.ex
```diff
@@ -18,13 +18,13 @@ defmodule Wanda.Catalog do
   @doc """
   Get the checks catalog with all checks
   """
-  @spec get_catalog() :: [Check.t()]
-  def get_catalog do
+  @spec get_catalog(map()) :: [Check.t()]
```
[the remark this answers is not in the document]
fabriziosestito: You don' need required since required is the default (vs optional)

```elixir
%{String.t() => String.t()}
```
is enough

lib/wanda/executions/server.ex
fabriziosestito: it wasn't, but since it came free without writing code I ask @dottorblaster  to do the filtering here.
wdyt?

test/fixtures/catalog/when_condition_check.yaml
fabriziosestito: + 1, i think we can squash several checks fixtures in just two of them, one being the "valid" and one being "malformed"

lib/wanda/executions/server.ex
arbulu89: It is fine, but I think this should be documented somewhere, as It can lead to confusion

fabriziosestito: I think we can document that the general behavior is  "best effort". We do the same for malformed or non-existing checks. @dottorblaster proposes to add a field in the result with the skipped checks this is an interesting option as well (it could also help us remove non-existing checks from the check selection, the migration we were talking about).

test/catalog_test.exs
dottorblaster: > that the catalog is retrieved entirely if no env is given

Good catch, it actually makes sense to return the whole catalog if no env is given. Current PR doesn't work that way.

lib/wanda/executions/server.ex
arbulu89: Ok, let's keep in the radar. In the current runner we have the skipped thing, but It never worked out pretty well.
Let's go with the current approach. I just only wanted to highlight that these things must be of general knowledge

nelsonkopliku (commented):
Nice job!
Not much to add to what already shared.

The only ambiguous thing I find here is that we use an empty env as the signal to get the whole catalog (where env should be irrelevant) and that we are not considering an empty env as a case when asking the selectable checks.

Splitting into two operations for the two actions could reduce ambiguity, however I understand it might be too much and that the current implementation just works, so I'm also fine with it.

lib/wanda/catalog.ex
```diff
@@ -53,14 +53,29 @@ defmodule Wanda.Catalog do
...
     end)
+    |> Enum.filter(&when_condition(&1, env))
+  end
+
+  defp when_condition(_, env) when env == %{}, do: true
+
+  defp when_condition(%Check{when: nil}, _), do: true
+
+  defp when_condition(%Check{when: when_clause}, env) do
+    scope = %{"env" => env}
```
jamie-suse, line 73: Tiny nitpick; why do you place the map into a temporary variable `scope` here? Why not just `case Rhai.eval(when_clause, %{"env" => env}) do` as it's not used anywhere else?

dottorblaster: You're right, didn't seem like a good one to me, then your comment made me change my mind :smile:

lib/wanda/executions/server.ex
fabriziosestito: @dottorblaster could we document the best effort approach in the Execution Server module documentation?

test/wanda_web/controllers/catalog_controller_test.exs
```diff
@@ -14,5 +14,15 @@ defmodule WandaWeb.CatalogControllerTest do
...
     end
+
+    test "listing the checks catalog produces a CatalogResponse when filtered", %{conn: conn} do
+      json =
+        conn
+        |> get("/api/checks/catalog?provider=azure&foo=bar")
+        |> json_response(200)
+
+      api_spec = ApiSpec.spec()
+      assert_schema(json, "CatalogResponse", api_spec)
```
fabriziosestito, line 25: we should test at least the length of the response here, to see if it matches the expected number of checks

lib/wanda_web/schemas/catalog/check.ex
```diff
@@ -14,6 +14,11 @@ defmodule WandaWeb.Schemas.CatalogResponse.Check do
     properties: %{
       id: %Schema{type: :string, description: "Check ID"},
       name: %Schema{type: :string, description: "Check name"},
+      when: %Schema{
+        type: :string,
+        description: "When clause used to filter checks",
```
fabriziosestito, line 19: rephrase "Expression to determine whether a check should run"

fabriziosestito (changes requested):

fabriziosestito (approved):

PR #155, opened 2023-01-10
Document best practices

arbulu89:
In order to have some "order" in the way checks are written, here some best practices and conventions.
I recommend continuing adding more items as we go, so we have a rich list of things to do and not do.


rtorrero (approved):
Looks great, thanks @arbulu89

dottorblaster (approved):
LGTM

PR #156, opened 2023-01-11
Add starting the targets section on the docs

arbulu89:
Add new section in docs explaining how to start the trento-agent on the targets

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,10 @@ The [trento-agent](https://github.com/trento-project/agent) must be up and runni
 Here an example on how to start it:
 
 ```
-./trento-agent start --ssh-address 0 --api-key 0 --factsengine --facts-service-url amqp://wanda:wanda@localhost:5674
+./trento-agent start --ssh-address <ssh-address> --api-key <api-key> --factsengine --facts-service-url amqp://wanda:wanda@localhost:5674
 ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,10 +37,31 @@ Wanda is exposed on port `4000` and the API documentation is available at http:/
...
 - checking the state of an execution
 
+#### **Starting the targets**
+
+The [trento-agent](https://github.com/trento-project/agent) must be up and running on the targets to run a correct execution, otherwise a timeout error is raised.
+
+Here an example on how to start it:
+
+```
+./trento-agent start --ssh-address 0 --api-key 0 --factsengine --facts-service-url amqp://wanda:wanda@localhost:5674
```
nelsonkopliku, line 52: What about using some placeholders instead of `0`

`./trento-agent start --ssh-address <an-ssh-address> --api-key <an-api-key> --factsengine --facts-service-url...`

I understand that those options are not really needed for the scenario of checks execution, however `0` might result a bit confusing, at least to me :sweat_smile:

nelsonkopliku (approved):
Nice and thanks!

I only have a doubt about using `0` in the sample. Maybe a placeholder would do it as well.

README.md
arbulu89: You are right!
I put placeholders and note comment to be more helpful

PR #157, opened 2023-01-12
Fix integer mapping

fabriziosestito:
[Structpb](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#Value_NumberValue) does not differentiate between integers and floats, so we receive numbers such as `1.0` as integers from protobuf.

This PR adds updates the mapping functions to return integers in these situations.

Nothing really changes in Rhai since `1 == 1.0` and vice-versa, but this makes the facts gathered more consistent.


nelsonkopliku (approved):
:eyes:

dottorblaster (approved):
LGTM

arbulu89 (approved):

PR #158, opened 2023-01-12
Document the trento-agent facts command

arbulu89:
More documentation about how to debug the fact gathering process using the `trento-agent facts` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ With a running setup, it is possible to easily test Checks and their Execution b
 - consulting the catalog
 - starting a Checks Execution
 - checking the state of an execution
+- debugging the gathered facts
 
 #### **Consulting the catalog**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -174,6 +175,33 @@ curl --request GET 'http://localhost:4000/api/checks/executions/205e326d-0c25-4f
...
+```
+
+Which would return the next where the `Value` is the available value in the written check:
+
+```
+{
+  "Name": "totem.token",
+  "CheckID": "",
+  "Value": {
+    "Value": 30000
```
nelsonkopliku, line 199: Do you think it is worth adding a note to clarify the double `Value`?
It's tricky though, because it carries serialization concerns which might be out of scope for this piece of doc.

Maybe we could just add a reference to the relation with the DSL, for instance just saying that there's a serialization/transformation process that allows to access the fact like `facts.<fact_name>` and get back `30000`.

To make it more realistic we could chose another fact name for this example because `totem.token` is more of an argument and I don't know whether we support, or even if we should support, dotted fact names like `totem.token`.

arbulu89: We cannot change the fact name, this is how the `facts` subcommand works. It puts the passed argument as name

nelsonkopliku: I see. Maybe it could just be fixed to `fact_name` or similar :smile: in the end this is a debugging tool.
However, I am also fine with how it is now.

arbulu89: Maybe we should update the `facts` subcommand in the agent side, and add a `--json` flag or something to print something readable, and not the internal `FactValue` stuff

fabriziosestito: tracked as debt, we could even try to output rhai syntax

PR #159, opened 2023-01-12
Clean up doc saying that expectation evaluation has access to the env

nelsonkopliku:
As per the title. A small clean up.


arbulu89 (approved):

arbulu89:
I hope we don't have to add sonnish the `env` rendering the expectation XD
(I think we will, sooner than later hehe)

nelsonkopliku:
ahaha @arbulu89  I was indeed thinking: should I just add the env to the expectation evaluation?
However, we'll fix when needed, at least now doc and code are aligned :smile:

PR #163, opened 2023-01-16
Add premium flag option to catalog loading code

arbulu89:
Add `premium` field to the checks speficication

### DIFF
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -26,7 +26,8 @@ defmodule Wanda.Factory do
       severity: Enum.random([:critical, :warning, :passing]),
       facts: build_list(10, :catalog_fact),
       values: build_list(10, :catalog_value),
-      expectations: build_list(10, :catalog_expectation)
+      expectations: build_list(10, :catalog_expectation),
+      premium: Enum.random([false, true])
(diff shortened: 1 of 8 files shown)


nelsonkopliku (approved):
Premium :tm:

EMaksy (commented):
LGTM :D

test/support/factory.ex
```diff
@@ -26,7 +26,8 @@ defmodule Wanda.Factory do
       severity: Enum.random([:critical, :warning, :passing]),
       facts: build_list(10, :catalog_fact),
       values: build_list(10, :catalog_value),
-      expectations: build_list(10, :catalog_expectation)
+      expectations: build_list(10, :catalog_expectation),
+      premium: false
```
dottorblaster, line 30:
```suggestion
      premium: Enum.random([true, false])
```

dottorblaster (changes requested):
Just a tiny change request then I think we can merge

test/support/factory.ex
arbulu89: Done

fabriziosestito (approved):

PR #164, opened 2023-01-16
Force rhai_rustler build 

fabriziosestito:
There is no real gain in using precompiled NIFs in Wanda, since we can assume we have the rust toolchain installed.
Also we want to build the container in OBS/IBS, where we can't really use precompiled NIFs.

This PR bumps `rhai_rustler` to v0.1.4 and disables the precompilation.


arbulu89 (approved):
Thanks!

PR #165, opened 2023-01-17
Add list & map examples for `corosync-cmapctl` docs

jamie-suse:
Adds some non-trivial examples to the documentation of requesting and returning lists and maps from the `corosync-cmapctl` gatherer.

### DIFF
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -159,15 +159,15 @@ This gatherer allows accessing the output of the `corosync-cmapctl` tool. It sup
 
 Example arguments:
 
-| Name                         | Return value                     |
-| :--------------------------- | :------------------------------- |
-| `totem.token`                | extracted value from the command |
-| `runtime.config.totem.token` | extracted value from the command |
-| `totem.transport`            | extracted value from the command |
-| `totem.transport`            | extracted value from the command |
-| `nodelist.node.0.ring0_addr` | extracted value from the command |
-| `nodelist.node`              | extracted value from the command |
-| `nodelist.node.1`            | extracted value from the command |
+| Name                                | Return value                     |
+| :---------------------------------- | :------------------------------- |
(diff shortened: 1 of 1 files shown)


guides/gatherers.md
```diff
@@ -209,6 +219,15 @@ Example output (in Rhai):
 
 // node_0_ring0addr
 20;
+
```
arbulu89, line 222: Make me a favor. It was not in your PR, but change the line above (`// node_0_ring0addr` to have `"10.80.1.11"` value below). It was wrong from before

guides/gatherers.md
```diff
@@ -209,6 +219,15 @@ Example output (in Rhai):
...
+// node_list
+#{
+  "0": #{ "nodeid": 1, "ring0_addr": "10.80.1.11" },
```
arbulu89, line 225: Maybe multi-line representation looks better, but well, this one is good as well

arbulu89 (approved):
Thanks @jamie-suse ,
Please, change the commented line before merging (the `node_0_ring0addr` thing)

guides/gatherers.md
jamie-suse: Oh wow, totally missed that 👀 thanks

guides/gatherers.md
jamie-suse: Change to multiline 👍

guides/gatherers.md
arbulu89: Why does it add parenthesis there??
It happened the same with `udpu`. In that case was due the prettier. Could you remove them?

arbulu89: I hope it doesn't affect you @fabriziosestito

jamie-suse: Removed 👍 I looked through the rest of the file and didn't find other instances of it happening

PR #166, opened 2023-01-17
Set critical state on agent timeout and warning severity

arbulu89:
When an agent timeout error happens, instead of using the severity field to set the error, critical is set directly.
The timeout is critical scenario, so the code sets this way.

PD: Object to change if at some point we add a new field to store the "errored" checks, to not use the critical state

PD2: The Web frontend representation will be still quite strange, as we don't really use the check results as result (we use the expectations and do some computations), as the code still has a "agent" based visualization. So it can show some results as critical even though they should be warning. It should be fixed once we move to the check based visualization

### DIFF
--- a/lib/wanda/executions/evaluation.ex
+++ b/lib/wanda/executions/evaluation.ex
@@ -242,7 +242,10 @@ defmodule Wanda.Executions.Evaluation do
            not errors?(agents_check_results) do
         :passing
       else
-        severity
+        Enum.find_value(agents_check_results, severity, fn
+          %AgentCheckError{type: :timeout} -> :critical
+          _ -> false
+        end)
       end
 
     %CheckResult{check_result | result: result}
--- a/test/executions/evaluation_test.exs
+++ b/test/executions/evaluation_test.exs
@@ -626,6 +626,45 @@ defmodule Wanda.Executions.EvaluationTest do
                result: :warning
              } = Evaluation.execute(execution_id, group_id, checks, gathered_facts, %{})
     end
+
+    test "should return a critical result if an agent times out and severity is warning" do
+      execution_id = UUID.uuid4()
+      group_id = UUID.uuid4()
+
+      [%Catalog.Fact{name: fact_name}] = catalog_facts = build_list(1, :catalog_fact)
+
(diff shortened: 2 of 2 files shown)


lib/wanda/executions/evaluation.ex
```diff
@@ -242,7 +242,7 @@ defmodule Wanda.Executions.Evaluation do
            not errors?(agents_check_results) do
         :passing
       else
-        severity
+        eval_severity(severity, agents_check_results)
```
nelsonkopliku, line 245: `eval_` might be confused with an actual expression evaluation.

What about `result_with_severity`, `severity_based_result`, `with_severity` or something similar? 
Something that says the result is being calculated based also on severity.

test/executions/evaluation_test.exs
```diff
@@ -626,6 +626,45 @@ defmodule Wanda.Executions.EvaluationTest do
                result: :warning
              } = Evaluation.execute(execution_id, group_id, checks, gathered_facts, %{})
     end
+
+    test "should return a critical result agent times out and severity is warning" do
```
nelsonkopliku, line 630: Maybe there's a missing `when` in the sentence
> should return a critical result _when_ agent times out and severity is warning

nelsonkopliku (commented):
I only have a doubt on the `eval_severity` naming. Other than that LGTM.

test/executions/evaluation_test.exs
arbulu89: Will change to `...critical result if an agent times ...` to follow the same syntax than the others.

lib/wanda/executions/evaluation.ex
arbulu89: I'm thinking to put this piece of code even directly on the `aggregate_check_result` else side. It is pretty simple, so maybe we don't need an extra function.
What do you think?

nelsonkopliku: That's also a valid alternative :+1:

arbulu89: Update amended

nelsonkopliku (approved):
Great, thanks!

fabriziosestito (approved):

PR #167, opened 2023-01-17
rewrite trento community check regarding 'running corosync rings' 

angelabriel:
fixing jsc#CFSA-1076, check 1.1.9.runtime/32CFC6

### DIFF
--- a/priv/catalog/32CFC6.yaml
+++ b/priv/catalog/32CFC6.yaml
@@ -1,0 +1,49 @@
+id: "32CFC6"
+name: corosync running 2 ring configuration
+group: Corosync
+description: |
+  Corosync is running with at least 2 rings
+remediation: |
+  ## Abstract
+  It is strongly recommended to add a second ring to the corosync communication.
+
+  ## References
+  Azure:
+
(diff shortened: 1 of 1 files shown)


arbulu89 (approved):
Perfect!

priv/catalog/32CFC6.yaml
```diff
@@ -0,0 +1,49 @@
...
+values:
+  - name: expected_totem_interfaces
+    default: 2
+    conditions:
+      - value: 1
+        when: env.provider == "azure" || env.provider == "gcp"
+
+expectations:
+  - name: expectations_totem_interfaces
+    expect: facts.totem_interfaces.len() == values.expected_totem_interfaces
```
nelsonkopliku, line 49: Does the `at least` particle in the description mean we need `>=`? 
Also in relation to discussions had here about the check  **DA114A** https://github.com/trento-project/wanda/pull/141#issuecomment-1361642249

nelsonkopliku (approved):
Super! Just a tiny doubt, but we're good.

priv/catalog/32CFC6.yaml
angelabriel: oh, oh, you are so right. Will fix it asap. Many thanks for the attention

PR #168, opened 2023-01-18
Add JWT auth

fabriziosestito:
This PR adds JWT auth to Wanda by adding a Plug that deals with the authentication.
The plug verifies the JWT from the header, similarly on what we did on web.


lib/wanda_web/auth/access_token.ex
```diff
@@ -0,0 +1,16 @@
+defmodule WandaWeb.Auth.AccessToken do
+  @moduledoc """
+    Jwt Token is the module responsible for creating a proper jwt access token.
+    Uses Joken as jwt base library
+  """
+
+  use Joken.Config, default_signer: :access_token_signer
+
+  @iss "https://github.com/trento-project/web"
```
arbulu89, line 9: Shouldn't have these 2 in the config file as in the web project?
It should use `wanda` url by the way.

config/config.exs
```diff
@@ -50,6 +50,8 @@ config :phoenix, :json_library, Jason
 # Disable rustler precompiled NIFs
 config :rustler_precompiled, :force_build, rhai_rustler: true
 
+config :wanda, :jwt_authentication, enabled: true
```
arbulu89, line 53: Are we using the default expiration time?
I think it would be valuable to make it optional. It is 2 hours always otherwise

config/test.exs
```diff
@@ -55,3 +55,9 @@ config :wanda, Wanda.Messaging.Adapters.AMQP,
 
 config :wanda,
   children: [Wanda.Messaging.Adapters.AMQP.Publisher, Wanda.Messaging.Adapters.AMQP.Consumer]
+
+config :joken,
+  access_token_signer: "s2ZdE+3+ke1USHEJ5O45KT364KiXPYaB9cJPdH3p60t8yT0nkLexLBNw8TFSzC7k",
+  refresh_token_signer: "L0wvcZh3ACQpibVhV/nh5jd/NaZWL4ijZxTxGJMGpacuXIBc4In3YCwXeVM98ygp"
```
arbulu89, line 61: copy paste error on the refresh token

test/wanda_web/auth/jwt_auth_plug_test.exs
```diff
@@ -0,0 +1,69 @@
...
+      conn =
+        build_conn()
+        |> put_req_header("authorization", "Bearer " <> jwt)
+        |> JWTAuthPlug.call(%{})
+
+      refute conn.halted
+      assert conn.private.user_id == 1
+    end
+
+    test "should  be halted and have unauthorized status if the token is expired" do
```
arbulu89, line 36: extra space after `should`

test/wanda_web/auth/jwt_auth_plug_test.exs
```diff
@@ -0,0 +1,69 @@
...
+      expired_jwt =
+        "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJ0cmVudG8tcHJvamVjdCIsImV4cCI6MTY3MTY0MjQxNCwiaWF0IjoxNjcxNjQxODE0LCJpc3MiOiJodHRwczovL2dpdGh1Yi5jb20vdHJlbnRvLXByb2plY3Qvd2ViIiwianRpIjoiMnNwaTFvbmxxbml1ZnE5dnVrMDAwMG9hIiwibmJmIjoxNjcxNjQxODE0LCJzdWIiOjEsInR5cCI6IkJlYXJlciJ9.oub6_NsHcVIyd0de14Lzk3SuCMMgr8O-sSWLr7Gxcp8"
```
arbulu89, line 38: just as curiosity, how do you create this invalid keys? like this expired example

arbulu89 (commented):
Nice!
I have noted some typos and minor comments.
Besides that, we would need some docs, and keep in mind that once this is merged, we need to update the web part to use the token in the requests

lib/wanda_web/auth/access_token.ex
fabriziosestito: The issuer is the web since it acts as an identity provider.

I don't think those two are configurations, even if the web they are. In fact, I wanted to change  web, because we don't need them to be dynamic.

config/config.exs
fabriziosestito: we are not, expiration is just for generating jwt and wanda does not generate them just read them.

test/wanda_web/auth/jwt_auth_plug_test.exs
fabriziosestito: you can use http://jwtbuilder.jamiekurtz.com/

lib/wanda_web/auth/access_token.ex
arbulu89: Ok, I see that it shares the same issuer with the web as this 1st one is the one responsible of issuing keys for wand too.

dottorblaster (approved):

nelsonkopliku (commented):
Nice and clean! Great!

When I run this branch locally and browse to http://localhost:4001/swaggerui I get an error loading the api definition at http://localhost:4001/api/openapi because it is authenticated.

Am I doing something wrong? Or we simply need to keep that resource openly accessible?

fabriziosestito:
> Nice and clean! Great!
> 
> When I run this branch locally and browse to http://localhost:4001/swaggerui I get an error loading the api definition at http://localhost:4001/api/openapi because it is authenticated.
> 
> Am I doing something wrong? Or we simply need to keep that resource openly accessible?

@nelsonkopliku good catch, that route shouldn't be protected

fabriziosestito:
@arbulu89 @nelsonkopliku addressed the comments and disabled jwt authentication in the checks development scenario

arbulu89 (approved):

PR #172, opened 2023-01-20
Fix default value when getting system env for JWT_AUTHENTICATION_ENABLED

nelsonkopliku:
`System.get_env/2` expects a string as a second argument.

Caught while working on building wanda in OBS.


fabriziosestito (approved):
Woops, my bad.
Thanks for taking care of this!

PR #173, opened 2023-01-20
Build in obs

nelsonkopliku:
# Description

This PR adds submission of wanda to OBS.

More in detail
- a hack directory added with necessary scripts for CI
- added suse Docker packaging files
- vendored rhai rustler cargo deps

Reminder to self: before merging update obs meta and merge web bci base update https://github.com/trento-project/web/pull/1130

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -171,7 +171,7 @@ jobs:
       packages: write
     env:
       REGISTRY: ghcr.io
-      IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/wanda
+      IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/trento-wanda
       IMAGE_TAG: "${{ (github.event_name == 'release' && github.event.release.tag_name) || (github.event_name == 'push' && github.ref_name == 'main' && 'rolling') || github.sha }}"
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,10 @@ jobs:
     name: Elixir dependencies
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
       - name: Checkout
         uses: actions/checkout@v3
         with:
@@ -70,6 +74,10 @@ jobs:
     needs: elixir-deps
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
       - name: Checkout
         uses: actions/checkout@v3
         with:
@@ -126,6 +134,10 @@ jobs:
         ports:
           - 5674:5672
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
       - name: Checkout
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
 
   tlint:
     name: Lint checks
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
fabriziosestito, line 59: why is this needed?

nelsonkopliku: No strict need detected actually, I guess I just pinned os version for all the jobs.
I'll revert that to `ubuntu-latest` if it is preferable.

fabriziosestito: If you don't need to do so, I suggest dropping the commit.
This might be needed if we decide to go with a matrix testing and test against an OTP version that is not supported by ubuntu 22.04, we can re-add modification in that PR if needed (there is already a debt for this).

nelsonkopliku: Makes sense, thanks! Changed back to latest

.github/workflows/ci.yaml
```diff
@@ -91,7 +99,7 @@ jobs:
             _build/test
             priv/plts
           key: ${{ runner.os }}-${{ env.OTP_VERSION }}-${{ env.ELIXIR_VERSION }}-${{ hashFiles('mix.lock') }}
-      - name: Check for Unused Dependencies
+      - name: Check for unused dependencies
```
arbulu89, line 102: I don't really get this name change, as all the remaining step names are in title case

.github/workflows/ci.yaml
```diff
@@ -221,3 +233,77 @@ jobs:
...
+    name: Commit the project on OBS
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
+    needs: [static-code-analysis, test]
+    container:
+      image: ghcr.io/trento-project/continuous-delivery:main
+      env:
+        GITHUB_OAUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        DEST_FOLDER: "/tmp/osc_project"
+        NAME: trento-wanda-image
```
arbulu89, line 247: This naming could be confusing, as the container in the public environment is named `wanda`, without any trento prefix.
We should think carefully how we want to name this

packaging/suse/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: Apache-2.0
+#!BuildTag: trento/trento-wanda:latest
+#!BuildTag: trento/trento-wanda:%%VERSION%%
+#!BuildTag: trento/trento-wanda:%%VERSION%%-build%RELEASE%
+#!UseOBSRepositories
+
+FROM bci/rust:1.66 AS release
```
arbulu89, line 7: Let's test this first in IBS before merging

arbulu89, line 7: Just out of curiosity, why do we need a `rust` image here if the rust files are already vendored?

.github/workflows/ci.yaml
nelsonkopliku: Just copied from web :smile: Putting back as it was.

packaging/suse/Dockerfile
nelsonkopliku: Tried and seems to be working https://build.opensuse.org/package/show/home:nkopliku:trento-factory/trento-wanda-image

nelsonkopliku: Because the deps are vendored ineed, not built. They get built later in OBS.

arbulu89: That's an OBS url. Not IBS, which has a different repositories/package configuration

nelsonkopliku: True :sweat_smile: I sent you the link to IBS

.github/workflows/ci.yaml
nelsonkopliku: I see. It just follows the same naming pattern we use for web.
| Project | gh image name | OBS/IBS image name |
|---	|---	|---	|
| web | `trento-web` | `trento-web-image` |
| wanda | `wanda` | `trento-wanda-image` |

We are somewhat consistently inconsistent :sweat_smile: 
What should we do?

nelsonkopliku: Options:
- rename the gh image to `trento-wanda`.
- move away from `wanda` to something checks engine related (`trento-checks`, `trento-checks-engine` or similar)

I'd go for the first one, sounds simpler.

fabriziosestito: +1 to keep consitency with the trento prefix if possible

nelsonkopliku: https://github.com/trento-project/wanda/pull/173/commits/8a1590b339ad85aaee969d809d395871535e8d85

fabriziosestito (approved):

arbulu89 (approved):

PR #174, opened 2023-01-20
Fix cors plug integration

CDimonaco:
Fix integration with the cors_plug library, according to the library readme.

The plug should be added in the endpoint before the routing middleware chain in order to match all the preflight requests.


fabriziosestito (approved):

fabriziosestito (changes requested):
@CDimonaco you need to sign your commits

CDimonaco:
Signed

fabriziosestito:
@CDimonaco Could you split this into two PRs?  Updating the contracts library shouldn't be squashed with this fix (what if we need to debug with git bisect?)

CDimonaco:
Updated Contract ref is already on main, I think something happened with some rebase when I signed the commits afterwards

nelsonkopliku (approved):

fabriziosestito:
> Updated Contract ref is already on main, I think something happened with some rebase when I signed the commits afterwards
@CDimonaco could you drop the commits then so we can merge?

fabriziosestito (approved):

PR #175, opened 2023-01-23
Use new `trento-wanda` image name for check development environment

nelsonkopliku:
As per the title. We changed the name of the built image from `wanda` to `trento-wanda`


arbulu89 (approved):
Let's remove the old one afterwards, and maybe we need to give some additional config change to the new

fabriziosestito (approved):

PR #176, opened 2023-01-24
Fix JWT plug runtime config

fabriziosestito:
This PR fixes the JWT runtime config that was not read since the router uses compile-time macros, by properly utilizing the Plug `init` callback.

It also adds a test for the disabled scenario


nelsonkopliku (approved):
Super!

PR #179, opened 2023-01-26
Build wanda with premium checks, if available

nelsonkopliku:
Allows extending the Checks Catalog with premium Checks when available.

### DIFF
--- a/packaging/suse/Dockerfile
+++ b/packaging/suse/Dockerfile
@@ -5,7 +5,8 @@
 #!UseOBSRepositories
 
 FROM bci/rust:1.66 AS release
-ADD wanda.tar.gz /build/
+ADD wanda.tar.gz premium-checks.tar.gz* /build/
+RUN mv /build/priv/catalog/* /build/wanda/priv/catalog || true && rm -fr /build/priv/
 # Workaround for https://github.com/openSUSE/obs-build/issues/487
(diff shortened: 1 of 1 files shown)


packaging/suse/Dockerfile
```diff
@@ -5,7 +5,8 @@
 #!UseOBSRepositories
 
 FROM bci/rust:1.66 AS release
-ADD wanda.tar.gz /build/
+ADD wanda.tar.gz premium-checks.tar.gz* /build/
```
arbulu89, line 8: It wouldn't be better:
```
ADD wanda.tar.gz /build/
ADD premium-checks.tar.gz* /build/wanda/
```
As the `ADD` command extract the tar file in the destination folder.
Didn't test it though

nelsonkopliku: Tested and doesn't work when `premium-checks.tar.gz` is not there. Docker wants at least the first item for an `ADD` to actually be there.

arbulu89: ok!
Thank you for trying out

arbulu89 (approved):

dottorblaster (approved):

rtorrero (approved):
LGTM

EMaksy (commented):
LGTM :D

PR #180, opened 2023-02-06
Start amqp consumer and publisher only if needed in tests

arbulu89:
Start AMQP consumer and publisher only in the needed tests. The code waits to a amqp telemetry message to know that the connection of the consumer/publisher is open, and we need to wait some milliseconds as the usage of the queue is not instantaneous even once the connection is open.

Created a new module as recommended in the docs instead of an anonymous function: https://hexdocs.pm/telemetry/telemetry.html#attach/4

PD: 
Here how the telemetry modules works:
https://github.com/beam-telemetry/telemetry
And specifically for the amqp module:
https://github.com/meltwater/gen_rmq/blob/master/lib/gen_rmq/consumer/telemetry.ex

### DIFF
--- a/test/wanda/messaging/adapters/amqp/consumer_test.exs
+++ b/test/wanda/messaging/adapters/amqp/consumer_test.exs
@@ -3,11 +3,41 @@ defmodule Wanda.Messaging.Adapters.AMQP.ConsumerTest do
 
   import Mox
 
-  alias Wanda.Messaging.Adapters.AMQP.Publisher
+  alias Wanda.Messaging.Adapters.AMQP.{
+    Consumer,
+    Publisher
+  }
+
+  alias Wanda.Support.Messaging.AMQPEventHandler
 
   setup [:set_mox_from_context, :verify_on_exit!]
 
   @moduletag :integration
+
+  setup_all do
+    pid = self()
+
+    :ok =
+      :telemetry.attach_many(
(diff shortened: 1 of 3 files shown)


test/wanda/messaging/adapters/amqp/consumer_test.exs
```diff
@@ -3,12 +3,42 @@ defmodule Wanda.Messaging.Adapters.AMQP.ConsumerTest do
...
+
+    start_supervised!(Publisher)
+    start_supervised!(Consumer)
+
+    assert_receive {:connected, :publisher}, 1_000
+    assert_receive {:connected, :consumer}, 1_000
+
+    :timer.sleep(300)
+
+    :ok
```
fabriziosestito, line 39: The trick of hooking up to the telemetry is brilliant.
However, do we really need this? Having the AMQP processes idle in our tests is very lightweight, and this seems more complex than what we had before. What is the rationale for this?

arbulu89: We had a tech-debt ticket for this. We can close both, I don't mind hehe

arbulu89:
Closing the PR as the solution is maybe more cumbersone than the current way.
Anyway, a new entry on the docs added explaining how to use telemetry events on tests:
https://github.com/trento-project/docs/pull/15

PR #181, opened 2023-02-08
Run CI `test` step on different versions of Elixir & OTP

jamie-suse:
This change aims to run the test step of the CI against both the versions of Elixir + OTP that are packaged with OBS/IBS and the one used for development.

Note that this change will need to be propagated through the other repos that use Elixir also.


fabriziosestito:
🥳

PR #187, opened 2023-02-15
Add new demo env that uses faked execution server

rtorrero:
Initial implementation of a faker to generate data for the checks execution. 

It achieves this by providing a implementation for the execution server that instead of triggering executions on agents uses the factories from the tests to generate such results.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -213,6 +213,45 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+  build-demo-img:
+    name: Build the docker image for the demo environment
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
+    needs: [static-code-analysis, test, test-e2e]
+    permissions:
+      contents: read
+      packages: write
+    env:
+      MIX_ENV: demo
+      REGISTRY: ghcr.io
+      IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/trento-wanda
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: docker/setup-buildx-action@v2
+      - name: Log in to the Container registry
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@507c2f2dc502c992ad446e3d7a5dfbe311567a96
+        with:
+          images: ${{ env.IMAGE_REPOSITORY }}
+      - name: Build and push container image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          push: true
+          tags: ${{ env.IMAGE_REPOSITORY }}:demo
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          build-args: MIX_ENV=demo
+
   generate-docs:
     name: Generate project documentation
     runs-on: ubuntu-20.04
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -215,9 +215,9 @@ jobs:
 
   build-demo-img:
     name: Build the docker image for the demo environment
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
-    needs: [static-code-analysis, test, test-e2e]
+    needs: [static-code-analysis, test]
     permissions:
       contents: read
       packages: write
@@ -246,11 +246,11 @@ jobs:
         with:
           context: .
           push: true
-          tags: ${{ env.IMAGE_REPOSITORY }}:demo
+          tags: ${{ env.IMAGE_REPOSITORY }}:${{ env.MIX_ENV }}
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          build-args: MIX_ENV=demo
+          build-args: MIX_ENV=${{ env.MIX_ENV }}
 
   generate-docs:
     name: Generate project documentation
--- a/config/demo.exs
+++ b/config/demo.exs
@@ -9,15 +9,6 @@ config :wanda, WandaWeb.Endpoint,
 
 config :logger, level: :debug
 
-config :wanda, Wanda.Messaging.Adapters.AMQP,
-  consumer: [
-    connection: "amqp://wanda:wanda@localhost:5672"
-  ],
-  publisher: [
-    connection: "amqp://wanda:wanda@localhost:5672"
-  ],
-  processor: Wanda.Messaging.Adapters.AMQP.Processor
-
 config :cors_plug,
   origin: [
     "http://localhost:4000",
--- a/config/demo.exs
+++ b/config/demo.exs
@@ -8,11 +8,3 @@ config :wanda, WandaWeb.Endpoint,
   server: true
 
 config :logger, level: :debug
-
-config :cors_plug,
-  origin: [
-    "http://localhost:4000",
-    "http://demo.trento-project.io",
-    "https://demo.trento-project.io"
-  ],
-  allow_credentials: true
--- a/config/demo.exs
+++ b/config/demo.exs
@@ -1,0 +1,27 @@
+import Config
+
+config :wanda, Wanda.Policy, execution_server_impl: Wanda.Executions.FakeServer
+
+config :wanda, WandaWeb.Endpoint,
+  check_origin: :conn,
+  cache_static_manifest: "priv/static/cache_manifest.json",
+  server: true
+
+config :logger, level: :debug
+
+config :wanda, Wanda.Messaging.Adapters.AMQP,
+  consumer: [
+    connection: "amqp://wanda:wanda@localhost:5672"
+  ],
+  publisher: [
+    connection: "amqp://wanda:wanda@localhost:5672"
+  ],
+  processor: Wanda.Messaging.Adapters.AMQP.Processor
+
+config :cors_plug,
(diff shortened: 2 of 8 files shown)


demo/fake_server.ex
```diff
@@ -0,0 +1,63 @@
...
+    Result
+  }
+
+  alias Wanda.Messaging
+
+  @impl true
+  def start_execution(execution_id, group_id, targets, env, config \\ []) do
+    create_fake_execution(execution_id, group_id, targets)
+    Process.sleep(2_000)
+    complete_fake_execution(execution_id, group_id, targets, :passing)
```
fabriziosestito, line 21: Can we randomize the result with [https://hexdocs.pm/elixir/1.12/Enum.html#random/1](https://hexdocs.pm/elixir/1.12/Enum.html#random/1) 

You can skip the function parameter and do it directly there

fabriziosestito (changes requested):

config/demo.exs
```diff
@@ -0,0 +1,27 @@
...
+config :wanda, Wanda.Policy, execution_server_impl: Wanda.Executions.FakeServer
+
+config :wanda, WandaWeb.Endpoint,
+  check_origin: :conn,
+  cache_static_manifest: "priv/static/cache_manifest.json",
+  server: true
+
+config :logger, level: :debug
+
+config :wanda, Wanda.Messaging.Adapters.AMQP,
```
arbulu89, line 12: Do we need this messaging configuration? It is the same as in the `config.exs` file

config/demo.exs
```diff
@@ -0,0 +1,27 @@
...
+  consumer: [
+    connection: "amqp://wanda:wanda@localhost:5672"
+  ],
+  publisher: [
+    connection: "amqp://wanda:wanda@localhost:5672"
+  ],
+  processor: Wanda.Messaging.Adapters.AMQP.Processor
+
+config :cors_plug,
```
arbulu89, line 21: I don't think we need this `cors_plug`. We can use the same as in config, and simply set some env variable to `CORS_ORIGIN`. It is even changed in runtime

demo/fake_server.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Wanda.Executions.FakeServer do
```
arbulu89, line 1: Do we need a `demo/fake_server.ex` file?
Why not simply put the `FakeServer` module in the `executions` folder

demo/fake_server.ex
```diff
@@ -0,0 +1,63 @@
...
+  def receive_facts(execution_id, group_id, agent_id, facts) do
+    :ok
+  end
+
+  defp create_fake_execution(execution_id, group_id, targets) do
+    insert(:execution,
+      status: :running,
+      execution_id: execution_id,
+      group_id: group_id,
+      targets: Enum.map(targets, &Map.from_struct/1)
```
arbulu89, line 36: We need to be careful with this 1:1 targets conversion.
We could fall in previous errors, like sending back check results that are not even in the catalog. You might want to visit how it works in the real execution.

arbulu89 (commented):
I find the current implementation too dump honestly.
We need to think on having certain logic like:
If the expectations are true -> the result of the check is true, etc
We cannot have failed expectations and passing results.
Now that we will implement the new view of the results using all these data, it would look totally illogical.

I think that we need to invest much more on this.
If you want to do in a subsequent PR is still good.

demo/fake_server.ex
rtorrero: Since FakeServer imports `Wanda.Factory` and `Wanda.Factory` is only available while testing, we need to play a bit with the paths in mix.exs to avoid always importing the Factory when it's not needed but allow it to be there when we are on the demo env. See `mix.exs`:

```
  # Specifies which paths to compile per environment.
  defp elixirc_paths(:test),
    do: [
      "lib",
      "test/support"
    ]

  defp elixirc_paths(:demo),
    do: [
      "test/support/factory.ex",
      "demo/fake_server.ex",
      "lib"
    ]
 ```

File: demo/fake_server.ex (hunk 5 above)
Comment (rtorrero) [reply demo/fake_server.ex:36]: Is this relevant for an environment in which we control what gets send? I don't think this code will be ever used outside of demo / development purposes

File: config/demo.exs (hunk 2 above)
Comment (rtorrero) [reply config/demo.exs:12]: You are right, I think we can let it go. Since I was making basically a clone of the prod env but with the faker, I copied this from there too. It probably means we can also remove this from `prod.exs`

File: config/demo.exs (hunk 3 above)
Comment (rtorrero) [reply config/demo.exs:21]: CORS_ORIGIN wasn't enough in this case, I had a conversation with @CDimonaco about it. Apparently we don't have this happening in prod as in our helm chart everything comes from the same domain.

Comment (arbulu89) [reply config/demo.exs:12]: But you could still use the `CORS_ORIGIN` and only leave the `allow_crendentials` value (even though I don't find this fields documented anywhere nor in the code).
I think having `http://demo.trento-project.io` should be enough, as the https is not even available in our demo

File: demo/fake_server.ex (hunk 5 above)
Comment (arbulu89) [reply demo/fake_server.ex:36]: well, we already saw errors here @rtorrero 
it is not about using the code somewhere else (which won't happen), it is about having a potential broken view if we receive some check id that it is not in the catalog.
Anyway, I don't want to extend this discussion. I see the risk. If you are fine with the solutions, so be it

File: demo/fake_server.ex (hunk 4 above)
Comment (arbulu89) [reply demo/fake_server.ex:1]: Ok, I see

**Review by arbulu89 — APPROVED** [review]
Thanks @rtorrero 
I still think that there are unfinished things, but I'm fine merging this first to at least enable the demo environment, and we can work on it later at some point

File: config/demo.exs (hunk 3 above)
Comment (arbulu89) [reply config/demo.exs:12]: I think this `allow_credentials` usage is a mistake and it is coming from other package usage, [corsica](https://github.com/whatyouhide/corsica)

File: demo/fake_server.ex (hunk 1 above)
Comment (arbulu89) [reply demo/fake_server.ex]: @fabriziosestito In our offline conversation, i just was saying, that we can have this execution result coming from the checks results, which now I see is fairly simple. It doesn't need to be random.
You just need to loop through the check results, check the result of them, and simply weight them (or some basic operation to find the worst check result).
Anyway, I'm fine on improving this on future iterations

File: config/demo.exs (hunk 3 above)
Comment (rtorrero) [reply config/demo.exs:12]: Fair enough, I'm currently working on building a demo container in the CI for deployment in our demo env. I'll revisit this issue and remove this if I can by using env vars where possible.

Comment (rtorrero) [reply config/demo.exs]: I've removed it for now as it's only needed for development. The demo environment when deployed through helm doesn't need it.

File: .github/workflows/ci.yaml [hunk 6]
```diff
@@ -213,6 +213,45 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+  build-demo-img:
+    name: Build the docker image for the demo environment
+    runs-on: ubuntu-latest
```
arbulu89, line 218: Let's use `runs-on: ubuntu-20.04` as in the others

.github/workflows/ci.yaml
```diff
@@ -213,6 +213,45 @@ jobs:
...
+    if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
+    needs: [static-code-analysis, test, test-e2e]
```
arbulu89, line 220: We need to adjust the `needs` here. Remove the `test-e2e`

.github/workflows/ci.yaml
```diff
@@ -213,6 +213,45 @@ jobs:
...
+        id: meta
+        uses: docker/metadata-action@507c2f2dc502c992ad446e3d7a5dfbe311567a96
+        with:
+          images: ${{ env.IMAGE_REPOSITORY }}
+      - name: Build and push container image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          push: true
+          tags: ${{ env.IMAGE_REPOSITORY }}:demo
```
arbulu89, line 249: I think we could use `MIX_ENV` as tag as well here, instead of hardcoded

.github/workflows/ci.yaml
```diff
@@ -213,6 +213,45 @@ jobs:
...
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          build-args: MIX_ENV=demo
```
arbulu89, line 253: Same, we could use the env variable

demo/fake_server.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  alias Wanda.Executions.{
+    Execution,
+    Result
+  }
+
+  alias Wanda.Messaging
+
+  @impl true
+  def start_execution(execution_id, group_id, targets, env, config \\ []) do
```
arbulu89, line 18: `config` and `env` are unused, we should replace them with `_`.

demo/fake_server.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  import Wanda.Factory
+
+  require Logger
+
+  alias Wanda.Executions
+
+  alias Wanda.Executions.{
+    Execution,
+    Result
```
arbulu89, line 12: These `Execution` and `Result` aliases are unused

demo/fake_server.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  defp complete_fake_execution(execution_id, group_id, targets) do
+    check_results = build(:check_results_from_targets, targets: targets)
+
+    build_result =
+      build(:result,
+        check_results: check_results,
+        execution_id: execution_id,
+        group_id: group_id,
+        result: Enum.random([:passing, :warning, :critical])
```
arbulu89, line 52: I would put this back to `passing` only, as the all the created checks have this result. Otherwise, it is misleading to see all checks green but the health with something else

arbulu89 (approved):
LGTM.
Some few typo,copy/paste and linting errors.
You need to add the `demo` folder to the `.formatter.exs` and `.credo.exs` file, to be included in the checks

demo/fake_server.ex
fabriziosestito: if you move the enum random to a result variable and pass it both to this factory and to the `:check_results_from_targets` you'll have consistent results, as the `:check_results_from_targets` gives consistent results (also the expectations are updated consistently)

fabriziosestito:
```elixir
result = Enum.random([:passing, :warning, :critical])
 check_results = build(:check_results_from_targets, targets: targets, result: result)

    build_result =
      build(:result,
        check_results: check_results,
        execution_id: execution_id,
        group_id: group_id,
        result: result
```

fabriziosestito (approved):

PR #188, opened 2023-02-17
Remotely trigger demo deploy on new wanda image

rtorrero:
This PR makes uses of [this GH action](https://github.com/peter-evans/repository-dispatch) to remotely trigger a new demo deployment in trento-web.

Should remain as draft:
 - [x] until the web repo includes code to handle such request.
 - [x] until the org (PAT) token has been created and added to the secrets:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -256,7 +256,7 @@ jobs:
         with:
           token: ${{ secrets.PAT }}
           repository: ${{ github.repository_owner }}/web
-          event-type: demo-deploy
+          event-type: deploy-demo
 
   generate-docs:
     name: Generate project documentation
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -251,6 +251,12 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           build-args: MIX_ENV=${{ env.MIX_ENV }}
+      - name: Remotely trigger trento-web demo deployment
```
arbulu89, line 254: We could maybe add an if condition to deploy only when we want:
```
if: vars.DEPLOY_DEMO == 'true'
```
This way we continue having control over deploying or not.

arbulu89 (approved):

PR #204, opened 2023-02-27
Facts schema value lists maps

arbulu89:
Add lists and maps to the `Fact` schema.
The free-form is represented without using properties or items: https://swagger.io/docs/specification/data-models/data-types/

I have included some property testing, mainly for sake of learning hehe
But they work fine. Anyway, if you think it is too much, i can remove it.


test/wanda_web/controllers/v1/execution_controller_test.exs
```diff
@@ -119,6 +120,35 @@ defmodule WandaWeb.V1.ExecutionControllerTest do
       assert_schema(json, "ExecutionResponse", api_spec)
     end
 
+    property "should accept all different types of fact values", %{conn: conn} do
```
fabriziosestito, line 123: Property testing is a very powerful tool, but here is a bit overkill.
The goal of property testing is to ensure our code is working properly under generated types of inputs that we wouldn't cover with normal tests manually.
Here the fact is that we are not really testing our code, as the property testing is proving that the `open_api_spex` library works correctly, which we already know.
Even if the data generation comes in handy, this runs `max_runs = 100` times, which also means 100 database writes, making the test suite slower.
Consider refactoring the test by using a fixed set of data generated with `faker` instead.

arbulu89: I agree that it can be overkill, but the test is indeed testing our code, which checks that the schema is properly written. Change the schema (remove that arrays are part of it) and you will see how the test fails :wink: 

About being slower, I don't at least notice the difference hehe

Anyway, I will change the and use fixed values!

fabriziosestito: Of course, this is testing our code as well, but do you need 100 iterations of generated strings to test that our code is returning something compliant with a schema? or the property that you are testing is more about the functionality of the Open API schema library (e.g. it still works with a string containing utf8 characters)

arbulu89: Yes, yes! I'm giving you the reason! I see that it is overkill. I just answered to:

> Here the fact is that we are not really testing our code, as the property testing is proving that the open_api_spex library works correctly, which we already know.

nelsonkopliku (approved):

fabriziosestito (approved):

PR #205, opened 2023-02-28
Enrich the faker by using catalog data

rtorrero:
This PR improves the faker so that it:
 - Facts now have the correct check_id and name from the corresponding check (fact values are still randomly generated though)
 - Expectation evaluation now also features real expectations from the actual check.
 - Values (expected values) are now also fetched from the catalog, assuming always the fallback / default provider.

### DIFF
--- a/demo/fake_evaluation.ex
+++ b/demo/fake_evaluation.ex
@@ -34,7 +34,13 @@ defmodule Wanda.Executions.FakeEvaluation do
   end
 
   defp build_check_result_from_target(check_id, check_targets, result, checks) do
-    check = Enum.find(checks, &(&1.id == check_id))
+    %{facts: facts, expectations: expectations, values: values} =
+      Enum.find(checks, &(&1.id == check_id))
+
+    expect_same_values =
+      Enum.into(expectations, %{}, fn %{name: name} ->
+        {name, Faker.StarWars.character()}
+      end)
 
     build(:check_result,
       check_id: check_id,
@@ -43,47 +49,48 @@ defmodule Wanda.Executions.FakeEvaluation do
           build(:agent_check_result,
             agent_id: agent_id,
             facts:
-              Enum.map(check.facts, fn %{name: name} ->
+              Enum.map(facts, fn %{name: name} ->
                 build(:fact, check_id: check_id, name: name)
               end),
-            expectation_evaluations: expectation_evaluations_from_check(check, result),
+            expectation_evaluations:
+              build_expectation_evaluations(expectations, result == :passing, expect_same_values),
             values:
-              Enum.map(check.values, fn %{name: name, default: value} ->
+              Enum.map(values, fn %{name: name, default: value} ->
                 %{name: name, value: value}
               end)
           )
         end),
       expectation_results:
         Enum.map(
-          expectation_evaluations_from_check(check, result),
+          expectations,
           &build(:expectation_result, name: &1.name, type: &1.type, result: result == :passing)
         ),
       result: result
     )
   end
 
-  defp expectation_evaluations_from_check(check, result) do
-    Enum.map(check.expectations, fn %{type: type, name: name} ->
+  defp build_expectation_evaluations(expectations, result, expect_same_values) do
+    Enum.map(expectations, fn %{name: name, type: type} ->
       case type do
         :expect ->
           build(:expectation_evaluation,
             type: :expect,
             name: name,
-            return_value: result == :passing
+            return_value: result
           )
 
         :expect_same ->
           build(:expectation_evaluation,
             type: :expect_same,
             name: name,
-            return_value:
-              if result == :passing do
-                "same_value"
-              else
-                Faker.StarWars.character()
-              end
+            return_value: build_expect_same_result(result, name, expect_same_values)
           )
       end
     end)
   end
+
+  defp build_expect_same_result(true, name, expect_same_values),
+    do: Map.get(expect_same_values, name)
+
+  defp build_expect_same_result(false, _, _), do: Faker.StarWars.character()
 end
--- a/demo/fake_evaluation.ex
+++ b/demo/fake_evaluation.ex
@@ -5,24 +5,6 @@ defmodule Wanda.Executions.FakeEvaluation do
 
   import Wanda.Factory
 
-  alias Wanda.Executions
-
-  @spec create_fake_execution(
-          String.t(),
-          String.t(),
-          [Wanda.Executions.Target.t()]
-        ) :: Wanda.Executions.Execution.t()
-  def create_fake_execution(execution_id, group_id, targets) do
-    insert(:execution,
-      status: :running,
-      execution_id: execution_id,
-      group_id: group_id,
-      targets: Enum.map(targets, &Map.from_struct/1)
-    )
-
-    Executions.create_execution!(execution_id, group_id, targets)
-  end
-
   @spec complete_fake_execution(
           String.t(),
           String.t(),
@@ -33,20 +15,12 @@ defmodule Wanda.Executions.FakeEvaluation do
     result = Enum.random([:passing, :warning, :critical])
     check_results = build_check_results_from_targets(targets, result, checks)
 
-    build_result =
-      build(:result,
-        check_results: check_results,
-        execution_id: execution_id,
-        group_id: group_id,
-        result: result
-      )
-
-    Executions.complete_execution!(
-      execution_id,
-      build_result
+    build(:result,
+      check_results: check_results,
+      execution_id: execution_id,
+      group_id: group_id,
+      result: result
     )
-
-    build_result
   end
 
   defp build_check_results_from_targets(targets, result, checks) do
--- a/demo/fake_evaluation.ex
+++ b/demo/fake_evaluation.ex
@@ -35,13 +35,6 @@ defmodule Wanda.Executions.FakeEvaluation do
 
   defp build_check_result_from_target(check_id, check_targets, result, checks) do
     check = Enum.find(checks, &(&1.id == check_id))
-    expectation_evaluations = expectation_evaluations_from_check(check, result)
-
-    expectation_results =
-      Enum.map(
-        expectation_evaluations,
-        &build(:expectation_result, name: &1.name, type: &1.type, result: result == :passing)
-      )
 
     build(:check_result,
       check_id: check_id,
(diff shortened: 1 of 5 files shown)


demo/fake_server.ex
```diff
@@ -56,22 +56,97 @@ defmodule Wanda.Executions.FakeServer do
...
+        expectation_evaluations,
+        &build(:expectation_result, name: &1.name, type: &1.type, result: result == :passing)
+      )
+
+    build(:check_result,
+      check_id: check_id,
+      agents_check_results:
+        Enum.map(check_targets, fn target ->
+          build(:agent_check_result,
+            agent_id: target.agent_id,
```
arbulu89, line 139: We shouldn't be using dot access here. We should be deconstructing in the `fn` clause:
```
Enum.map(check_targets, fn %{agent_id: agent_id} ->
```
And this applies to all the places where I see dot access

demo/fake_server.ex
```diff
@@ -56,22 +56,97 @@ defmodule Wanda.Executions.FakeServer do
...
+            facts:
+              Enum.map(check.facts, fn fact ->
+                build(:fact, check_id: check_id, name: fact.name)
+              end),
+            expectation_evaluations: expectation_evaluations,
+            values: check.values
```
arbulu89, line 145: How is this `check.values` working?
Is this just sending the same `values` field from the check itself?
We don't have to do this. The `values` are transformed during the evaluation. 
We should be sending something like:
```
values: [
{
name: faked_name,
value: some_value_from_the_check_values (it could be the default one)
}
]
```

demo/fake_server.ex
```diff
@@ -56,22 +56,97 @@ defmodule Wanda.Executions.FakeServer do
...
+
+    build(:check_result,
+      check_id: check_id,
+      agents_check_results:
+        Enum.map(check_targets, fn target ->
+          build(:agent_check_result,
+            agent_id: target.agent_id,
+            facts:
+              Enum.map(check.facts, fn fact ->
+                build(:fact, check_id: check_id, name: fact.name)
```
arbulu89, line 142: If we totally randomize the value, the `expect_same` will continue being super inconsistent.
As you will have different values and a passing result. Are we fine with that?

demo/fake_server.ex
```diff
@@ -56,22 +56,97 @@ defmodule Wanda.Executions.FakeServer do
...
+    |> Enum.uniq()
+    |> Enum.map(fn check_id ->
+      check_targets = Enum.filter(targets, &(check_id in &1.checks))
+      check_result_from_target(check_id, check_targets, result)
+    end)
+  end
+
+  defp expectation_evaluations_from_check(check, result) do
+    Enum.map(check.expectations, fn expectation ->
+      case expectation.type do
```
arbulu89, line 101: Avoid dot access

demo/fake_server.ex
```diff
@@ -56,22 +56,97 @@ defmodule Wanda.Executions.FakeServer do
...
+            type: :expect,
+            name: expectation.name,
+            return_value: result == :passing
+          )
+
+        :expect_same ->
+          build(:expectation_evaluation,
+            type: :expect_same,
+            name: expectation.name,
+            return_value:
```
arbulu89, line 113: What for is this `if/esle` with the hardcoded the `same_value`?
As far as I see, the value will be the same in all cases, as the the same expectation_evaluation is attached to all targets.
Maybe we could simply put the faked value as `return_value`.

arbulu89 (changes requested):
Hey @rtorrero 
Some comments from my side

demo/fake_server.ex
```diff
@@ -56,22 +56,97 @@ defmodule Wanda.Executions.FakeServer do
...
         execution_id: execution_id,
-        group_id: group_id,
-        result: result
+        group_id: group_id
       )
 
     Executions.complete_execution!(
       execution_id,
-      build_result
+      Map.put(build_result, :result, get_result(build_result))
```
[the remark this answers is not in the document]
rtorrero: My logic here was that:
```
    check_results = check_results_from_targets(targets, result)

    build_result =
      build(:result,
        check_results: check_results,
        execution_id: execution_id,
        group_id: group_id
      )
``` would get me a check_results inside the results, but that the result itself wouldn't be consistent with the content of the checks_results, because I see in the factory:
```
  def result_factory do
    %Result{
      execution_id: UUID.uuid4(),
      group_id: UUID.uuid4(),
      check_results: build_list(5, :check_result),
      result: :passing
    }
  end
  ```
  
where result is set to `:passing` regardless of the checks_results it contains. I think I'm probably not understanding the meaning of that `result: :passing`

Should I just remove the `get_result` and `result_weight` functions?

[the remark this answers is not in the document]
arbulu89: It would be:
```
build_result =
      build(:result,
        check_results: check_results,
        execution_id: execution_id,
        group_id: group_id,
        result: result
      )
```
Simply, because the generated expectations and results are based already in the previously randomized result, so we don't need to evaluate again.

PD: yes, remove `get_result` and `result_wight`

demo/fake_server.ex
```diff
@@ -56,22 +56,97 @@ defmodule Wanda.Executions.FakeServer do
...
+              else
+                Faker.StarWars.quote()
+              end
+          )
+      end
+    end)
+  end
+
+  defp check_result_from_target(check_id, check_targets, result) do
+    {_, check} = Catalog.get_check(check_id)
```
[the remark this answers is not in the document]
fabriziosestito: > In my point of view it would be better to pass the checks from top to button instead of getting them here.

I agree with this, we already load checks in the main function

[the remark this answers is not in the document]
fabriziosestito: > @arbulu89 PD: I won't enter if this code should be tested or not (I think it should be)

I would test the entire module as black-box and keep those functions private, which means basically starting and execution which certain targets/checks and assert on the side-effects (message published through the mock/execution saved in the database).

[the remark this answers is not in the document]
arbulu89: Yes, I meant that. We don't need to specifically need to test this function, but the whole faked execution code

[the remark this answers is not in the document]
rtorrero: Got it.

test/demo/fake_server_test.exs
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  alias Wanda.Executions.{
+    Execution,
+    FakeServer
+  }
+
+  setup :verify_on_exit!
+
+  test "start_execution publishes execution started and completed messages" do
```
arbulu89, line 19: I think we should mock the `sleep` function, otherwise this test will take 2 seconds everytime, which is already more than the whole suite

rtorrero:
@arbulu89 It took me a bit to see what you saw but now the expectation results are generated for newly for each agent. That way it is possible to control when the values should be the same (passing case) and when they should be different.

Let me know if it looks better now :+1:

demo/fake_evaluation.ex
```diff
@@ -0,0 +1,89 @@
...
+    check = Enum.find(checks, &(&1.id == check_id))
+
+    build(:check_result,
+      check_id: check_id,
+      agents_check_results:
+        Enum.map(check_targets, fn %{agent_id: agent_id} ->
+          build(:agent_check_result,
+            agent_id: agent_id,
+            facts:
+              Enum.map(check.facts, fn %{name: name} ->
```
arbulu89, line 46: I still see this dot access (`checks.facts` and `checks.values`, and others around the code) here, which we could get rid off with proper pattern matching)

demo/fake_evaluation.ex
```diff
@@ -0,0 +1,89 @@
...
+        Enum.map(
+          expectation_evaluations_from_check(check, result),
+          &build(:expectation_result, name: &1.name, type: &1.type, result: result == :passing)
+        ),
+      result: result
+    )
+  end
+
+  defp expectation_evaluations_from_check(check, result) do
+    Enum.map(check.expectations, fn %{type: type, name: name} ->
```
arbulu89, line 66: Dot access

demo/fake_evaluation.ex
```diff
@@ -0,0 +1,89 @@
...
+            expectation_evaluations: expectation_evaluations_from_check(check, result),
+            values:
+              Enum.map(check.values, fn %{name: name, default: value} ->
+                %{name: name, value: value}
+              end)
+          )
+        end),
+      expectation_results:
+        Enum.map(
+          expectation_evaluations_from_check(check, result),
```
arbulu89, line 58: I don't think you need this `expectation_evaluations_from_check` here. You could loop on the `check.expectations` (without using dot access please)

demo/fake_evaluation.ex
```diff
@@ -0,0 +1,89 @@
...
+            name: name,
+            return_value: result == :passing
+          )
+
+        :expect_same ->
+          build(:expectation_evaluation,
+            type: :expect_same,
+            name: name,
+            return_value:
+              if result == :passing do
```
arbulu89, line 80: Last time I comment on this XD
I still think that we could improve this easily to avoid this `if/else` and a hardcoded `same_value`, just generating the value outside of this function depending on the result, and passing it here, which would just set the value.

test/demo/fake_evaluation_test.exs
```diff
@@ -0,0 +1,127 @@
...
+  describe "complete_fake_execution/4" do
+    test "should complete a running fake execution" do
+      [
+        %Fact{
+          name: fact_name_1
+        },
+        %Fact{
+          name: fact_name_2
+        }
+      ] = catalog_facts = build_list(2, :catalog_fact)
```
arbulu89, line 28: Why not `simply` facts? as the next is named `values`

test/demo/fake_evaluation_test.exs
```diff
@@ -0,0 +1,127 @@
...
+          id: check_id_1
+        },
+        %Check{
+          id: check_id_2
+        }
+      ] = checks = build_list(2, :check, facts: catalog_facts, values: values)
+
+      [
+        %Execution.Target{
+          agent_id: agent_id
```
arbulu89, line 50: Can we add at least other agent?
At the end, the major part of the logic consists on doing things for 2 agents, which this doesn't cover

arbulu89 (changes requested):
Hey @rtorrero ,
Good enough from my side if we improve the tests. Everything else is more cosmetical

test/demo/fake_evaluation_test.exs
rtorrero: The idea was to make it more obvious that this is about a `%Catalog.Fact{}` and not a `%Wanda.Executions.Fact{}`. Is this already obvious and not worth?

arbulu89: My ocd detector simply raises because you use `catalog_facts/values` combo instead of `facts/values` or `catalog_facts/catalog_values`

arbulu89 (approved):
Thanks @rtorrero 
All good from my side

PR #217, opened 2023-03-14
Refactor API errors

fabriziosestito:


lib/wanda_web/controllers/fallback_controller.ex
```diff
@@ -0,0 +1,19 @@
...
+  alias WandaWeb.ErrorView
+
+  def call(conn, {:error, :no_checks_selected}) do
+    conn
+    |> put_status(:unprocessable_entity)
+    |> put_view(ErrorView)
+    |> render(:"422", reason: "No checks were selected.")
+  end
+
+  def call(conn, {:error, :already_running}) do
```
arbulu89, line 13: Should we have a "just in case" error here?
If for some strange reason the `DynamicSupervisor` fails to start or so.
I know that you stretched the possible errors in the spec above hehe

lib/wanda_web/views/error_view.ex
```diff
@@ -1,24 +1,39 @@
...
+      ]
+    }
   end
 
   # By default, Phoenix returns the status message from
   # the template name. For example, "404.json" becomes
   # "Not Found".
   def template_not_found(template, _assigns) do
-    %{errors: %{detail: Phoenix.Controller.status_message_from_template(template)}}
+    %{
```
arbulu89, line 30: We could remove/update the comment above this. It looks a leftover

arbulu89 (approved):
:rocket: :rabbit:

lib/wanda_web/controllers/fallback_controller.ex
fabriziosestito: done, I've also added a test that covers the generic error case.
I've omitted specific errors tests, since we cover them from the execution controller tests.

arbulu89 (approved):

PR #220, opened 2023-03-17
Fix execution of SBD related checks  (jsc#CFSA-1961)

angelabriel:
Skip the SBD related checks on the providers 'aws' and 'gcp' instead of executing the checks ONLY on the provider 'azure', 'kvm' or 'nutanix' as we do today. This gives us the possibility to run the SBD related checks on a (for trento) currently 'unknown' provider too.

### DIFF
--- a/priv/catalog/0B6DB2.yaml
+++ b/priv/catalog/0B6DB2.yaml
@@ -48,7 +48,7 @@ remediation: |
     - https://documentation.suse.com/sle-ha/15-SP3/html/SLE-HA-all/cha-ha-storage-protect.html#pro-ha-storage-protect-sbd-config
     - https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#id-adapting-the-sbd-configuration
 
-when: env.provider in ["azure", "nutanix", "kvm"]
+when: env.provider != "gcp" & env.provider != "aws"
 
 facts:
   - name: sbd_pacemaker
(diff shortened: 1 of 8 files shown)


priv/catalog/0B6DB2.yaml
```diff
@@ -48,7 +48,7 @@ remediation: |
     - https://documentation.suse.com/sle-ha/15-SP3/html/SLE-HA-all/cha-ha-storage-protect.html#pro-ha-storage-protect-sbd-config
     - https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#id-adapting-the-sbd-configuration
 
-when: env.provider in ["azure", "nutanix", "kvm"]
+when: env.provider != "gcp" & env.provider != "aws"
```
arbulu89, line 51: What about `env.provider !in ["gcp", "aws"]`? (notice the `!` before the `in`).
It looks more beautiful and scalable if we want to change the array

https://rhai.rs/book/language/in.html

angelabriel: yes, it's simpler to enhance. Thanks

arbulu89 (approved):
Thanks @angelabriel 
Fine to merge from my side, if you are finished with the changes of course hehe

angelabriel:
@arbulu89 Thanks for the review
Yes, I'm finished with the changes. Please go ahead with the merge

nelsonkopliku (approved):

PR #225, opened 2023-03-27
Add healthcheck and readiness endpoints

jamie-suse:
Related to https://github.com/trento-project/web/pull/1286

# Description

Adds `/api/healthz` and `/api/readyz` endpoints to get the health status and readiness of the Trento Web server respectively.

In this PR, the healthcheck only checks that the database/Repo is online and responding. A follow up PR will add a check for the RabbitMQ Consumer and Publishers.

## How was this tested?

By adding integration tests for each endpoint

### DIFF
--- a/lib/wanda_web/controllers/health_controller.ex
+++ b/lib/wanda_web/controllers/health_controller.ex
@@ -1,0 +1,47 @@
+defmodule WandaWeb.HealthController do
+  use WandaWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias Ecto.Adapters.SQL
+
+  alias WandaWeb.Schemas.{
+    Health,
+    Ready
+  }
+
+  operation :ready,
+    summary: "Wanda ready",
+    tags: ["Platform"],
+    description: "Check if Wanda is ready",
+    security: [],
+    responses: [
+      ok: {"Wanda is ready", "application/json", Ready}
+    ]
+
+  def ready(conn, _) do
+    conn
+    |> put_status(200)
+    |> json(%{ready: true})
+  end
+
+  operation :health,
+    summary: "Wanda health",
+    tags: ["Platform"],
+    description: "Get the health status of the Wanda platform",
+    security: [],
+    responses: [
+      ok: {"Wanda health status", "application/json", Health}
+    ]
+
+  def health(conn, _) do
+    db_status =
+      case SQL.query(Wanda.Repo, "SELECT 1", []) do
+        {:ok, _} -> :pass
+        {:error, _} -> :fail
+      end
+
+    conn
+    |> put_status(if db_status == :pass, do: 200, else: 500)
+    |> render("health.json", health: %{database: db_status})
+  end
+end
--- a/lib/wanda_web/views/health_view.ex
+++ b/lib/wanda_web/views/health_view.ex
@@ -1,0 +1,5 @@
+defmodule WandaWeb.HealthView do
+  use WandaWeb, :view
+
+  def render("health.json", %{health: health}), do: health
+end
--- a/lib/wanda_web/router.ex
+++ b/lib/wanda_web/router.ex
@@ -32,6 +32,12 @@ defmodule WandaWeb.Router do
     get "/openapi", OpenApiSpex.Plug.RenderSpec, []
   end
 
+  scope "/api", WandaWeb do
+    pipe_through :api
+    get "/healthz", HealthController, :health
+    get "/readyz", HealthController, :ready
+  end
+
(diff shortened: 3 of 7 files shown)


lib/wanda_web/controllers/health_controller.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  operation :healthcheck,
+    summary: "Wanda Web Healthcheck",
+    tags: ["Platform"],
+    description: "Get the health status of the Wanda Web platform",
+    security: [],
+    responses: [
+      ok:
+        {"User credentials", "application/json",
+         %Schema{
```
arbulu89, line 20: Maybe we could move this to the schemas folder

lib/wanda_web/controllers/health_controller.ex
```diff
@@ -0,0 +1,47 @@
+defmodule WandaWeb.HealthController do
+  @moduledoc """
+    Health-related functions
+  """
+
+  use WandaWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias Ecto.Adapters.SQL
```
arbulu89, line 9: Instead of aliasing this, did you try `import Ecto.Query` ?

lib/wanda_web/views/health_view.ex
```diff
@@ -0,0 +1,10 @@
+defmodule WandaWeb.HealthView do
+  use WandaWeb, :view
```
arbulu89, line 2: Now hardly will break, but can we have a test on this?

lib/wanda_web/views/health_view.ex
```diff
@@ -0,0 +1,10 @@
...
+
+  def render(
+        "health.json",
+        %{health: %{database: db_status}}
```
arbulu89, line 6: you could simply `%{health: health}), do: health`

arbulu89 (commented):
Some initial comments

lib/wanda_web/controllers/health_controller.ex
jamie-suse: I don't see a way to construct the query `SELECT 1` using the `Ecto.Query` DSL. It seems that you need to provide a scheme to `SELECT` from.

lib/wanda_web/controllers/health_controller.ex
```diff
@@ -0,0 +1,47 @@
...
+  alias OpenApiSpex.Schema
+
+  operation :healthcheck,
+    summary: "Wanda Web Healthcheck",
+    tags: ["Platform"],
+    description: "Get the health status of the Wanda Web platform",
+    security: [],
+    responses: [
+      ok:
+        {"User credentials", "application/json",
```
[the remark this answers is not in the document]
jamie-suse: It should be a description of the response object, changing it now.

https://spec.openapis.org/oas/v3.1.0#response-object

lib/wanda_web/controllers/health_controller.ex
```diff
@@ -0,0 +1,47 @@
...
+  alias Ecto.Adapters.SQL
+  alias OpenApiSpex.Schema
+
+  operation :healthcheck,
+    summary: "Wanda Web Healthcheck",
+    tags: ["Platform"],
+    description: "Get the health status of the Wanda Web platform",
+    security: [],
+    responses: [
+      ok:
```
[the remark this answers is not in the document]
jamie-suse: Good catch, I missed this 🤦‍♂️

The references I found use either `500 Internal Server Error` or `503 Service Unavailable` for an unhealthy state. I think `503` is more accurate and informative though.

https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-7.0#customize-the-http-status-code

https://developers.redhat.com/sites/default/files/blog/2017/11/microprofile-health-spec.pdf (Appendix A: REST interface specifications, Status Codes)

lib/wanda_web/views/health_view.ex
jamie-suse: Added

rtorrero (approved):
LGTM @jamie-suse ! I told you yesterday that I may have used `:passing` instead of `:healthy` but a quick search reveals that it seems more common to use `healthy` as you did.

I've also seen that it seems to be common that the endpoint for the readyness is under `/health`, e.g.: `/api/v1/health/ready`. See https://testfully.io/blog/api-health-check-monitoring/#json-response for more details. But feel free to defer this to a future PR

arbulu89 (approved):
Thanks @jamie-suse 
Green light from my side

PR #226, opened 2023-03-28
add api versioning to the readme examples

angelabriel:
With the current wanda implementation the path to communicate with the wanda api is versioned, so the examples in the documentation need an additional 'v1' part to work as expected.


nelsonkopliku (approved):
Thanks! LGTM

PR #230, opened 2023-04-04
Add failure message documentation

dottorblaster:
Just adding some failure message documentation.

### DIFF
--- a/guides/specification.md
+++ b/guides/specification.md
@@ -12,7 +12,7 @@ Checks are, in Trento's domain, the crystallization of SUSE's best practices whe
 
 ## Checks Execution
 
-*Checks Execution* is the process that determines whether the best practices defined in the [Checks Specifications](#anatomy-of-a-check) are being followed on a target infrastructure.
+_Checks Execution_ is the process that determines whether the best practices defined in the [Checks Specifications](#anatomy-of-a-check) are being followed on a target infrastructure.
 
 > [Requesting an Execution](#requesting-an-execution) -> [Facts Gathering](#facts-gathering) -> [Expectation Evaluation](#expectation-evaluation)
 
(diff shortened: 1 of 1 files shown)


dottorblaster:
Formatting the file with `prettier` reformatted some of the existing lines. Should I keep it this way or revert it?

guides/specification.md
```diff
@@ -513,6 +516,32 @@ Considering the previous scenario what happens is that:
 >     expect_same: facts.installed_rpm_version
 > ```
 
+### failure_message
+
+An optional failure message can be declare for every expectation.
```
arbulu89, line 521: `declare` -> `declared`

guides/specification.md
```diff
@@ -513,6 +516,32 @@ Considering the previous scenario what happens is that:
...
+
+In case of an `expect` one, the failure message can interpolate `facts` and `values` present in the evaluation scope to provide more meaningful insights:
```
arbulu89, line 523: I would put "of an `expect` expectation".
I know it looks redundant, but is reads better than with `one`.

Besides that, `evalution scope` looks a bit technical. What about? `can interpolate facts and values present in the check definitition`

guides/specification.md
```diff
@@ -513,6 +516,32 @@ Considering the previous scenario what happens is that:
...
+
+```yaml
+expectations:
+  - name: awesome_expectation
+    expect: values.awesome_constant_value == facts.awesome_fact
+    failure_message: The expectation did not match ${values.awesome_constant_value}
+```
+
+The outcome of the interpolation will be available into `ExpectationEvaluation` inside the API response.
```
arbulu89, line 532: Not a big a fan of future term. I would rephrase to `... interpolation is available into`

guides/specification.md
```diff
@@ -513,6 +516,32 @@ Considering the previous scenario what happens is that:
...
+In case of an `expect_same` one, the failure message has to be a plain string:
+
+```yaml
+expectations:
+  - name: awesome_expectation
+    expect_same: facts.awesome_fact
+    failure_message: Boom!
+```
+
+This plain string will be available into `ExpectationResult` inside the API response.
```
arbulu89, line 543: The same for this `will`

arbulu89 (approved):
Great @dottorblaster ,
One simple typo to fix.
Everything else is personal preference.
PD: I'm personally fine with the prettier output
PD2: Before I forget, and as I don't know where to put. We should update the `demo` faker code so it sends the failure messages if it doesn't now. Maybe just changing the factories would do it

PR #235, opened 2023-04-20
enhance remediation section to clarify the value setting

angelabriel:
because of some confusion during the check testing we decided to enhance the remediation section to clarify the value settings and the behavior of the expectation.
This fixes https://jira.suse.com/browse/TRNT-1078

### DIFF
--- a/priv/catalog/373DB8.yaml
+++ b/priv/catalog/373DB8.yaml
@@ -6,7 +6,9 @@ description: |
 remediation: |
   ## Abstract
   The fencing timeout (`stonith-timeout`) determines the time Pacemaker will wait for fencing to succeed.
-  The recommended values on Azure are `144` seconds for SBD only or `900` seconds when using SBD combined with the Azure Fence agent.
+  The remommended default value is at least `150` seconds for SBD.
+  The recommended values on Azure are at least `144` seconds for SBD only or `900` seconds when using SBD combined with the Azure Fence agent.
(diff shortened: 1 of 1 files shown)


priv/catalog/373DB8.yaml
```diff
@@ -6,7 +6,9 @@ description: |
 remediation: |
   ## Abstract
   The fencing timeout (`stonith-timeout`) determines the time Pacemaker will wait for fencing to succeed.
-  The recommended values on Azure are `144` seconds for SBD only or `900` seconds when using SBD combined with the Azure Fence agent.
+  The remommended default value is at least `150` seconds for SBD.
+  The recommended values on Azure are at least `144` seconds for SBD only or `900` seconds when using SBD combined with the Azure Fence agent.
```
arbulu89, line 10: @angelabriel I would super quickly rewrite this to be like:
```
The recommended values on Azure...
The recommended value on AWS..
The recommended value on GCP...
```

Otherwise it looks like AWS and GCP are kind of coming together and look like less important than Azure

arbulu89 (commented):
Hi @angelabriel ,
Small suggestion, we can merge afterwards

priv/catalog/373DB8.yaml
angelabriel: Thanks for the suggestion. We don't want to hurt any of our partners.

arbulu89 (approved):

PR #248, opened 2023-06-01
Document cluster type support

arbulu89:
Document `cluster_type` env variable support.

Besides that, I have fixed a corner case issue, if some day we decide to send a `null` value in an env entry (unlikely, but hey, it won't break hehe)


abravosuse (approved):
Looks good to me!

nelsonkopliku (approved):

dottorblaster (approved):
LGTM

CDimonaco (approved):

fabriziosestito (approved):

PR #253, opened 2023-06-12
Add when conditions for resource types, propagate the resource type in the ExecutionCompleted event

dottorblaster:
This PR:

- Adds the resource type (`cluster`) when condition to all the checks
- Propagates the resource type through the ExecutionCompleted event so that the web can route a command to the right aggregate


nelsonkopliku (approved):
Nice job! Insta-approve :smile: 
I was already looking at it.

PR #256, opened 2023-06-16
Document target_type inside env

dottorblaster:
Just adding a row to the table that documents what `env` can contain.

### DIFF
--- a/guides/specification.md
+++ b/guides/specification.md
@@ -584,10 +584,12 @@ Scopes are namespaced and access to items in the scope is name based.
 `env` is a map of information about the context of the running execution, it is set by the system on each execution/check compilation.
 
 Examples of entries in the scope. What is actually available during the execution depends on the scenario. Find the updated values in the reference column link.
+
 | name | Type | Reference
 | ---- | -----| ----------
 | `env.provider` | one of `azure`, `aws`, `gcp`,`kvm`,`nutanix`, `vmware`, `unknown` | [Providers](https://github.com/trento-project/web/blob/main/lib/trento/domain/enums/provider.ex)
 | `env.cluster_type` | one of `hana_scale_up`, `hana_scale_out`, `ascs_ers`, `unknown` | [Cluster types](https://github.com/trento-project/web/blob/main/lib/trento/domain/enums/cluster_type.ex)
+| `env.target_type` | one of `cluster`, `host` | No enum available
 
 #### **facts**
 


guides/specification.md
```diff
@@ -584,10 +584,12 @@ Scopes are namespaced and access to items in the scope is name based.
 `env` is a map of information about the context of the running execution, it is set by the system on each execution/check compilation.
 
 Examples of entries in the scope. What is actually available during the execution depends on the scenario. Find the updated values in the reference column link.
+
 | name | Type | Reference
 | ---- | -----| ----------
 | `env.provider` | one of `azure`, `aws`, `gcp`,`kvm`,`nutanix`, `vmware`, `unknown` | [Providers](https://github.com/trento-project/web/blob/main/lib/trento/domain/enums/provider.ex)
 | `env.cluster_type` | one of `hana_scale_up`, `hana_scale_out`, `ascs_ers`, `unknown` | [Cluster types](https://github.com/trento-project/web/blob/main/lib/trento/domain/enums/cluster_type.ex)
+| `env.resource_type` | one of `cluster`, `host` | No enum available
```
nelsonkopliku, line 592: I think we should communicate somewhere that the `env.resource_type` is somewhat mandatory.

At the end it is safe to say that we expect checks to have the `when` condition with at least the `env.resource_type == "something"` expression.

Then we might also want to update the start api example doc, but that could be part of a follow up pr when we actually implement starting an execution with the resource type.

arbulu89: This is a now a new topic, but since we are trying to generalize the usage of the env without really adding this logic to the code, I don't think we should make it mandatory.
Actually, I'm thinking that we should update our contract messages and return back the whole `env` so the senders decides what to do with that.
Otherwise, the logic of not wanting a new field in the `dsl` because it was not generic, but adding in contract messages loses a bit of context in my opinion.

arbulu89, line 592: Shouldn't be `target_type`?

nelsonkopliku: Good points. Indeed `target_type` is not mandatory from a purely wanda perspective.

However how should we (if we should) deal with the fact that in the broader picture of trento we actually do expect checks to have the when condition with at least the `env.resource_type == "something"` expression? 
Maybe a note in this regard would be inappropriate in wanda as such, but what about the larger scope? :thinking: 

BTW: I like the idea of propagating the whole env received at execution start, regardless of who's gonna react to an emitted message (it could be the _sender_, yes, but not limited to that I guess). I would defer this, tho.

arbulu89: Yes, I'm not against adding a comment in the docs. It looks appropriate. My comment was now more about the engineering itself hehe
Of course we can defer.

dottorblaster: I think you're right @arbulu89, do you folks think we should collect something in the backlog?

arbulu89: Hard to say :sweat_smile: 
I think we are realizing about these things on the fly.
It deserves some thinking anyway.
Luckily the changes are simple to do

dottorblaster: I agree a lot with @arbulu89, maybe we can discuss this having a beer :smile:

Meanwhile I updated the PR with the right field.

arbulu89 (approved):

EMaksy (approved):
LGTM :D

nelsonkopliku (approved):

PR #258, opened 2023-06-23
Update contracts usage

nelsonkopliku:
# Description

Bumped contracts reference and updated deprecated usages of `new(!)` functions.
Relates to https://github.com/trento-project/contracts/pull/47

Additionally makes sure to provide a default value fof `target_type` in the evaluation expressions (cluster by now) until web starts publishing the proper `target_type` in the execution requested event.


lib/wanda/executions/server.ex
```diff
@@ -36,6 +36,9 @@ defmodule Wanda.Executions.Server do
   """
   @impl true
   def start_execution(execution_id, group_id, targets, env, config \\ []) do
+    # TODO: this should be removed when web properly sets the target type in ExecutionRequested event
```
arbulu89, line 39: Personally I wouldn't include this little change here, as it is a bit hidden.
But it is really small, and I understand the incovenience of creating a PR for only this.
Anyway, are we in the need to add this "hidden" step instead of having the actual implementation?

arbulu89 (approved):
All good from my side!

lib/wanda/executions/server.ex
nelsonkopliku: Thing is that if we don't provide the target type when building the ExecutionCompleted event, it would break.

The actual implementation where web sets target type in ExecutionRequested event is part of another story (actually starting a host check execution), which we're gonna address in next iteration.

Just a matter of scope. Here we we're focusing on allowing the selection of host checks. Next we'll introduce execution of host checks.

Would this work?

arbulu89: Alright, so "yes", we need it to move the PR forward hehe
Get it!
Yes, go on. I guess it will be changed really soon

jamie-suse (approved):

PR #259, opened 2023-06-27
initial checks for VMware vSphere (jsc#TRNT-1682)

yeoldegrove:
initial checks for VMware vSphere (jsc#TRNT-1682)

### DIFF
--- a/priv/catalog/00081D.yaml
+++ b/priv/catalog/00081D.yaml
@@ -42,6 +42,10 @@ remediation: |
 
     - https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#id-example-for-etccorosynccorosync-conf
 
+  VMware vSphere:
+
+    - https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#id-example-for-etccorosynccorosync-conf
+
 when: env.target_type == "cluster"
 
 facts:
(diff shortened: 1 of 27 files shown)


nelsonkopliku (approved):
Thanks @yeoldegrove!

arbulu89 (commented):
Hey @yeoldegrove ,
Glad to see you back :muscle: 
Just a question. I just see documentation changes without any change in the values.
Does this mean that the `default` values are the correct ones?
In the other hand, we never used the `vSphere` term at least internally. Is this the correct term? Asking to @abravosuse as well

PD: You didn't need to add each change in one commit hehe one commit would be good enough

yeoldegrove:
@arbulu89 I doubled checked the term `VMware vSphere` with @gereonvey ... If someone comes up with another term, we can change it, sure. I found it e.g. here: https://wiki.scn.sap.com/wiki/display/VIRTUALIZATION/SAP+HANA+on+VMware+vSphere

There is no publicly available documentation from VMware nor SUSE for the HA side of things on SAP systems. So we went for the `default` values for now.

gereonvey:
VMware vSphere is the correct/official product name (https://www.vmware.com/products/vsphere.html) for the platform that SAP systems would be running on, so we should use that.

abravosuse:
@arbulu89 when you talk about SAP on VMware vSphere you normally just say "VMware". So we stick to that name.

abravosuse:
VMware is good for me. And I believe that other vendors/providers/platforms are all spelt out in lowercase.

yeoldegrove:
@abravosuse `env.provider` is spelled in lowercase. The other vendors in the `remediation` section have very different spelling like `Azure, Nutanix, SUSE / KVM`, so `VMware` would look good there in my opinion.
I added [5312e01](https://github.com/trento-project/wanda/pull/259/commits/5312e017ae772a4a88626a86594e4c8871579cfe) which changes this in this way.

nelsonkopliku (approved):

arbulu89 (approved):

arbulu89:
@yeoldegrove All good from my side. Just thumbs up here if it is ready to be merged!

PR #268, opened 2023-07-13
Add a basic host check

nelsonkopliku:
For demo purposes.

### DIFF
--- a/priv/catalog/TRNT01.yaml
+++ b/priv/catalog/TRNT01.yaml
@@ -1,0 +1,29 @@
+id: "TRNT01"
+name: A host check
+group: Trento
+description: |
+  Trento agent service is enabled and running
+remediation: |
+  ## Abstract
+  If not enabled, trento-agent service will not start automatically after reboots, affecting Trento's functionalities.
+
+  To enable the service, run:
+  ```
+  systemctl enable trento-agent
+  ```
+
+when: env.target_type == "host"
+
+facts:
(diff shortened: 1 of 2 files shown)


EMaksy (commented):
Hey the check by itself is great!
For development this will be very useful, we may still need to discuss if want to mix real checks with fake checks.
So code wise LGTM, let's hear what the rest thinks about including it or not :D

abravosuse (approved):
From the understanding I have about the check yaml, this looks good to me Nelson...

dottorblaster (commented):
I'd rather not mix fake and real checks.

nelsonkopliku:
> I'd rather not mix fake and real checks.

I promise I will delete it as soon as we have a _real_ host check.

Think of it as as a basic noop :sweat_smile:

arbulu89:
I'm with @dottorblaster on this.
Why do we need to push this fake check?
Just to have the demo web updated? I though that the checks in the frontend would be disabled as we do with `ascs/ers` cluster types, with some sort of `Coming next` banner. Until we have something "real" to deliver

nelsonkopliku:
Yes, we'll have the coming soon communication.

Nevertheless before adding the coming soon, we'll deliver a working feature that might be interesting testing.

If we don't like this I can ditch it. It will be just a bit more cumbersome testing host checks without the support of an engineer.

No hard feelings :smile:

PR #279, opened 2023-08-04
Load fake facts config from proper location

nelsonkopliku:
A missing part that didn't allow loading the demo config file.


EMaksy (approved):
Ah nice, thanks for fixing!  --> LGTM :+1:

PR #284, opened 2023-08-24
Add target_type

dottorblaster:
Just adding target_type from the events and in the execution endpoint


test/wanda/executions/server_test.exs
```diff
@@ -112,7 +112,7 @@ defmodule Wanda.Executions.ServerTest do
       )
 
       assert {:error, :already_running} =
-               Server.start_execution(UUID.uuid4(), group_id, targets, %{})
+               Server.start_execution(UUID.uuid4(), group_id, targets, :cluster, %{})
```
nelsonkopliku, line 115: we won't be getting atoms, rather strings. Just `"cluster"` should be enough.

nelsonkopliku (commented):
Looking great, thanks!

Just a couple of comments.

nelsonkopliku (approved):
LGTM.
So if we get no target type upon request, then a `nil` is passed down to contract messages. It should not break anything with target type being a `string target_type`, right?

dottorblaster:
It shouldn't, correct :+1:

PR #297, opened 2023-09-27
Handle gathered fact nil value

arbulu89:
Handle `nil` gathered value. Otherwise, it was stored as string with `:NULL_VALUE` value.
In rhai, we can use `facts.value == ()` to check if it is null

Related to: https://github.com/trento-project/agent/pull/263


dottorblaster (approved):
#absolutelyunrequestedreviews

LGTM

nelsonkopliku (approved):
LGTM

EMaksy (approved):
LGTM!

PR #300, opened 2023-09-28
Fix cache-miss in CI pipeline

jamie-suse:
# Description

This PR rebuilds the `main` branch dependencies if they are missing, in order to allow the `BC-check` step to run.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -180,6 +180,7 @@ jobs:
   main-branch-deps:
     name: Rebuild main branch dependencies
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):

dottorblaster (changes requested):
Great! Surely a step in the right direction :smile: 

Could you only add an `if` clause to the `main-branch-deps` job that skips this entirely if we are on `main`? I know we already do something like that to know the cache hit but I'd like to get rid of all of this when CI runs on `main`, we'd spare some seconds :sweat_smile:

dottorblaster (approved):
LGTM

PR #301, opened 2023-09-28
Add fstab gatherer docs

CDimonaco:
Add docs for agent fstab gatherer


guides/gatherers.md
```diff
@@ -238,6 +239,44 @@ Example output (in Rhai):
 #{ nodeid: 2, ring0_addr: "10.80.1.12" };
 ```
 
+### fstab
+
+**Argument required**: no.
+
+This gatherer allows access to the /etc/fstab file, returning all entries available at the file.
```
arbulu89, line 246: I know it is generic linux knowledge, but we could say that if  `backup` or `check_order` are not present in the fstab file, `0` will be returned

arbulu89 (approved):
Thank you!
Take care of the ordering once you merge the different docs, to keep the entries sorted

nelsonkopliku (approved):

PR #305, opened 2023-10-04
Add sap_profiles gatherer documentation

arbulu89:
`sap_profiles` gatherer documentation.
Based on: https://github.com/trento-project/agent/pull/267


rtorrero (approved):
Perfect, thanks @arbulu89 :+1:

nelsonkopliku (approved):
LGTM.

Not sure it's worth specifying that this gatherer not only has no _required_ argument, but that has no usecase with arguments.

arbulu89:
> Not sure it's worth specifying that this gatherer not only has no _required_ argument, but that has no usecase with arguments.

We already have a bunch of gatherers like that. I would say it is "obvious" with the examples.
Let's see if any user has issues with that :stuck_out_tongue_closed_eyes:

PR #308, opened 2023-10-06
Refactor checks from when to metadata

nelsonkopliku:
Refactors checks to use metadata instead of when condition.

Gotta test the whole flow with checks having metadata and wanda's implementation here https://github.com/trento-project/wanda/pull/309 before merging


dottorblaster (approved):
No blockers from me

jamie-suse (approved):
LGTM :rocket:

nelsonkopliku:
@abravosuse all these community checks are targeting clusters. What about the cluster type? Do they need specification in that regard?

abravosuse:
@nelsonkopliku as discussed in the planning meeting, lets set the cluster type as HANA scale-up. I know some of the checks are also valid for ASCS/ERS clusters but let's leave that for the Checks Team.

PR #309, opened 2023-10-09
Add metadata support

dottorblaster:
Just adding support to a metadata field to annotate checks and filter upon.


jamie-suse (approved):

nelsonkopliku (commented):
Some initial notes.

`GET /api/v1/checks/catalog?provider=azure`
This query is not returning checks without an explicit provider in the metadata.

ie
```yaml
metadata:
  target_type: host
```
However considering that if a check doesn't have provider set up, then it is valid for any provider, I believe it should be part of the response.

The same query to the catalog does not return a check having a list as the `provider` metadata value
```yaml
metadata:
  target_type: host
  provider: [azure, nutanix, kvm, vmware]
```

Seems to be only functioning with a direct string match
```yaml
metadata:
  provider: azure
```
Either I am missing something or we need to iterate over.

nelsonkopliku (commented):
`GET /api/v1/checks/catalog?provider=azure`
This query is not returning checks without an explicit provider in the metadata.

ie a check with the following metadata (host or cluster is irrelevant) is not returned
```yaml
metadata:
  target_type: host
```

It does however return correctly the following cases
```yaml
metadata:
  target_type: cluster
  provider: azure

metadata:
  target_type: host
  provider: [azure, nutanix, kvm, vmware]
```

dottorblaster:
@nelsonkopliku damn you're right :sweat_smile: I pushed a fix for this and added a test case :+1:

nelsonkopliku (commented):
Definitely a step forward, we should be able to support at least current cases.

However here's a couple of other notes:

Our checks schema definition specifies metadata can be also booleans or numbers, besides strings
```yaml
metadata:
  foo: some_value
  foo: true
  bar: 42
```
but unfortunately a query like the following 
`/api/v1/checks/catalog?foo=some_value&bar=true&baz=42` returns a validation error. and the problem is in `lib/wanda_web/schemas/env.ex:14` where we specify the properties of the env to be oneOf: [string, integer, boolean, a-nested-map]
```elixir
additionalProperties: %Schema{
  oneOf: [
    %Schema{type: :string},
    %Schema{type: :integer},
    %Schema{type: :boolean},
    %Schema{type: :array, items: __MODULE__}
  ]
}
```
and by changing that `oneOf` to an `anyOf` we could at least pass the query string validation.

But again, what comes in the query string is indeed a string, so, unluckily enough, in order to have a match the check would need to be rewritten like
```yaml
metadata:
  foo: some_value
  foo: "true"
  bar: "42"
```
unless we make sure to have typed query strings (ie. using some library that does so).

This applies for both the catalog and the start execution endpoints, which use the same `Env` schema. The execution seems to be working, but it is subject to the same typings concerns.

Bottom line, this implementation is kinda enough to start with but it has the highlighted shortcomings.
If we want to start by merging this I'd be ok that, however I do believe we need to iterate over again.

nelsonkopliku:
good point on the when condition, we should have the same issue there as well, never tried indeed something different than strings :sweat_smile: 

However I did a quick search on google for `elixir typed query string` and I got 
https://hexdocs.pm/elixir/1.12/URI.html
https://elixirforum.com/t/typed-url-parameters-in-phoenix/2633/3

hope they help :wink:

dottorblaster:
Fixed using the `CastAndValidate` plug, we just needed to match integers and booleans before :grimacing:

CDimonaco (approved):

nelsonkopliku (approved):
Just tried again with the latest changes.

Works with strings, numbers, and booleans. Super! Thanks!

PR #315, opened 2023-10-13
Update rabbitmq image in docker-compose

dottorblaster:
Just updating RabbitMQ inside docker-compose because it has better support for OTP 26 clients


nelsonkopliku (approved):
LGTM. 
What about aligning web and agent as well if needed?

dottorblaster:
@nelsonkopliku yeah I think the web needs that, the agent already has a generic `3`

PR #317, opened 2023-10-17
Pin all the gatherer version in checks to v1

CDimonaco:
Pin all the gatherers name to v1 version to match the new versioning feature in trento-agent

DEPENDS ON: https://github.com/trento-project/agent/pull/277


arbulu89 (approved):
great!

dottorblaster (approved):
LET'S DO THIS

PR #319, opened 2023-10-18
Add systemd@v2 gatherer documentation

arbulu89:
`systemd@v2` gatherer documentation.
Based on: https://github.com/trento-project/agent/pull/278

### DIFF
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -1101,11 +1101,11 @@ Example output (in Rhai):
 
 Gather systemd units information. It returns an object with multiple fields about the systemd unit.
 
-The provided service name must include the extension, such as `.service` or `.mount`.
+The provided unit name must include the extension, such as `.service` or `.mount`.
 
-Only a subset of properties are returned. The information about them is available in
+Only a subset of properties are returned. Additional information about these is available in
 the [systemd](https://www.man7.org/linux/man-pages/man5/org.freedesktop.systemd1.5.html#UNIT_OBJECTS) man pages, 
-with some additional details in the `Properties` sub-chapter.
+with some detailed description in the `Properties` sub-chapter.
 
 Example arguments:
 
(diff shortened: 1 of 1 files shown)


guides/gatherers.md
```diff
@@ -1094,6 +1095,62 @@ Example output (in Rhai):
 "inactive";
 ```
 
+### systemd@v2
+
+**Argument required**: yes.
+
+Gather systemd units information. It returns an object with multiple fields about the systemd unit.
+
+The provided service name must include the extension, such as `.service` or `.mount`.
```
rtorrero, line 1104: "The provided service name" -> "The provided unit name"

guides/gatherers.md
```diff
@@ -1094,6 +1095,62 @@ Example output (in Rhai):
...
+
+Only a subset of properties are returned. The information about them is available in
```
rtorrero, line 1106: "The information about them" -> "Additional information about these"

rtorrero (approved):
LGTM, two minor improvements proposed

rtorrero:
@arbulu89 after a bit of thought, might it make sense to maybe point out the differences between the v1 and v2? so someone can easily spot what changed?

arbulu89:
> I just noticed something wrong, the index points to `#systemdv2` but the section is called `systemd@v2` which prevents the linking from working AFAICS

Did you try @rtorrero ?
I did, and this is the unique way that the link works. It was not a typo at least, I wrote that deliberately

rtorrero:
> > I just noticed something wrong, the index points to `#systemdv2` but the section is called `systemd@v2` which prevents the linking from working AFAICS
> 
> Did you try @rtorrero ? I did, and this is the unique way that the link works. It was not a typo at least, I wrote that deliberately

I've tested with a few different editors and you are actually right, in GH flavored MD (most common I guess), the symbols are simply removed from the links

rtorrero (approved):

arbulu89:
@rtorrero (and others). Based on the markdown chapters linking conversation, I have tried with the `ex_doc` docs, and unfortunately the linking fails there...
How github and hex create the links for the chapters is different.
I didn't find any universal solution for all of them.
One think we could do, is ti name the chapter as `systemd - v2` instead of using the `@` which is the troublemaker

arbulu89:
Following up the last comment, the last change I did adding the `<span id="systemd-v2"></span>` makes the job.
This tag tricks the markdown page to move to the correct position in both github and hex docs.
Not a elegant solution, but we have both documentations at least working, and it is not really invasive.
Opinions?

CDimonaco:
For me the important thing is the documentation tbh, markdown allows html tags exactly for this edge cases, for me it's the most correct way @arbulu89

EMaksy (approved):
Hey the docs are looking good :D

CDimonaco (approved):

PR #322, opened 2023-10-19
Add versioning info to gatherers docs

rtorrero:
Add relevant versioning information to the docs as well as tag/pin all existing gatherers to @v1

### DIFF
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -1144,7 +1144,7 @@ Example output (in Rhai):
 "inactive";
 ```
 
-<span id="systemd-v2"></span>
+<span id="systemdv2"></span>
 
 ### systemd@v2
 
@@ -1201,6 +1201,8 @@ Example output (in Rhai):
 }
 
 ```
+
+<span id="verify_passwordv1"></span>
 
 ### verify_password@v1
 
--- a/guides/gatherers.md
+++ b/guides/gatherers.md
@@ -16,7 +16,8 @@ fact = gatherer(argument)
 
 ## Gatherers versioning
 
-When an update to the trento-agent includes a non-backwards compatible change in a gatherer (e.g., changes to the Rhai output format), its version is
+The gatherers implementation supports a versioning mechanism in order to enable non-backwards compatibility changes in any of them. When an update to
+the trento-agent includes a non-backwards compatible change in a gatherer (e.g., changes to the Rhai output format), its version is
 bumped by incrementing the @vN suffix that follows the gatherer's name, where 'N' represents the new version of that gatherer.
 Example:
 
@@ -26,9 +27,9 @@ Example:
 Note that when writing a check, if no tag is specified (e.g. `systemd`), the latest version is used. It is **strongly** recommended to always pin your
 checks to a specific version of a gatherer.
 
-Also note that not all changes in a released gatherer get a new version tag. A new version tag is released only for breaking changes,
-while non-breaking changes such as additional fields in the Rhai output reuse the latest existing tag. To use a check that relies on a newer
-field introduced after an update, upgrade the agent to the latest version to ensure that the required gatherers are also up-to-date.
+Not all changes in a released gatherer get a new version tag. A new version tag is released only for breaking changes, while non-breaking changes such
+as additional fields in the Rhai output reuse the latest existing tag. To use a check that relies on a newer field introduced after an update, upgrade
+the agent to the latest version to ensure that the required gatherers are also up-to-date.
 
 ## Available Gatherers
 
@@ -56,7 +57,7 @@ Here's a collection of built-in gatherers, with information about how to use the
 | [`systemd@v2`](#systemdv2)                   | [trento-project/agent/../gatherers/systemd_v2.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/systemd_v2.go)           |
 | [`verify_password@v1`](#verify_passwordv1)   | [trento-project/agent/../gatherers/verifypassword.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/verifypassword.go)   |
 
-<span id="cibadmin-v1"></span>
+<span id="cibadminv1"></span>
 
 ### cibadmin@v1
 
@@ -117,7 +118,7 @@ Example output (in Rhai):
 };
 ```
 
-<span id="corosyncconf-v1"></span>
+<span id="corosyncconfv1"></span>
 
 ### corosync.conf@v1
 
@@ -180,7 +181,7 @@ Example output (in Rhai):
 
 For extra information refer to [trento-project/agent/../gatherers/corosyncconf_test.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/corosyncconf_test.go)
 
-<span id="corosync-cmapctl-v1"></span>
+<span id="corosync-cmapctlv1"></span>
 
 ### corosync-cmapctl@v1
 
@@ -267,7 +268,7 @@ Example output (in Rhai):
 #{ nodeid: 2, ring0_addr: "10.80.1.12" };
 ```
 
-<span id="dir_scan-v1"></span>
+<span id="dir_scanv1"></span>
 
 ### dir_scan@v1
 
@@ -307,7 +308,7 @@ Example output (in Rhai):
   ]
 ```
 
-<span id="fstab-v1"></span>
+<span id="fstabv1"></span>
 
 ### fstab@v1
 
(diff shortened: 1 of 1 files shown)


guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
 fact = gatherer(argument)
 ```
 
+## A note of gatherers versioning
```
arbulu89, line 17: Maybe simply `Gatherers versioning`

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
...
+
+When a gatherer includes a breaking change (e.g. the Rhai output format changes), a new version of that same gatherer will be introduced with a new
```
arbulu89, line 19: `When a change on a gatherer includes...`
And avoid future term `... of that same gatherer is introduced with a new...`

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
...
+
+- `systemd@v1` -> Represents the first version of the systemd gatherer
+- `systemd@v2` -> Represents the second version of the systemd gatherer
+
+Note that when writing a check, if no tag is specified (e.g. `systemd`), the latest version will be used. Because of this it is **strongly** recommended
+to always pin your checks to a specific version of a gatherer.
+
+Also note that not all changes in a released gatherer will get a new version tag. A new version tag will be released only for breaking changes,
+while non-breaking changes such as additional fields in the Rhai output will reuse the latest existing tag. This means that if you want to use a check
+that relies on a newer field introduced after an update, you will need to upgrade both wanda and the agent to the latest version.
```
arbulu89, line 32: why do you need to update wanda to the latest version?

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
...
+| [`sbd_config@v1`](#sbd_configv1)             | [trento-project/agent/../gatherers/sbd.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/sbd.go)                         |
+| [`sbd_dump@v1`](#sbd_dumpv1)                 | [trento-project/agent/../gatherers/sbddump.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/sbddump.go)                 |
+| [`sysctl@v1`](#sysctlv1)                     | [trento-project/agent/../gatherers/sysctl.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/sysctl.go)                   |
+| [`systemd@v1`](#systemdv1)                   | [trento-project/agent/../gatherers/systemd.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/systemd.go)                 |
+| [`verify_password@v1`](#verify_passwordv1)   | [trento-project/agent/../gatherers/verifypassword.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/verifypassword.go)   |
+
+### cibadmin@v1
```
arbulu89, line 59: Unfortunately, this breaks the redirecting in `hex` docs.
We will need to add the `span` trick, as I did for systemd@v2.
Other option is to only do the change in the index, and leave the `v1` version title without the version suffix

arbulu89 (changes requested):
Hey @rtorrero 
2 things:
- Try to update the future term in the text by present. Avoid usage of `will`
- We need to fix the redirecting of the index for `hex`. You can try this out locally running `mix docs`. This created the docs locally for you

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
...
+
+When a gatherer includes a breaking change (e.g. the Rhai output format changes), a new version of that same gatherer will be introduced with a new
+version labeled as a suffix to the gatherers name.
+
+Example:
+
+- `systemd@v1` -> Represents the first version of the systemd gatherer
+- `systemd@v2` -> Represents the second version of the systemd gatherer
+
+Note that when writing a check, if no tag is specified (e.g. `systemd`), the latest version will be used. Because of this it is **strongly** recommended
```
EMaksy, line 27: What about skipping  because ? 
Before: `Because of this it is strongly recommended ...`
After: `It is strongly recommended`

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
...
+to always pin your checks to a specific version of a gatherer.
+
+Also note that not all changes in a released gatherer will get a new version tag. A new version tag will be released only for breaking changes,
```
EMaksy, line 30: Just an optional suggestion: 
Two times `note` like in line 27. 
`Not all changes in a released gatherer will get a new version tag. A new version tag will be released only for breaking changes`

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
...
+while non-breaking changes such as additional fields in the Rhai output will reuse the latest existing tag. This means that if you want to use a check
```
EMaksy, line 31: If i have it properly in memory we should try to avoid the "you" form and be more imperative so we stick more to the suse style guide docs. 
What do you think about something similar like this
`To use a check that relies on a newer field, which was introduced after an update, upgrade wanda and the agent to the latest version`

guides/gatherers.md
EMaksy, line 19: Maybe we could use present tense  like suggested in the [suse styleguide for documentation](https://documentation.suse.com/style/current/single-html/docu_styleguide/#sec-tense). 
So instead of `will be introduced`  use maybe `is introduced`
or what do you think ? :D

guides/gatherers.md
EMaksy, line 27: Maybe also here present tense?
Instead of `will be used`  use `is used`

guides/gatherers.md
rtorrero: What I meant is something like (using what eugen suggested):
To use a check that relies on a newer field introduced after an update, upgrade not only wanda but also the agent to the latest version

but after reading you, I think it creates more confusion like this... I'll remove that

rtorrero:
@arbulu89 @EMaksy thanks! take a look at the changes, if everything is OK I'll rebase and merge :+1:

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
...
-| [`sysctl`](#sysctl)                     | [trento-project/agent/../gatherers/sysctl.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/sysctl.go)                   |
-| [`systemd`](#systemd)                   | [trento-project/agent/../gatherers/systemd.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/systemd.go)                 |
-| [`verify_password`](#verify_password)   | [trento-project/agent/../gatherers/verifypassword.go](https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/verifypassword.go)   |
-
-### cibadmin
+| Name                                         | Implementation                                                                                                                                              |
+| :------------------------------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
arbulu89, line 38: This header line looks not aligned. Have a look, I think one space is missing

guides/gatherers.md
```diff
@@ -14,32 +14,49 @@ Facts Gathering process in a nutshell
 fact = gatherer(argument)
 ```
 
+## Gatherers versioning
+
+When an update to the trento-agent includes a non-backwards compatible change in a gatherer (e.g., changes to the Rhai output format), its version is
```
arbulu89, line 19: I think we should have a better 1st phrase. Something like:
```
The gatherers implementation supports a versioning mechanism in order to enable non-backwards compatibility changes in any of them.
When an update to the trento-agent includes a non-backwards...
```
At the end, we want to start describing a bit the chapter, not just how it is implemented

guides/gatherers.md
```diff
@@ -98,7 +115,8 @@ Example output (in Rhai):
 };
 ```
 
-### corosync.conf
+<span id="corosyncconf-v1"></span>
```
arbulu89, line 118: I would go with the `corosyncconfv1` id. This way, `vscode` editor continues working, and github pages/hex docs as well.
So just having the `v1` attached directly to the name
Could you try?

arbulu89 (approved):
Good from my side.
Please update the links id usage.
Once conflicts are fixed, we can merge

guides/gatherers.md
rtorrero: yep, works on both editor and docs now :+1:

PR #330, opened 2023-10-24
Fix redirection when a query string is involved

nelsonkopliku:
# Description

Fixes redirection when also a query string is involved in the initial request.
Needs to go after https://github.com/trento-project/wanda/pull/326
## How was this tested?

Automated test added.

### DIFF
--- a/lib/wanda_web/plugs/api_redirector.ex
+++ b/lib/wanda_web/plugs/api_redirector.ex
@@ -70,7 +70,7 @@ defmodule WandaWeb.Plugs.ApiRedirector do
   end
 
   defp redirect(conn, to) do
-    Controller.redirect(conn, to: to <> conn.query_string)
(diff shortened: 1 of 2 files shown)


lib/wanda_web/plugs/api_redirector.ex
```diff
@@ -64,4 +64,7 @@ defmodule WandaWeb.Plugs.ApiRedirector do
       _ -> true
     end
   end
+
+  defp maybe_add_query_string(path, ""), do: path
+  defp maybe_add_query_string(path, query_string), do: path <> "?#{query_string}"
```
dottorblaster, line 69:
```suggestion
  defp maybe_add_query_string(path, query_string), do: "#{path}?#{query_string}"
```

dottorblaster (changes requested):
Just a small thing but the content is absolutely legit :+1: 

Once rebased, we can merge

dottorblaster (approved):

PR #337, opened 2023-10-30
Port execution view to v2 and cover execution controller with tests

dottorblaster:
As above, porting execution view to v2 and covering everything left with tests

### DIFF
--- a/lib/wanda_web/views/v2/execution_view.ex
+++ b/lib/wanda_web/views/v2/execution_view.ex
@@ -1,0 +1,15 @@
+defmodule WandaWeb.V2.ExecutionView do
+  use WandaWeb, :view
+
+  def render("start.json", %{
+        accepted_execution: %{
+          execution_id: execution_id,
+          group_id: group_id
+        }
+      }) do
+    %{
+      execution_id: execution_id,
+      group_id: group_id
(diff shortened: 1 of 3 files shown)


jamie-suse (approved):
LGTM :rocket:

lib/wanda_web/views/v2/execution_view.ex
```diff
@@ -0,0 +1,15 @@
+defmodule WandaWeb.V2.ExecutionView do
```
nelsonkopliku, line 1: where and how does this view get used? I am probably not remembering the wiring with the controller :thinking: 

Shouldn't we have in our controllers something like
```elixir
render(conn, "start.json", page: %{
          execution_id: execution_id,
          group_id: group_id
        })
```

we are doing a plain
```elixir
conn
      |> put_status(:accepted)
      |> render(
        accepted_execution: %{
          execution_id: execution_id,
          group_id: group_id
        }
      )
```

dottorblaster: It's a Phoenix convention, if we don't declare this module the controller explodes :sweat_smile:

nelsonkopliku (approved):

PR #347, opened 2023-12-04
Enforce `target_type` in Check's metadata

jamie-suse:
This change adds the `target_type` as a required key in the `metadata` map.

`target_type` has type `string`.

Updated `specification.md` documentation to reflect this.

Also the indentation in `check_definition.schema.json` was out, so it was corrected.

### DIFF
--- a/guides/check_definition.schema.json
+++ b/guides/check_definition.schema.json
@@ -1,202 +1,202 @@
 {
-    "$schema": "http://json-schema.org/draft-07/schema#",
-    "title": "Trento Check Definition Schema",
-    "additionalProperties": false,
-    "properties": {
-      "id": {
-        "type": "string"
-      },
-      "severity": {
-        "type": "string"
-      },
-      "name": {
-        "type": "string"
-      },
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
LGTM.

We just need to be conscious of the language we let in wanda, if we want to keep it being agnostic.

jamie-suse:
Do you mean that we shouldn't explicitly refer to `host` and `cluster`?

Should we define `target_type` as simply a string then? That way we can just say that `target_type` needs to be a string.

jamie-suse:
Thanks for the explanation, I think considering your comment, it's better to change it to a string to maintain the agnosticism.

dottorblaster (approved):
Green light for me, we can think about moving the whole `target_type` checking inside tlint in a subsequent phase.

jamie-suse:
Though side note: isn't `target_type` still quite a Trento-related terminology? Simply including it takes away _some_ agnosticism :thinking:

PR #364, opened 2024-02-09
Rpm package

rtorrero:
# Description
This PR adds RPM package generation on OBS in addition to the existing container image building. Note that it most likely will fail to build in OBS but after we set up the linking on IBS, there it should build correctly.

It follows the same logic as the changes introduced in https://github.com/trento-project/web/pull/2258


.github/workflows/ci.yaml
```diff
@@ -463,13 +463,85 @@ jobs:
...
+          cargo vendor
+          printf '\n\n[source.crates-io]\nreplace-with = "vendored-sources"\n\n[source.vendored-sources]\ndirectory = "vendor"\n' >> .cargo/config
+      - name: Create deps tarball
+        run: tar -c deps | gzip -n > deps.tar.gz          
+      - name: Checkout and prepare OBS package
+        run: |
+          osc checkout $OBS_PROJECT trento-wanda -o $OSC_CHECKOUT_DIR
+          cp $FOLDER/_service $OSC_CHECKOUT_DIR
+          cp $FOLDER/trento-wanda.spec $OSC_CHECKOUT_DIR
+          rm -v $OSC_CHECKOUT_DIR/*.tar.gz
```
arbulu89, line 530: Maybe we should put a `-f` here, so we don't fail if there isn't any tarball

.github/workflows/ci.yaml
```diff
@@ -463,13 +463,85 @@ jobs:
...
+          VERSION=$(./hack/get_version_from_git.sh)
+          sed -i 's~%%REVISION%%~${{ github.sha }}~' $FOLDER/_service && \
+          sed -i 's~%%REPOSITORY%%~'"${REPOSITORY}"'~' $FOLDER/_service && \
+          sed -i 's~%%VERSION%%~'"${VERSION}"'~' $FOLDER/_service
+      - name: Setup rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.66
+      - name: Get mix deps
+        run: mix local.hex --force && mix deps.clean --all && mix deps.get
```
arbulu89, line 517: Should we install `rebar` as we do in the docker style?
Or is it not needed?

arbulu89 (commented):
Hey @rtorrero 
In general it looks great.
The unique thing that i see pending is the `premium-checks` installation.
In the dockerfile, we have an additional `premium-checks.tar.gz` file, which we should have here as well.
Most probably we should do something similar here.

.github/workflows/ci.yaml
rtorrero: I think it's only used in the build phase. We do install it as a BuildRequire in OBS (see the spec file) and it seems to be enough based on the fact that the resulting binary seems to work correctly

arbulu89 (approved):
Thank you

PR #368, opened 2024-02-22
Remove leftover start execution schema

arbulu89:
I guess we forgot to remove it at some point. Here most probably: https://github.com/trento-project/wanda/pull/309
We already have the V1 and V2 version, so no need to have the old one.


nelsonkopliku (approved):
:+1:

PR #370, opened 2024-02-26
Add unknown expect type in openapi schemas

arbulu89:
Add unknown expect type in openapi schemas.
This allows us to avoid "complex/strange" renderization of old views for these responses.
Otherwise, everytime we add a new expect time, we need to deal with those checks, removing them, etc, which is pretty annoying.

I know that this is breaking change, but it is pretty small, and honestly, I don't think anybody will notice it.


nelsonkopliku (approved):

dottorblaster (approved):
LGTM, also I'm pretty sure nobody will notice the change.

PR #371, opened 2024-02-26
Load expect_enum expectations from catalog

arbulu89:
Load expect_enum expectations from catalog.
The catalog schemas has been updated and a new `v3` version created.
In the `v1` and `v2` versions of the catalog, the `expect_enum` expectations are converted to `unknown`. 

PD: The non backward compatible for v3 in the CI must be added in a next PR, as it doesn't exist in main now, so it cannot be compared

PD2: Ton shit of new lines to simply add a new enum to a field :facepalm:


lib/wanda_web/schemas/v1/catalog/check.ex
```diff
@@ -80,7 +80,7 @@ defmodule WandaWeb.Schemas.CatalogResponse.Check do
         items: %Schema{
           type: :object,
           properties: %{
-            name: %Schema{type: :string, description: "Expecation name"},
+            name: %Schema{type: :string, description: "Expectation name"},
```
EMaksy, line 83: Nice catch :+1:

EMaksy (approved):
Hey man nice job, adding v3 :rocket: 
LGTM

PR #373, opened 2024-02-27
Update execution endpoint schemas to include `expect_enum` expectation

arbulu89:
More schema work...
Most of the new lines are copies from V1 to V2, but still, they need to be there...

- Move execution schemas to V1
- Create V2 schemas
- Update controllers

### DIFF
--- a/test/wanda_web/controllers/v2/execution_controller_test.exs
+++ b/test/wanda_web/controllers/v2/execution_controller_test.exs
@@ -11,6 +11,173 @@ defmodule WandaWeb.V2.ExecutionControllerTest do
 
   setup :verify_on_exit!
 
+  describe "list executions" do
+    test "should return a list of executions", %{conn: conn} do
+      insert_list(5, :execution)
+
+      json =
+        conn
+        |> get("/api/v1/checks/executions")
+        |> json_response(200)
+
+      api_spec = ApiSpec.spec()
+      assert_schema(json, "ListExecutionsResponse", api_spec)
+    end
+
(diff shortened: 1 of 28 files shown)


test/wanda_web/controllers/v2/execution_controller_test.exs
```diff
@@ -11,6 +11,173 @@ defmodule WandaWeb.V2.ExecutionControllerTest do
 
   setup :verify_on_exit!
 
+  describe "list executions" do
```
arbulu89, line 14: These tests are copied from the `v1` version controller tests. At the end, they need to test the same thing.
I don't have right now any option to have some sort of generic code to test the same in both modules.
Any ideas?

test/wanda_web/views/v2/execution_view_test.exs
```diff
@@ -1,9 +1,253 @@
...
+    CheckResult,
+    Execution,
+    ExpectationEvaluation,
+    ExpectationResult
+  }
+
   alias WandaWeb.V2.ExecutionView
 
   describe "ExecutionView" do
+    test "renders index.json" do
```
arbulu89, line 18: The same here. The tests are copied from v1

test/wanda_web/controllers/v2/execution_controller_test.exs
dottorblaster: Top of my mind, I think that having the same test would be fine, but I have no strong opinion about that :+1:

PR #383, opened 2024-03-18
Openapi no additional properties take two

EMaksy:
# Description

This PR aims to fix the properties of OpenApi Schemas for wanda.
This change involves mostly changes in the schemas.

Adding additionalProperties: false to schemas to ensure which properties are expected in request and response payloads to enforce stricter validation of input data.

Additionally, the view had to be adjusted to filter out the warning message from v1. 
Updated catalog factory and fixed existing tests

### DIFF
--- a/lib/wanda_web/schemas/v1/catalog/check.ex
+++ b/lib/wanda_web/schemas/v1/catalog/check.ex
@@ -11,12 +11,16 @@ defmodule WandaWeb.Schemas.V1.Catalog.Check do
     title: "Check",
     description: "A single check from the catalog",
     type: :object,
+    additionalProperties: false,
     properties: %{
       id: %Schema{type: :string, description: "Check ID"},
       name: %Schema{type: :string, description: "Check name"},
-      when: %Schema{
-        type: :string,
-        description: "Expression to determine whether a check should run",
+      group: %Schema{type: :string, description: "Check group"},
+      description: %Schema{type: :string, description: "Check description"},
+      remediation: %Schema{type: :string, description: "Check remediation"},
+      metadata: %Schema{
+        type: :object,
+        description: "Check metadata",
         nullable: true
       },
       severity: %Schema{
@@ -24,14 +28,11 @@ defmodule WandaWeb.Schemas.V1.Catalog.Check do
         description: "Check severity: critical|warning",
         enum: [:critical, :warning]
       },
-      premium: %Schema{
-        type: :boolean,
-        description: "Check is Premium or not"
-      },
       facts: %Schema{
         type: :array,
         items: %Schema{
           type: :object,
+          additionalProperties: false,
           properties: %{
             name: %Schema{type: :string, description: "Fact name"},
             gatherer: %Schema{type: :string, description: "Used gatherer"},
@@ -44,6 +45,7 @@ defmodule WandaWeb.Schemas.V1.Catalog.Check do
         type: :array,
         items: %Schema{
           type: :object,
+          additionalProperties: false,
           properties: %{
             name: %Schema{type: :string, description: "Value name"},
             default: %Schema{
@@ -54,6 +56,7 @@ defmodule WandaWeb.Schemas.V1.Catalog.Check do
               type: :array,
               items: %Schema{
                 type: :object,
+                additionalProperties: false,
                 properties: %{
                   value: %Schema{
                     oneOf: [
@@ -79,6 +82,7 @@ defmodule WandaWeb.Schemas.V1.Catalog.Check do
         type: :array,
         items: %Schema{
           type: :object,
+          additionalProperties: false,
           properties: %{
             name: %Schema{type: :string, description: "Expectation name"},
             type: %Schema{
@@ -96,10 +100,32 @@ defmodule WandaWeb.Schemas.V1.Catalog.Check do
               description: "Message returned when the check fails"
             }
           },
-          required: [:name, :type, :expression]
+          required: [:name, :type, :expression, :failure_message]
         }
+      },
+      when: %Schema{
+        type: :string,
+        description: "Expression to determine whether a check should run",
+        nullable: true
+      },
+      premium: %Schema{
+        type: :boolean,
+        description: "Check is Premium or not"
       }
     },
-    required: [:id, :name, :severity, :facts, :values, :expectations]
+    required: [
+      :id,
+      :name,
+      :group,
+      :description,
+      :remediation,
+      :metadata,
+      :severity,
+      :facts,
+      :values,
+      :expectations,
+      :when,
+      :premium
+    ]
   })
 end
--- a/lib/wanda_web/schemas/v3/catalog/check.ex
+++ b/lib/wanda_web/schemas/v3/catalog/check.ex
@@ -11,12 +11,16 @@ defmodule WandaWeb.Schemas.V3.Catalog.Check do
     title: "Check",
     description: "A single check from the catalog",
     type: :object,
+    additionalProperties: false,
     properties: %{
       id: %Schema{type: :string, description: "Check ID"},
       name: %Schema{type: :string, description: "Check name"},
-      when: %Schema{
-        type: :string,
-        description: "Expression to determine whether a check should run",
+      group: %Schema{type: :string, description: "Check group"},
+      description: %Schema{type: :string, description: "Check description"},
+      remediation: %Schema{type: :string, description: "Check remediation"},
+      metadata: %Schema{
+        type: :object,
+        description: "Check metadata",
         nullable: true
       },
       severity: %Schema{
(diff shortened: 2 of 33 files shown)


test/wanda_web/views/v1/catalog_view_test.exs
```diff
@@ -9,16 +9,43 @@ defmodule WandaWeb.V1.CatalogViewTest do
   describe "CatalogView" do
     test "renders catalog.json" do
       checks = [
-        build(:check, expectations: build_list(3, :catalog_expectation, type: :expect)),
-        build(:check, expectations: build_list(3, :catalog_expectation, type: :expect_same))
+        build(:check, expectations: build_list(2, :catalog_expectation, type: :expect)),
+        build(:check, expectations: build_list(2, :catalog_expectation, type: :expect_same))
       ]
 
+      updated_checks =
```
EMaksy, line 16: Reusing here the code to format the response of the view for v1,v2. 
Maybe there is a better solution for this.
Open for suggestions.

test/wanda_web/views/v1/catalog_view_test.exs
```diff
@@ -9,16 +9,43 @@ defmodule WandaWeb.V1.CatalogViewTest do
...
   end
 
+  test "renders catalog.json like expected in the v1 schema" do
+    checks = [
+      build(:check, expectations: build_list(2, :catalog_expectation, type: :expect)),
+      build(:check, expectations: build_list(2, :catalog_expectation, type: :expect_same))
+    ]
+
+    rendered_catalog = render(CatalogView, "catalog.json", catalog: checks)
+    refute Access.get(rendered_catalog, :warning_message)
```
EMaksy, line 46: Added this test as I expected more than one additional parameter that are not needed, but it was only one in the end. 
I think it does not, what do you think ?

test/support/factory.ex
```diff
@@ -23,10 +23,20 @@ defmodule Wanda.Factory do
...
       id: UUID.uuid4(),
       name: Faker.StarWars.character(),
+      group: Faker.Company.name(),
+      description: Faker.Lorem.sentence(),
+      remediation: Faker.Lorem.sentence(),
+      metadata: %{
+        target_type: Faker.StarWars.character(),
+        cluster_type: Faker.Lorem.sentence(),
+        provider:
+          Enum.take_random(["azure", "nutanix", "kvm", "vmware, gcp, aws"], Enum.random(0..6))
```
arbulu89, line 33: I think that you don't need this `Enum.random(0..6)`.
`Enum.random(["azure", "nutanix", "kvm", "vmware, gcp, aws"])` should be enough

test/support/factory.ex
```diff
@@ -23,10 +23,20 @@ defmodule Wanda.Factory do
     %Catalog.Check{
       id: UUID.uuid4(),
       name: Faker.StarWars.character(),
+      group: Faker.Company.name(),
+      description: Faker.Lorem.sentence(),
+      remediation: Faker.Lorem.sentence(),
+      metadata: %{
+        target_type: Faker.StarWars.character(),
```
arbulu89, line 30: If you want to be picky, you could get these values from arrays as well:
```target_type: Enum.random(["cluster", "host])` (check the values maybe in the web project.
The same for `cluster_type`

File: lib/wanda_web/schemas/v1/catalog/check.ex [hunk 5]
```diff
@@ -96,10 +100,32 @@ defmodule WandaWeb.Schemas.V1.Catalog.Check do
...
+      :name,
+      :group,
+      :description,
+      :remediation,
+      :metadata,
+      :severity,
+      :facts,
+      :values,
+      :expectations,
+      :when,
```
arbulu89, line 127: I don't know if we should put `when` here as it is nullable. Having `null` value passes the `required` test?
Anyway, not a big deal. Maybe it is good to have it in the required list

lib/wanda_web/schemas/v3/catalog/check.ex
```diff
@@ -102,10 +106,32 @@ defmodule WandaWeb.Schemas.V3.Catalog.Check do
                 "Message returned when the check return value is warning. Only available for `expect_enum` expectations"
             }
           },
-          required: [:name, :type, :expression]
+          required: [:name, :type, :expression, :failure_message, :warning_message]
```
arbulu89, line 109: Here maybe, we could remove the `warning_message` from the required list, and even the field from the view if the type is not `expect_enum`.
But this could be an improvement in future PRs as well, out of scope of this current work

lib/wanda_web/views/v1/catalog_view.ex
```diff
@@ -13,10 +13,19 @@ defmodule WandaWeb.V1.CatalogView do
 
   def adapt_v1(%{expectations: expectations} = check) do
     adapted_expectations =
-      Enum.map(expectations, fn
+      expectations
+      |> Enum.map(fn
         %{type: :expect_enum} = expectation -> Map.put(expectation, :type, :unknown)
         expectation -> expectation
       end)
+      |> Enum.map(fn %{
```
arbulu89, line 21: what about:
```
|> Enum.map(&(Map.drop(&1, [:warning_message]))
```
At the end, we want to remove that field, right? It makes it more obvious

test/support/factory.ex
```diff
@@ -23,10 +23,20 @@ defmodule Wanda.Factory do
...
+        cluster_type: Faker.Lorem.sentence(),
+        provider:
+          Enum.take_random(["azure", "nutanix", "kvm", "vmware, gcp, aws"], Enum.random(0..6))
+      },
       severity: Enum.random([:critical, :warning, :passing]),
       facts: build_list(10, :catalog_fact),
       values: build_list(10, :catalog_value),
       expectations: build_list(10, :catalog_expectation),
+      when: Faker.Lorem.sentence(),
```
arbulu89, line 39: Not for this line.
Shouldn't you add `failure_message` and `warning_message` in the `catalog_expectation_factory`?

test/wanda_web/views/v1/catalog_view_test.exs
arbulu89: Some comments:
- Move this test to the `adapt to V1 version` describe and name `warning_message is removed`
- You can just use one check `build_list(1, :check)`
And the assertion could be like:
```
[%{expectations: [expectation, _]}] = render(CatalogView, "catalog.json", catalog: checks)
refute Map.has_key?(expectation, :warning_message)
```

test/wanda_web/views/v1/catalog_view_test.exs
arbulu89:
```
adapted_checks =
  Enum.map(%{expectations: expectations} = check ->
    %{check | expectations: Enum.map(expectations, &(&Map.drop(&1, [:warning_message])})
```

arbulu89 (commented):
Looking good.
Some comments from my side

test/support/factory.ex
EMaksy: It is not required, my idea was to populate a random amount of provider to simulate the structure of a check.

Example out of checks: 
`  provider: [azure, nutanix, kvm, vmware]`
  
So we can also remove Enum.random(0..6) and make a fix number. No strong opinion here :+1:

lib/wanda_web/schemas/v1/catalog/check.ex
EMaksy: I just wanted to have everything what is represented in the struct, having nil seems not to harm.

test/support/factory.ex
EMaksy: Added!

test/support/factory.ex
EMaksy: Improved :D

lib/wanda_web/schemas/v3/catalog/check.ex
EMaksy: Let's skip this for now, as it goes a little bit out of scope. But I would track this down

CDimonaco (approved):

test/support/factory.ex
```diff
@@ -23,10 +23,20 @@ defmodule Wanda.Factory do
     %Catalog.Check{
       id: UUID.uuid4(),
       name: Faker.StarWars.character(),
+      group: Faker.Company.name(),
+      description: Faker.Lorem.sentence(),
+      remediation: Faker.Lorem.sentence(),
+      metadata: %{
+        target_type: Enum.random(["cluster", "host"]),
+        cluster_type: Enum.random(["ASCS_ERS", "HANA_SCALE_UP"]),
```
arbulu89, line 31: small case please

test/wanda_web/views/v2/catalog_view_test.exs
```diff
@@ -13,8 +13,23 @@ defmodule WandaWeb.V2.CatalogViewTest do
         build(:check, expectations: build_list(3, :catalog_expectation, type: :expect_same))
       ]
 
+      updated_checks =
```
arbulu89, line 16: I think we should have the same code here than in `v1`

arbulu89 (approved):
Thanks!
A couple of nitpics. Ready to merge

PR #432, opened 2024-05-22
Phoenix 1.7 upgrade

gagandeepb:
Upgrading Phoenix from 1.6 to 1.7.

Some comments:
- `phoenix_view` is now a separate package and needs to be added. 
- Elixir version in mix file is now in sync with the .tool-versions file.
- Fixing of one failling test that relies on internal implementation/non-public interface of `Phoenix.Router`:
(Old arg order in Phoenix 1.6: https://github.com/phoenixframework/phoenix/blob/9da43f0313482ea8eb62e7345cc9b8f18be087ef/lib/phoenix/router.ex#L433 
New arg order in Phoenix 1.7.12:
https://github.com/phoenixframework/phoenix/blob/bdc2d7398ac51514d964e02e302515d2ce95b6ff/lib/phoenix/router.ex#L584)
- The last point above hints at future opportunities for refactor of the `api_redirector` module from a plug to a controller (assuming there is some broad agreement) in order to not have to rely on reflection/non-public interfaces.


nelsonkopliku (approved):
Super! Thanks!

dottorblaster:
Great work @gagandeepb! If you want you can also have a stab at migrating from the old views to the new JSON rendering feature:

https://hexdocs.pm/phoenix/json_and_apis.html#rendering-json

Anyway, thank you so much!

gagandeepb:
@dottorblaster Thanks. I am not quite sure how the JSON rendering feature interacts with the `open_api_spex` schema format/serialization. I suspect/guess there might be some duplication of this Ser/De logic somewhere if both approaches were to co-exist, and this might need a deeper look/some internal conversations before proceeding.

PR #469, opened 2024-06-05
Updated metadata of cluster_type with hana_scale_out for existing checks

ksanjeet:
This pull request will update the metadata for cluster_type of existing checks with value for hana_scale_out


arbulu89 (approved):
Thank you @ksanjeet

PR #475, opened 2024-06-26
Adapted the existing checks to work with ascs_ers cluster

ksanjeet:
Adapted 3 community checks "6E9B82", "C620DC" and "D78671" to work with ascs_ers cluster.  The existing checks had constant values for corosync configurations "two_node" and "expected_vote" which in this adaptation is changed to use conditional checks and depends upon how many nodes are configured in corosync and cib.

### DIFF
--- a/priv/catalog/6E9B82.yaml
+++ b/priv/catalog/6E9B82.yaml
@@ -48,18 +48,36 @@ remediation: |
 
 metadata:
   target_type: cluster
-  cluster_type: hana_scale_up
+  cluster_type: 
+    - hana_scale_up
+    - ascs_ers
 
 facts:
   - name: corosync_twonode
     gatherer: corosync.conf@v1
     argument: quorum.two_node
 
+  - name: corosync_num_nodes
+    gatherer: corosync.conf@v1
+    argument: nodelist.node
+
+  - name: cib_num_nodes
+    gatherer: cibadmin@v1
+    argument: cib.configuration.nodes.node
+
 values:
   - name: expected_twonode
     default: 1
 
+  - name: expected_threeormore
+    default: 0
+
 expectations:
+  - name: num_nodes_equal_in_corosync_and_cib
+    expect: facts.corosync_num_nodes.len == facts.cib_num_nodes.len
+    failure_message: Number of nodes mentioned in corosync.conf is not equal to number of nodes in cib.xml
+
   - name: twonode_parameter
-    expect: facts.corosync_twonode == values.expected_twonode
-    failure_message: Corosync 'two_node' value was expected to be '${values.expected_twonode}' but configured value is '${facts.corosync_twonode}'
+    expect: |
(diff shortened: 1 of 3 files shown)


priv/catalog/6E9B82.yaml
```diff
@@ -48,18 +48,36 @@ remediation: |
...
 expectations:
+  - name: num_nodes_equal_in_corosync_and_cib
+    expect: facts.corosync_num_nodes.len == facts.cib_num_nodes.len
+    failure_message: Number of nodes mentioned in corosync.conf is not equal to number of nodes in cib.xml
+
   - name: twonode_parameter
-    expect: facts.corosync_twonode == values.expected_twonode
-    failure_message: Corosync 'two_node' value was expected to be '${values.expected_twonode}' but configured value is '${facts.corosync_twonode}'
+    expect: |
+     if facts.corosync_num_nodes.len == 2 { facts.corosync_twonode == values.expected_twonode } else if facts.corosync_num_nodes.len >= 3 { facts.corosync_twonode == values.expected_threeormore }
```
arbulu89, line 82: Hey @ksanjeet 
Here we have a blank sport if for some reason the num of nodes is 1.
Maybe we should add this condition and return `false`, in the final `else` statement.
I know it is a corner case, but i don't know how our engine would behave if we don't return a boolean. Most probably the check would fail with a unknown reason.

priv/catalog/D78671.yaml
```diff
@@ -49,18 +49,36 @@ remediation: |
...
 expectations:
+  - name: num_nodes_equal_in_corosync_and_cib
+    expect: facts.corosync_num_nodes.len == facts.cib_num_nodes.len
+    failure_message: Number of nodes mentioned in corosync.conf and cib.xml are not equal
+  
   - name: expectations_two_node
-    expect: facts.runtime_two_node == values.expected_runtime_two_node
-    failure_message: Corosync 'two_node' value was expected to be '${values.expected_runtime_two_node}' but value of running config is '${facts.runtime_two_node}'
+    expect: |
+     if facts.corosync_num_nodes.len == 2 { facts.runtime_two_node == values.expected_runtime_two_node } else if facts.corosync_num_nodes.len >= 3 { facts.runtime_two_node == values.expected_runtime_threeormore }
```
arbulu89, line 83: I would add a final `else` here as well.
By the way, the code would looks easier to read if you add some new lines. To look like:
```
if facts.corosync_num_nodes.len == 2 {
  facts.runtime_two_node == values.expected_runtime_two_node
} else if facts.corosync_num_nodes.len >= 3 {
  facts.runtime_two_node == values.expected_runtime_threeormore
} else {
  false
}
```

arbulu89 (approved):
Really good job @ksanjeet 
Just a couple of comments. 
- Add a `else` statement returning fale
- Add new lines, it is easier to read
I will merge once you apply those 2 things

arbulu89:
Thank you @ksanjeet 
Ready to merge, right?

ksanjeet:
Yes, please.

PR #478, opened 2024-07-05
Set minimum required rust version to 1.66

stefanotorresi:
We realized SLES15-SP3 has 1.65 which makes the build fail.


dottorblaster (approved):
LGTM

PR #480, opened 2024-07-08
Fix deps in trento-wanda.spec

stefanotorresi:
- use `rust+cargo` as a package name to work around some dependency resolution issues
- use one line per requirement as is common convention in RPM packaging


dottorblaster (approved):
LGTM

PR #513, opened 2024-09-27
Migrate Wanda to Phoenix 1.7 and Remove Phoenix.View

EMaksy:
# Description

This PR has the main goal to update Wanda to Phoenix 1.7 standards 

As Wanda is an api only app, the change is mainly focused on getting rid of the now outdated Phoenix.View as the view is/will be replaced by the Phoenix.Component see[ here](https://hexdocs.pm/phoenix_view/Phoenix.View.html#module-replaced-by-phoenix-component). I left everything html related out of the migration as to this point we don't use it, and with it we don't need the Phoenix.Component itself. 
It is expected to separate json and html in Phoenix 1.7 so a small restructure is required.
[Base example:](https://elixirforum.com/t/how-to-convert-this-simple-view-to-a-phoenix-1-7-view/54266)
```
# Before Phoenix 1.7
defmodule MyAppWeb.SomeView do
  use MyAppWeb, :view

  def render("template_a.html", assigns) do
    …
  end
  def render("template_b.json", assigns) do
    …
  end
end

# Phoenix 1.7 and above

defmodule MyAppWeb.SomeHTML do
  use MyAppWeb, :html

  def template_a(assigns) do
    …
  end
end

defmodule MyAppWeb.SomeJSON do
  def template_b(assigns) do
    …
  end
end
```

## What has been done ?

- Updated configs to the new phoenix app style
- Removed Phoenix view from the Project
- Every view was deleted --> Added JSON render only functions. 
- Adjusted Json rendering[ like in docs](https://hexdocs.pm/phoenix/json_and_apis.html#rendering-json)
- Updated tests

## How was this tested ? 
- Run existing tests 
- Manual tested


lib/wanda_web/controllers/fallback_controller.ex
```diff
@@ -1,26 +1,26 @@
...
   use Phoenix.Controller
 
-  alias WandaWeb.ErrorView
+  alias WandaWeb.ErrorJSON
 
   def call(conn, {:error, :no_checks_selected}) do
     conn
     |> put_status(:unprocessable_entity)
-    |> put_view(ErrorView)
+    |> put_view(ErrorJSON)
```
arbulu89, line 9: Shouldn't this be `put_view(json: ErrorJSON)` ?

lib/wanda_web/plugs/api_redirector.ex
```diff
@@ -45,7 +45,7 @@ defmodule WandaWeb.Plugs.ApiRedirector do
       nil ->
         conn
         |> put_resp_content_type("application/json")
-        |> resp(:not_found, Jason.encode!(ErrorView.render("404.json", %{detail: "Not found"})))
+        |> resp(:not_found, Jason.encode!(ErrorJSON.render("404.json", %{detail: "Not found"})))
```
arbulu89, line 48: Most probably something like this could more appropriate:
```
...
|> put_resp_content_type("application/json")
|> put_status(:not_found)
|> put_view(json: ErrorJSON)
|> render(:"404")
|> halt()
|> render(

```

What do you think? To use the same convention we use in everywhere else

dottorblaster (approved):
This looks good to me, I just want @CDimonaco that is our living Phoenix manual to give it a look and then we're good to merge imho

lib/wanda_web/views/health_json.ex
```diff
@@ -0,0 +1,3 @@
+defmodule WandaWeb.HealthJSON do
+  def health(%{health: health}), do: health
```
arbulu89, line 2: Wit this change, don't you need to change `health_controller.ex` file, to change `health.json` simply by `:health` ?
Something like `render(:health, health: %{database: db_status})`

lib/wanda_web/views/v1/execution_json.ex
```diff
@@ -1,26 +1,25 @@
...
-  def render("execution.json", %{execution: execution}) do
+  def execution(%Execution{} = execution) do
     execution
-    |> render_one(WandaWeb.V2.ExecutionView, "execution.json")
+    |> WandaWeb.V2.ExecutionJSON.execution()
     |> adapt_v1()
   end
 
-  def render("start.json", %{
+  def show(%{execution: execution}) do
```
arbulu89, line 18: I would put this `show` after `index` (actually, you don't need the `ExecutionJSON` prefix here), and make the `def execute` a private function, as it is only used here. I guess you cannot make them private `v2` on, as they are used here

lib/wanda_web/views/v1/execution_json.ex
```diff
@@ -1,26 +1,25 @@
-defmodule WandaWeb.V1.ExecutionView do
-  use WandaWeb, :view
+defmodule WandaWeb.V1.ExecutionJSON do
+  alias Wanda.Executions.Execution
+  alias WandaWeb.V1.ExecutionJSON
```
arbulu89, line 3: I don't think you need to alias yourself, just use the function name

lib/wanda_web/views/v2/catalog_json.ex
```diff
@@ -0,0 +1,7 @@
+defmodule WandaWeb.V2.CatalogJSON do
+  alias WandaWeb.V1.CatalogJSON
+
+  def catalog(%{catalog: catalog}) do
+    %{items: Enum.map(catalog, fn check -> CatalogJSON.check(check) end)}
```
arbulu89, line 5: Maybe `Enum.map(catalog, &CatalogJSON.check/1)`

lib/wanda_web/views/v2/execution_json.ex
```diff
@@ -1,31 +1,27 @@
-defmodule WandaWeb.V2.ExecutionView do
-  use WandaWeb, :view
-
-  alias WandaWeb.V2.ExecutionView
+defmodule WandaWeb.V2.ExecutionJSON do
+  alias WandaWeb.V2.ExecutionJSON
```
arbulu89, line 2: You don't need to alias yourself

arbulu89 (changes requested):
Some few comments and suggestions.

Question:
Besides that, I cannot really understand how the rendering in the controller works, if you are not defining which rendering function to use.
In some of them, I see that it is defined, but in others it is not.
Does phoenix use the same rendering function as the controller function name somehow? Example: in the catalog controller, the `catalog` function, you don't say what rendering to use, so does it go directly to the `catalog` rendering function because they have the same name?

arbulu89:
> Question:
> Besides that, I cannot really understand how the rendering in the controller works, if you are not defining which rendering function to use.
> In some of them, I see that it is defined, but in others it is not.
> Does phoenix use the same rendering function as the controller function name somehow? Example: in the catalog controller, the `catalog` function, you don't say what rendering to use, so does it go directly to the `catalog` rendering function because they have the same name?

Answering myself. Yes, if `render/2` is used, and 2nd param is the assigns map, it will use the current controller function name to decide which template to use

lib/wanda_web/views/health_json.ex
EMaksy: Oh god yes!

arbulu89 (approved):
I would still apply my other suggestions, but everything else looks great!

lib/wanda_web/plugs/api_redirector.ex
EMaksy: We don't use it in web in a similar function but i see what you mean, it just throws error, going to have a look.
For me both is good :)

CDimonaco (approved):

lib/wanda_web/views/v1/execution_json.ex
```diff
@@ -70,6 +63,12 @@ defmodule WandaWeb.V1.ExecutionView do
 
   defp adapt_v1_agent_check_results(agent_check_results), do: agent_check_results
 
+  defp execution(%Execution{} = execution) do
+    execution
+    |> ExecutionJSON.execution()
```
arbulu89, line 68: I think it is better to alias `V2` and have here `V2.ExecutionJSON.execution()` it is much more obvious what's going on

dottorblaster: Fair point :+1:

PR #515, opened 2024-09-30
Fix swaggerui action

dottorblaster:
Just fixing the github token to be provided to the swaggerui action


gagandeepb (approved):
LGTM. Thanks for this!

PR #522, opened 2024-10-11
Add premium deprecation warning

janvhs:
This commit partly reverts 51765fb, but adds a deprecation annotation. This enables check authors to spot the deprecation using other tooling than tlint.  Furthermore, it allows us to add and remove deprecations without having to touch tlint every time.  Yes, this is part of the specification :D

Also https://github.com/trento-project/tlint/pull/24 needs it :P

### DIFF
--- a/guides/check_definition.schema.json
+++ b/guides/check_definition.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
   "title": "Trento Check Definition Schema",
   "additionalProperties": false,
   "properties": {


EMaksy (approved):
LGTM :)

gagandeepb (commented):
Are you sure that draft7 dialect supports the deprecated annotation? I thought it was the draft2019 dialect.

gagandeepb (changes requested):
Can you please link here the relevant bit from the draft7 doc that specifies support for the `deprecation` key word?

gagandeepb (commented):
Would you like to add some [CLI/ CI based tests](https://github.com/python-jsonschema/check-jsonschema) somewhere to verify the intended behaviour ?

dottorblaster:
@gagandeepb thanks for your feedback!

I see two key parts in it:

- Linking to the right spec :heavy_check_mark: 
- Having a set of linting tools for the openAPI spec: I would like to discuss this for the future and defer at the moment.

dottorblaster (approved):

gagandeepb (approved):

PR #547, opened 2025-01-14
Save operations in the database

arbulu89:
# Description

Save the operation progress in the database during the orchestration. It saves the initial, intermediate and final states.
This way, the operation are permanently stored and we can expose them in the http API.
Besides that, they can be used to recover ongoing operations if the APP crashes and we need to resume ongoing operations.

PD: I have taken the advantage to refactor the tests. Now, the tests don't relay that much in implementation details. So attachment is needed, to improve the timing of the database queries (instead of relaying in static sleeps).

## How was this tested?
Tested with new UT

### DIFF
--- a/lib/wanda/operations.ex
+++ b/lib/wanda/operations.ex
@@ -3,4 +3,71 @@ defmodule Wanda.Operations do
   Operations are combined actions dispatched to different agents in order to apply
   persistent changes on them.
   """
+
+  alias Wanda.Repo
+
+  alias Wanda.Operations.{
+    Operation,
+    OperationTarget,
+    StepReport
+  }
+
+  @doc """
+  Create a new operarion.
+
+  If the operation already exists, it will be returned.
+  """
+  @spec create_operation!(String.t(), String.t(), [OperationTarget.t()]) :: Operation.t()
+  def create_operation!(operation_id, group_id, targets) do
+    %Operation{}
+    |> Operation.changeset(%{
+      operation_id: operation_id,
+      group_id: group_id,
+      status: :running,
+      result: :not_executed,
+      targets: Enum.map(targets, &Map.from_struct/1)
+    })
+    |> Repo.insert!(on_conflict: :nothing)
+  end
+
+  @doc """
+  Get an operation by operation_id.
+  """
+  @spec get_operation!(String.t()) :: Operation.t()
+  def get_operation!(operation_id) do
+    Repo.get!(Operation, operation_id)
+  end
+
+  @doc """
+  Update agent_reports of an operation
+  """
+  @spec update_agent_reports!(String.t(), [StepReport.t()]) ::
+          Operation.t()
+  def update_agent_reports!(operation_id, agent_reports) do
+    Operation
+    |> Repo.get!(operation_id)
+    |> Operation.changeset(%{
+      agent_reports: agent_reports
+    })
+    |> Repo.update!()
+  end
+
+  @doc """
+  Marks a previously started operation as completed
+  """
+  @spec complete_operation!(
+          String.t(),
+          :updated | :not_updated | :failed | :rolled_back | :skipped
+        ) ::
+          Operation.t()
+  def complete_operation!(operation_id, result) do
+    Operation
+    |> Repo.get!(operation_id)
+    |> Operation.changeset(%{
+      result: result,
+      status: :completed,
+      completed_at: DateTime.utc_now()
+    })
+    |> Repo.update!()
+  end
 end
--- a/lib/wanda/operations.ex
+++ b/lib/wanda/operations.ex
@@ -12,6 +12,9 @@ defmodule Wanda.Operations do
     StepReport
   }
 
+  require Wanda.Operations.Enums.Result, as: Result
+  require Wanda.Operations.Enums.Status, as: Status
+
   @doc """
   Create a new operarion.
 
@@ -23,8 +26,8 @@ defmodule Wanda.Operations do
     |> Operation.changeset(%{
       operation_id: operation_id,
       group_id: group_id,
-      status: :running,
-      result: :not_executed,
+      status: Status.running(),
+      result: Result.not_executed(),
       targets: Enum.map(targets, &Map.from_struct/1)
     })
     |> Repo.insert!(on_conflict: :nothing)
@@ -55,17 +58,13 @@ defmodule Wanda.Operations do
   @doc """
   Marks a previously started operation as completed
   """
-  @spec complete_operation!(
-          String.t(),
-          :updated | :not_updated | :failed | :rolled_back | :skipped
-        ) ::
-          Operation.t()
+  @spec complete_operation!(String.t(), Result.t()) :: Operation.t()
   def complete_operation!(operation_id, result) do
     Operation
     |> Repo.get!(operation_id)
     |> Operation.changeset(%{
       result: result,
-      status: :completed,
+      status: Status.completed(),
       completed_at: DateTime.utc_now()
     })
     |> Repo.update!()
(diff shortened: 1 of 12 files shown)


lib/wanda/operations/operation.ex
```diff
@@ -0,0 +1,49 @@
...
+    embeds_many :targets, Target, primary_key: false do
+      @derive Jason.Encoder
+
+      field :agent_id, Ecto.UUID, primary_key: true
+      field :arguments, :map
+    end
+
+    field :agent_reports, {:array, :map}
+
+    timestamps(type: :utc_datetime_usec, inserted_at: :started_at, updated_at: false)
```
CDimonaco, line 35: Why we want not to keep updated_at?

CDimonaco: Also could we have timestamps as the last field in the struct? I know order does not matter but ocd kicks in 👁️

lib/wanda/operations/operation.ex
```diff
@@ -0,0 +1,49 @@
...
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @fields ~w(operation_id group_id result status agent_reports started_at completed_at)a
+  @target_fields ~w(agent_id arguments)a
+
+  @derive {Jason.Encoder, [except: [:__meta__]]}
+  @primary_key false
+  schema "operations" do
```
CDimonaco, line 17: Could we add some required fields? Just to have a better validation

lib/wanda/operations/server.ex
```diff
@@ -93,7 +102,10 @@ defmodule Wanda.Operations.Server do
 
     # Evaluate results using agent_reports
     # Start rollback?
-    # Publish and store results
+    # Publish results
+
+    # Result is failed or rolledback, depending on the evaluation result
+    Operations.complete_operation!(operation_id, :failed)
```
CDimonaco, line 108: And use the enum here

lib/wanda/operations/server.ex
```diff
@@ -140,6 +154,9 @@ defmodule Wanda.Operations.Server do
     # Evaluate results using agent_reports
     # Publish and store results
 
+    # Result based on evaluation result
+    Operations.complete_operation!(operation_id, :updated)
```
CDimonaco, line 158: Enum here

priv/repo/migrations/20250109154813_add_operation.exs
```diff
@@ -0,0 +1,20 @@
+defmodule Wanda.Repo.Migrations.AddOperation do
+  use Ecto.Migration
+
+  def change do
+    create table(:operations, primary_key: false) do
+      add :operation_id, :uuid, primary_key: true
+      add :group_id, :uuid, null: false
+      add :result, :string, null: false
+      add :status, :string, null: false
+      add :targets, :map, null: false, default: "[]"
```
CDimonaco, line 10: Please could we use `:jsonb`? This is the correct migration type, the way we are using map works because postgres serialize everything under json and it "works" but the correct postgres type is the  jsonb one

This also for the agent_reports type
(I know we are using the same approach in the current migration for checks execution, I think in the future we should take care also of that part of migration)

CDimonaco: Also we can add a proper index to jsonb fields to speed up the queries (to both fields)

CDimonaco (commented):
Good job, I left some comments on the db side

lib/wanda/operations/operation.ex
arbulu89: > Why we want not to keep updated_at?

We don't have it in the executions. I just copied, but well, in this case, we update the entry more often, so maybe it makes to keep it.
I will change this

priv/repo/migrations/20250109154813_add_operation.exs
arbulu89: Updated

arbulu89:
Hey @CDimonaco ,
All updated!
I have not added any new index in the database migration as I don't know what we should aim to "improve".
Any idea? Otherwise, we can keep as it is and add them if needed

lib/wanda/operations.ex
```diff
@@ -3,4 +3,70 @@ defmodule Wanda.Operations do
...
+  @spec update_agent_reports!(String.t(), [StepReport.t()]) ::
+          Operation.t()
+  def update_agent_reports!(operation_id, agent_reports) do
+    Operation
+    |> Repo.get!(operation_id)
+    |> Operation.changeset(%{
+      agent_reports: agent_reports
+    })
+    |> Repo.update!()
+  end
```
gagandeepb, line 56: I wonder if there's any value you see in doing this in a single db transaction? For example multiple/concurrent callers of this function can make this two transaction get and update difficult to reason with at runtime, in my limited understanding.

CDimonaco:
@arbulu89 Wonderful job, for the indexing part I think we can create a new migration with two gin indexes (to not mess up with the automatic rollback of the one we already have) and have something like that

```elixir
def up do
  execute("CREATE INDEX operation_targets ON operations USING GIN(targets)")
    execute("CREATE INDEX operation_agent_reports ON operations USING GIN(agent_reports)")
end

def down do
  execute("DROP INDEX operation_targets")
    execute("DROP INDEX operation_agent_reports")
end
```

CDimonaco (approved):
I just left the comment about the indexing, having it now can save us some time and headache for the feature

Anyway LGTM 🚀

lib/wanda/operations.ex
arbulu89: I understand you concern. In the past, when we implemented the executions code we had the same dilemma: use transaction or not.
We decided that it was not needed, as at the end, we update the operations only in the operations gen server, and there all the actions are queued in the mailbox, so having concurrent calls should be impossible.
Besides that, as we are simply changing one field, the last change should remain always, so it should be harmless.

Either way, I understand that if we start using these functions carelessly anywhere and combining the `update_agent_reports` and `complete_operation`, we might have some potential danger.
But I doubt this will happen honestly.

arbulu89:
@CDimonaco Thank you!
copy/pasted the indexes.
you are the boss!

priv/repo/migrations/20250116095113_add_operation_jsonb_indexes.exs
```diff
@@ -0,0 +1,13 @@
...
+
+  def up do
+    execute("CREATE INDEX operation_targets ON operations USING GIN(targets)")
+    execute("CREATE INDEX operation_agent_reports ON operations USING GIN(agent_reports)")
+  end
+
+  def down do
+    execute("DROP INDEX operation_targets")
+    execute("DROP INDEX operation_agent_reports")
+  end
```
gagandeepb, line 12: Questions: 
1. I wonder why is a GIN index needed here ? I ask since I also wonder if the GIN index operator class is of some consequence here. 
2. Secondly, any specific preference for raw SQL vs Ecto here in the migrations?

arbulu89: >     2. Secondly, any specific preference for raw SQL vs Ecto here in the migrations?

How this would like like with Ecto ?
```
def change 
  create index(:operations, [:targets], using: "GIN")
  create index(:operations, [:agent_reports], using: "GIN")
end
```

CDimonaco: After a chat with @gagandeepb I think we can switch to jsonb_path_ops as gin index options, could we make the change? @arbulu89

gagandeepb: Here's [an example usage from elsewhere](https://github.com/trento-project/web/blob/main/priv/repo/migrations/20241009160142_add_activity_log_metadata_index.exs) of what I was suggesting for consideration.

arbulu89: @CDimonaco @gagandeepb changed in the last commit

CDimonaco:
merge it! 🚀

PR #548, opened 2025-01-15
Create and use enum modules

arbulu89:
# Description

Create and use enum modules.
We already have this piece of code in `web`, which is pretty nice.
This simply uses this module, without adding any new change.


balanza (approved):
LGTM 

image

CDimonaco (approved):

PR #549, opened 2025-01-16
Operations registry

arbulu89:
# Description

Create the operations registry to give functionalities to load already existing operations.
Finally, I have decided to put 1 operation per file.
I don't know how this can scale, but it looks simpler to find things.
Besides, I have used meaningful names as `id`, as strange hex values don't say anything, and as this is an internal implementation detail, it looks better.

## How was this tested?
UT


nelsonkopliku (approved):
Catalogs everywhere 😄

CDimonaco (approved):

PR #550, opened 2025-01-20
Add customizable key to disable checks customization for certain checks

EMaksy:
# Description

This PR is one of the first steps to enable users to customize checks. 
It adds a new key "customization" for checks. By default all checks with values are customization. 
In order to disable this behavior it needs to be explicitly set to false . 
This is needed to disable checks  like [3A59DC](https://github.com/trento-project/checks/blob/93beb612654381f06832ea43bf907556686625ed/checks/3A59DC.yaml) as changing it's value may cause issues.

## To Do:

After merging this we need to update [tlint repo  reference](https://github.com/trento-project/tlint/commit/32be1c8e4b672a241d716a5e24eec727f2012f1b)  to the new reference.

### DIFF
--- a/guides/check_definition.schema.json
+++ b/guides/check_definition.schema.json
@@ -91,6 +91,9 @@
         "name": {
           "type": "string"
         },
+        "customizable": {
+          "type": "boolean"
+        },
         "default": {
           "anyOf": [
             {
--- a/guides/check_definition.schema.json
+++ b/guides/check_definition.schema.json
@@ -37,6 +37,9 @@
         "$ref": "#/definitions/Fact"
       }
     },
+    "customizable": {
+      "type": "string"
+    },
     "values": {
       "type": "array",
       "items": {
--- a/guides/check_definition.schema.json
+++ b/guides/check_definition.schema.json
@@ -38,7 +38,7 @@
       }
     },
     "customizable": {
-      "type": "string"
+      "type": "boolean"
     },
     "values": {
       "type": "array",
--- a/guides/specification.md
+++ b/guides/specification.md
@@ -353,15 +353,11 @@ Finally, gathered facts, are used in Check's [Expectations](#expectations) to de
 
 ## Customizable
 
-Wanda's built-in checks, which include expected values, are **customizable** by **default**. Users can modify these check values through the [Trento Web console](https://github.com/trento-project/web) to adapt to specific system and environmental configurations.
+Users can modify a check's [expected values](#values) to adapt to specific system and environmental configurations.
+
+Built-in checks are considered **customizable** by **default**, so the `customizable` flag disables customization for a particular check.
 
 To explicitly mark a check as non-customizable, set the `customizable`  key to `false`:
-
-```yaml
-customizable: false
-```
-
-To explicitly mark a check as customizable, set the `customizable` key to `true`:
 
 ```yaml
 customizable: false
--- a/guides/specification.md
+++ b/guides/specification.md
@@ -353,9 +353,15 @@ Finally, gathered facts, are used in Check's [Expectations](#expectations) to de
 
 ## Customizable
 
-Wanda's built-in checks, which include expected values,are **customizable** by **default**. This means you can modify their values in [Trento Web console](https://github.com/trento-project/web).
+Wanda's built-in checks, which include expected values, are **customizable** by **default**. Users can modify these check values through the [Trento Web console](https://github.com/trento-project/web) to adapt to specific system and environmental configurations.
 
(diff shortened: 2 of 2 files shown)


guides/check_definition.schema.json
```diff
@@ -37,6 +37,9 @@
         "$ref": "#/definitions/Fact"
       }
     },
+    "customizable": {
+      "type": "string"
```
nelsonkopliku, line 41:
```suggestion
      "type": "boolean"
```

guides/specification.md
```diff
@@ -348,6 +351,22 @@ facts:
 
 Finally, gathered facts, are used in Check's [Expectations](#expectations) to determine whether expected conditions are met for the best practice to be adhered.
 
+## Customizable
+
+Wanda's built-in checks, which include expected values, are **customizable** by **default**. Users can modify these check values through the [Trento Web console](https://github.com/trento-project/web) to adapt to specific system and environmental configurations.
```
nelsonkopliku, line 356:
```suggestion
Users can modify a check's [expected values](#values) to adapt to specific system and environmental configurations.

Built-in checks are considered **customizable** by **default**, so this flag is a way to opt-out from customization for a particular check.
```

guides/specification.md
```diff
@@ -348,6 +351,22 @@ facts:
...
+To explicitly mark a check as non-customizable, set the `customizable`  key to `false`:
+
+```yaml
+customizable: false
+```
+
+To explicitly mark a check as customizable, set the `customizable` key to `true`:
+
+```yaml
+customizable: false
```
nelsonkopliku, line 367:
```suggestion
```

Besides it being `true` in this case 😄 , I believe it might turn out a bit redundant. What about omitting it?

nelsonkopliku (commented):
Thanks Eugen, just a couple of comments.

EMaksy:
@arbulu89 just reworked that part :+1:  and added additional docs

guides/specification.md
```diff
@@ -348,6 +352,21 @@ facts:
 
 Finally, gathered facts, are used in Check's [Expectations](#expectations) to determine whether expected conditions are met for the best practice to be adhered.
 
+## Customizable
+
+Users can modify a check's [expected values](#values) to adapt to specific system and environmental configurations.
+
+Built-in checks are considered **customizable** by **default**, so the `customizable` flag disables customization for a particular check.
+
+The customizability flag can be set global in the check and|or in [values](#customizable-values).
```
nelsonkopliku, line 361:
```suggestion
The customizability flag can be set globally for a check and|or for [specific values](#customizable-values). 
```

guides/specification.md
```diff
@@ -348,6 +352,21 @@ facts:
...
 Finally, gathered facts, are used in Check's [Expectations](#expectations) to determine whether expected conditions are met for the best practice to be adhered.
 
+## Customizable
+
+Users can modify a check's [expected values](#values) to adapt to specific system and environmental configurations.
+
+Built-in checks are considered **customizable** by **default**, so the `customizable` flag disables customization for a particular check.
+
+The customizability flag can be set global in the check and|or in [values](#customizable-values). 
+When both levels are set, the **global** flag takes **precedence**, overriding any value-level customizability.
```
nelsonkopliku, line 362:
```suggestion
When customization is globally disabled for a check, that is it has been marked with `customizable: false`, it overrides any values specifics, otherwise specific values customizability is considered.
```

nelsonkopliku (commented):
Noice! Just a couple of adjustments.

arbulu89:
> @arbulu89 just reworked that part 👍 and added additional docs

But do we need both versions (general and value specific)?
Because if the specific one overrides the other, I don't know if we really need the general one

guides/specification.md
EMaksy: I have splitted it up into two sentences. I hope this is a bit easier for the reader :)

EMaksy:
@arbulu89 
The main reason for having the customizable field at both the top level and the value level is to provide control over whether a value is **scalar** or not. This allows a check to has values that are not permitted to be modified.
@nelsonkopliku  please feel free to add or correct me  :+1:

arbulu89:
> @arbulu89 The main reason for having the customizable field at both the top level and the value level is to provide control over whether a value is **scalar** or not. This allows a check to has values that are not permitted to be modified. @nelsonkopliku please feel free to add or correct me 👍

The thing to enable things that are not permitted looks like a conundrum XD
Either way, if arrays and maps are not permitted to change, but in the same moment are permitted to change... 
just force explicitily to the user to put `customizable: false` in those entries, removing any real logic that auto-detects this I guess.

PR #551, opened 2025-01-21
Operations timeout

arbulu89:
# Description

Handle operation step timeout. When the timeout is reached, the result of missing agent is set to `timeout`.
In order to make it more flexible, the timeout is defined in each step, so we can individualize this value.

By now, the operation is simply terminated. In the future, we can decide to do something else (some sort of rollback, for example)

## How was this tested?

UT added

### DIFF
--- a/lib/wanda/operations/catalog/step.ex
+++ b/lib/wanda/operations/catalog/step.ex
@@ -5,13 +5,17 @@ defmodule Wanda.Operations.Catalog.Step do
   defining if the step needs to be executed in a certain agent.
   """
 
+  @default_timeout 5 * 60 * 1_000
+
   defstruct [
     :operator,
-    :predicate
+    :predicate,
+    timeout: @default_timeout
   ]
 
   @type t :: %__MODULE__{
           operator: String.t(),
-          predicate: String.t()
+          predicate: String.t(),
+          timeout: integer()
         }
 end
--- a/lib/wanda/operations/catalog/step.ex
+++ b/lib/wanda/operations/catalog/step.ex
@@ -16,6 +16,6 @@ defmodule Wanda.Operations.Catalog.Step do
   @type t :: %__MODULE__{
           operator: String.t(),
           predicate: String.t(),
-          timeout: integer()
+          timeout: non_neg_integer()
         }
 end
--- a/lib/wanda/operations/server.ex
+++ b/lib/wanda/operations/server.ex
@@ -123,7 +123,8 @@ defmodule Wanda.Operations.Server do
           operation: %Operation{
             steps: steps
           },
-          current_step_index: current_step_index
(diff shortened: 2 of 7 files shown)


lib/wanda/operations/server.ex
```diff
@@ -19,10 +19,12 @@ defmodule Wanda.Operations.Server do
 
   require Logger
 
-  @default_timeout 5 * 60 * 1_000
-
   @impl true
-  def start_operation(operation_id, group_id, operation, targets, config \\ []) do
+  def start_operation(operation_id, group_id, operation, targets, config \\ [])
```
arbulu89, line 23: This change is unrelated to the timeout feature.
But I already had it there, so...
It basically discards operations if coming `targets` list is empty

lib/wanda/operations/catalog/step.ex
```diff
@@ -5,13 +5,17 @@ defmodule Wanda.Operations.Catalog.Step do
...
-    :predicate
+    :predicate,
+    timeout: @default_timeout
   ]
 
   @type t :: %__MODULE__{
           operator: String.t(),
-          predicate: String.t()
+          predicate: String.t(),
+          timeout: integer()
```
nelsonkopliku, line 19: **nitpick:** `non_neg_integer()`, might be more suitable.

nelsonkopliku: Other question: since timeout is per step, is there a timeout for the whole operation?

test/wanda/operations/server_test.exs
```diff
@@ -344,5 +357,104 @@ defmodule Wanda.Operations.ServerTest do
...
+          "agents" => [
+            %{"agent_id" => agent_id_1, "result" => "updated"},
+            %{"agent_id" => agent_id_2, "result" => "updated"}
+          ]
+        },
+        %{
+          "step_number" => 1,
+          "agents" => [
+            %{"agent_id" => agent_id_1, "result" => "timeout"},
+            %{"agent_id" => agent_id_2, "result" => "timeout"}
```
nelsonkopliku, line 452: **question:** It could also happen that one target did succeed while the other went timeout, correct?

nelsonkopliku (commented):
Nice, just questions for me to keep up to speed.

lib/wanda/operations/catalog/step.ex
arbulu89: No, the whole operation timeout is the sum of all the steps timeouts (in the worst case)

test/wanda/operations/server_test.exs
arbulu89: Yes, it can happen. In that case, only one agent would put its result as timeout putting the operation result as failed

CDimonaco (approved):

PR #552, opened 2025-01-21
Extract abilities from token so they can be matched for authorization

nelsonkopliku:
# Description

This PR makes sure that we're able to integrate authorization policies with Bodyguard.

Actual Policies and plug wire up will be added later.


arbulu89 (commented):
I don't want to establish the way you want to implement this feature, but this opens up some question in my head:
- How are we going to use this new private field?
- Are we going to use bodyguard? Bodyguard works with a user as concept (I think in reality, you can put whatever you want as user anyway)/
- Are we going to store some sort of "virtual" user in the `assign`?
Just wondering all of these, as this small piece of change doesn't really tell me much.

PD: Just to keep in mind that this change is kind of non backward compatible, as old Web versions are not going to send the `abilities` field. Just as a reminder.

nelsonkopliku:
Those are actually good points @arbulu89 thanks for raising them!

> How are we going to use this new private field?
> Are we going to use bodyguard? Bodyguard works with a user as concept (I think in reality, you can put whatever you want as user anyway)
>
I didn't think of using bodyguard because it needs indeed some `User` which we don't have and don't want.
Not sure whether it can work differently, would need to check the lib.

> Are we going to store some sort of "virtual" user in the assign?
>
In the connection private attributes we have `user_id` and `abilities`. Might be useful embedding them in a map.
And I mean a plain map, not even a struct at this point, which could be `conn.private.user` or in `assigns`. Any preference?

The idea is to have a plug that is either controller based or action based.
```elixir
# controller scoped
plug WandaWeb.FooPlug, 
    target: __MODULE__, 
    abilities: ["foo:bar"]

# action scoped
plug WandaWeb.FooPlug, 
    target: {__MODULE__, :foo_action}, 
    abilities: ["foo:bar"]
```

Open for suggestions of course.

arbulu89:
> we can have a simple module that extracts it for us without the whole pow heritage

Yes, we could have something like:
```  
plug Bodyguard.Plug.Authorize,
    policy: Trento.Operations.Policy,
    action: {Phoenix.Controller, :action_name},
    user: {OurNewModule :current_user},
    params: {__MODULE__, :get_policy_resource},
    fallback: WandaWeb.FallbackController
```
###
```
def current_user(conn) do
    %{
      user_id: conn.private[:user_id],
      abilities: conn.private[:abilities]
    }
end
```

And in the policy, we cannot match with a real struct, but it doesn't matter that much. We would still have a user map with `abilities`, which is what we want. Something like:
```
@behaviour Bodyguard.Policy

  def authorize(:index, user, _),
    do: user_has_ability?(user, %{name: "all", resource: "users"})

  def authorize(_, _, _), do: false

  defp user_has_ability?(%{abilities: abilities}, %{name: name, resource: resource}),
    do: Enum.any?(abilities, &(&1["name"] == name and &1["resource"] == resource))
```

arbulu89:
Maybe we can extend this PR a little bit with:
- Install bodyguard dependency
- Create new User module or functionality to get current user
- Handle forbidden failed callback (FallbackController)
- Add the `abilities_helper.ex` support module (as we have in web)
With that, we are ready to add new policies easily.
What do you think?

CDimonaco:
Thanks guys for all the comments, I agree with everything, thanks @arbulu89 for pointing out bodyguard, I completely agree we need to use that for consistency and for not reinventing the whole thing.

Anyway about having a plain map for the user, I suggest if we could use a real struct, I know it's not the "real user", but I think could be beneficial to have it as a typed struct, even named WandaUser or something else just with user id and the abilities

Anyway looking out for all the changes 🚀

nelsonkopliku:
@arbulu89 I believe I addressed everything. 
Actual policies and plug wire up can be deferred to when actually implementing the authorized actions.

@CDimonaco thank you as well for your input. 
Here I added the typed struct f7c695a58ae2974a3fda91bb88dcba23534c70c8 it does not harm and it does not cost much.

Thanks for the review.

lib/wanda_web/auth/user_detector.ex
```diff
@@ -0,0 +1,19 @@
...
+  @moduledoc """
+  This module allows extracting a very simple representation of a user from a connection.
+
+  Only the user_id and abilities are available.
+  """
+
+  alias Wanda.Users.User
+
+  def current_user(%{private: %{user_id: user_id, abilities: abilities}})
+      when not is_nil(user_id) and is_list(abilities) do
```
arbulu89, line 11: There is no way that `abilities` is something that is not a list, right?

test/wanda_web/auth/user_detector_test.exs
```diff
@@ -0,0 +1,59 @@
+defmodule WandaWeb.Auth.UserDetectorTest do
+  use WandaWeb.ConnCase, async: false
```
arbulu89, line 2: Why `async: false`?
Is there a chance to collide with other tests?

arbulu89 (approved):
Nice!
Looks really good.
Just some nitpick comments

lib/wanda_web/auth/user_detector.ex
nelsonkopliku: Hopefully not 😄 being extra careful. Do you prefer us to remove it?

test/wanda_web/auth/user_detector_test.exs
nelsonkopliku: bad copypasta 🙈

lib/wanda_web/auth/user_detector.ex
nelsonkopliku: Actually if in the jwt claims we have `"abilities" => null`, that might be a problem. Very edgy tho

arbulu89: I personally would. Now... Up to you :stuck_out_tongue_closed_eyes:

arbulu89: > Actually if in the jwt claims we have `"abilities" => null`, that might be a problem. Very edgy tho

But you are converting that to an empty list in the jwt code conversion

PR #553, opened 2025-01-21
Move json view tests to correct folder

arbulu89:
# Description

Move JSON rendering tests to correct folder, as we already moved the prod code. We don't have any `views` folder anymore.


EMaksy (approved):
Yeah thanks for taking care, i totally forgot about these!

nelsonkopliku (approved):

PR #555, opened 2025-01-22
Update gen_rmq to cdimonaco/gen_rmq version 5

CDimonaco:
# Description

The library https://github.com/meltwater/gen_rmq is no longer mantained, we need to find an alternative library for our rabbitmq operations, after a spike I decided instead of including the plain `amqp` client and build all the structure around it, to fork `gen_rmq` update to the latest `amqp` and become the mantainer of the library.

Considering `gen_rmq` is basically the same wrapper around `amqp` I was building from scratch, forking the library is the most reasonable choice.

We include `gen_rmq` as github repo because I don't have yet the access to the hex package to release a new `gen_rmq` version, I will update the mix file in the future.

For further information about the library and the fork, https://github.com/CDimonaco/gen_rmq


arbulu89 (approved):
:green_heart: 
Thank you @CDimonaco 
We will need to do the same for `web`.

nelsonkopliku (approved):
Thanks Carmine for the effort. Much appreciated.

PR #557, opened 2025-01-23
relax rust requirement to allow for patch updates

stefanotorresi:
of course the OBS build now fails because 1.81.0 was just pushed out by openSUSE maintainers.


arbulu89 (approved):
I hope we don't have related issues with the docker containers...

stefanotorresi:
nope, containers are fine because the `1.81` tag should automatically be bumped to whatever is the latest patch release.

PR #561, opened 2025-01-29
Customize check action

nelsonkopliku:
# Description

This PR adds on top of https://github.com/trento-project/wanda/pull/556 the initial action to apply checks.


test/wanda_web/controllers/v1/checks_customization_controller_test.exs
```diff
@@ -0,0 +1,278 @@
...
+              source: %{
+                pointer: "/values"
+              },
+              detail: "Array length 0 is smaller than minItems: 1"
+            }
+          ]
+        },
+        %{
+          body: %{
+            values: nil
```
EMaksy, line 60: What do you think about adding a test to check for the missing name ? 
Example:
```%{
  body: %{values: [%{value: 50}]}
  .....
  ```
  
  and 
  ```
  ```%{
  body: %{values: [%{name: "blabla"}]}
  .....
  ```
  ```

**Review by EMaksy — APPROVED** [review]
Hey man good job, just left some little suggestions about the test. 
Besides this good job!

File: test/wanda_web/controllers/fallback_controller_test.exs [hunk 2]
```diff
@@ -28,4 +28,21 @@ defmodule WandaWeb.FallbackControllerTest do
              ]
            } == json_response(conn, 403)
   end
+
+  test "should return not found when referencing a non existent check", %{conn: conn} do
```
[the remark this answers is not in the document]
nelsonkopliku: Added this commit 6aca25d30d5d3c76258fd16000933fadd736d6e8

test/wanda_web/controllers/v1/checks_customization_controller_test.exs
nelsonkopliku: Here we go a988f8d9a0d034f18bb392133f9ae2a5d679b4a3

lib/wanda_web/controllers/error_json.ex
```diff
@@ -10,6 +10,17 @@ defmodule WandaWeb.ErrorJSON do
     }
   end
 
+  def render("404.json", %{reason: reason} = context) when not is_map_key(context, :status) do
```
arbulu89, line 13: curiosity: what is this check for?

lib/wanda_web/controllers/fallback_controller.ex
```diff
@@ -24,6 +24,30 @@ defmodule WandaWeb.FallbackController do
...
+    |> put_status(:not_found)
+    |> put_view(json: ErrorJSON)
+    |> render(:"404", reason: "Referenced check was not found.")
+  end
+
+  def call(conn, {:error, :check_not_customizable}) do
+    conn
+    |> put_status(:forbidden)
+    |> put_view(json: ErrorJSON)
+    |> render(:"403", reason: "Referenced check is not customizable.")
```
arbulu89, line 38: question: shouldn't we put `400` here?
it looks more a thing that the user did wrong, rather than don't having access to this change.
they will have access only if the check itself is modified. he cannot gain real access otherwise

lib/wanda_web/schemas/v1/checks_customizations/custom_value.ex
```diff
@@ -0,0 +1,31 @@
...
+  """
+
+  alias OpenApiSpex.Schema
+
+  require OpenApiSpex
+
+  OpenApiSpex.schema(
+    %{
+      title: "CustomValue",
+      description: "A single check from the catalog",
```
arbulu89, line 13: I think we need to change the description

arbulu89 (approved):
Looks great!
Some nitpick comments

lib/wanda_web/controllers/error_json.ex
nelsonkopliku: some controllers might raise an error like this
```elixir
def show(conn, %{id: execution_id}) do
  execution = Executions.get_execution!(execution_id)

  render(conn, execution: execution)
end
```
and it bubbles up to the fallback controller in a map with `reason` being the error raised, `status` the status code to be returned, and entries.

I added it to keep things working, even though I personally prefer the `with` based approach returning {:error, :something}
```elixir
with {:ok, foo} <- Context.function(arg1, arg2) do
  render(conn, :foo, %{foo: foo})
end
```

We might consider streamlining implementations.

lib/wanda_web/controllers/fallback_controller.ex
nelsonkopliku: If someone calls directly the endpoint attempting to customize an existing check, but that check is not customizable I intend that as a forbidden operation.

Now, details. If you have a strong preference for a bad request I don't mind.

lib/wanda_web/schemas/v1/checks_customizations/custom_value.ex
nelsonkopliku: here we go 63c8aa2d719cf27e0c61c9167df86c7058b7ecca

arbulu89: I don't see the change :see_no_evil:

lib/wanda_web/controllers/fallback_controller.ex
arbulu89: Yeah, nevermind. It is a detail indeed.

lib/wanda_web/schemas/v1/checks_customizations/custom_value.ex
nelsonkopliku: Sorry did not push 😅  2e6b4850e991f53fd0a3eebc8418a4318cefee16

lib/wanda_web/controllers/error_json.ex
nelsonkopliku: This is the best I could get to keep also api redirector tests working c8ce80e07c2023f28020a9afc12f682ae99edb65

PR #562, opened 2025-01-30
Set operation as aborted

arbulu89:
# Description

Set the operation as aborted if the gen server handling the operation crashes abnormally.
For that, first of all, the gen server `restart` value is set to `temporary` so it doesn't restart in any circunstance.

Unfortunately, we cannot catch all the crashes in the gen server itself, so the `terminate/2` is a best-effort. It will put operations to aborted if possible.
If this cannot happen (database connection fails, whole application stops abruptly, brutal kill, etc), the functionality is completed in the `operations` code, where we can compute if an operation is aborted calculating its starting time, steps timeouts and current time.
Of course, this has the disadvantage that the database itself is not 100% accurate.

If we want to solve that and have 100% accurate database content, we need to have a recurring job in the system. This job can be as stupid as a super simple GenServer with a `Process.send_after(self(), :sync, @period)` in a `handle_info(:sync...`, which could run `Opeartions.abort_operation` using the same calculation system.

I don't personally like the recurring job options, as it includes more opaque things on the system. It would do background tasks and makes the system less deterministic as the entries change without any real event.

Opinions?

References:
- [terminate](https://hexdocs.pm/elixir/1.12/GenServer.html#c:terminate/2)
- [article](https://alexcastano.com/how-to-stop-a-genserver-in-elixir/)

## How was this tested?

UT

### DIFF
--- a/lib/wanda_web/controllers/v1/operation_controller.ex
+++ b/lib/wanda_web/controllers/v1/operation_controller.ex
@@ -37,10 +37,13 @@ defmodule WandaWeb.V1.OperationController do
     ]
 
   def index(conn, params) do
-    operations = Operations.list_operations(params)
-    enriched_operations = Enum.map(operations, &Operations.enrich_operation!(&1))
+    operations =
+      params
+      |> Operations.list_operations()
+      |> Enum.map(&Operations.enrich_operation!(&1))
+      |> Enum.map(&Operations.compute_aborted(&1, DateTime.utc_now()))
 
-    render(conn, operations: enriched_operations)
+    render(conn, operations: operations)
   end
 
   operation :show,
@@ -67,6 +70,7 @@ defmodule WandaWeb.V1.OperationController do
       operation_id
       |> Operations.get_operation!()
       |> Operations.enrich_operation!()
+      |> Operations.compute_aborted(DateTime.utc_now())
 
     render(conn, operation: operation)
   end
--- a/lib/wanda/operations.ex
+++ b/lib/wanda/operations.ex
@@ -14,7 +14,7 @@ defmodule Wanda.Operations do
     StepReport
   }
 
-  alias Wanda.Operations.Catalog.Registry
+  alias Wanda.Operations.Catalog.{Registry, Step}
 
   require Wanda.Operations.Enums.Result, as: Result
   require Wanda.Operations.Enums.Status, as: Status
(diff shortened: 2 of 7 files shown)


CDimonaco:
I think this soulution works just fine, having a cronjob just for the sake of computing this in database feels realy like a "cms" kind of thing. When the extreme edge case occours we know how to retrieve the failed.

Implementing other things seems a little bit pointless to me, just to mention that having other tasks that monitor the genserver and spawn them does not solve the problem, it just move the same problem to another component, just to mention because I can understand is a pattern we find a lot in elixir

lib/wanda_web/controllers/v1/operation_controller.ex
```diff
@@ -37,10 +37,13 @@ defmodule WandaWeb.V1.OperationController do
     ]
 
   def index(conn, params) do
-    operations = Operations.list_operations(params)
-    enriched_operations = Enum.map(operations, &Operations.enrich_operation!(&1))
+    operations =
+      params
+      |> Operations.list_operations()
+      |> Enum.map(&Operations.enrich_operation!(&1))
+      |> Enum.map(&Operations.compute_aborted(&1, DateTime.utc_now()))
```
nelsonkopliku, line 44: Have you considered computing whether an operation is aborted or not (based on the time) in the context (`list_operations` and `get_operation` functions).

It would add extra consistency, imho, as we'd always have that logic applied when listing or getting a single operation.

arbulu89: Yes, initially I did.
I would need to change the signature to add the current utc time, which looks a bit out of place (it could be an optional argument as well, like dependency injection only for tests).
The `list_operations` already has a default argument as well, and I could get the time from there, but it looks a bit hacky.
Otherwise testing  becomes impossible
Anyway, I don't have strong opinions

CDimonaco: I would prefer having them separated and composed when needed like in the controller which is the de-facto integration point instead of polluting the signature

nelsonkopliku: Up to you guys.

Anyway the signature might not need to be necessarily changed. We could rely on a clock/datetime behaviour which is mockable in tests. Something similar to what we do in web with `Trento.Support.DateService`

arbulu89: But you still need to pass the configurable `DateService` to the function. How do you run it otherwise?

nelsonkopliku (approved):
LGTM

CDimonaco (approved):

PR #564, opened 2025-02-03
Retrieve customizations for a specific execution group

nelsonkopliku:
# Description

This PR adds the ability to load the customizations applied for a specific checks execution group.

### DIFF
--- a/lib/wanda/checks_customizations.ex
+++ b/lib/wanda/checks_customizations.ex
@@ -20,8 +20,7 @@ defmodule Wanda.ChecksCustomizations do
   def get_customizations(group_id) do
     Repo.all(
       from c in CheckCustomization,
-        where: c.group_id == ^group_id,
-        select: c
+        where: c.group_id == ^group_id
     )
   end
 
--- a/lib/wanda/checks_customizations.ex
+++ b/lib/wanda/checks_customizations.ex
@@ -3,6 +3,8 @@ defmodule Wanda.ChecksCustomizations do
   Customization features.
   """
 
+  import Ecto.Query
+
   alias Wanda.Catalog
   alias Wanda.Catalog.{Check, Value}
   alias Wanda.Catalog.CheckCustomization
@@ -13,6 +15,15 @@ defmodule Wanda.ChecksCustomizations do
           name: String.t(),
           value: any()
         }
+
+  @spec get_customizations(Ecto.UUID.t()) :: [CheckCustomization.t()]
+  def get_customizations(group_id) do
+    Repo.all(
+      from c in CheckCustomization,
+        where: c.group_id == ^group_id,
+        select: c
+    )
+  end
 
   @spec customize(check_id :: String.t(), group_id :: Ecto.UUID.t(), [custom_value()]) ::
           {:ok, CheckCustomization.t()}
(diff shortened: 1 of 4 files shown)


lib/wanda/support/ecto/json.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  def load(data) when is_list(data) or is_map(data) do
+    {:ok, atomize_map_keys(data)}
+  end
+
+  def dump(data) when is_list(data) or is_map(data), do: {:ok, data}
+  def dump(_), do: :error
+
+  defp atomize_map_keys(map) when is_map(map) do
+    Map.new(map, fn {k, v} -> {String.to_existing_atom(k), v} end)
```
arbulu89, line 22: question: I'm wondering if we could make this to work with nested structs?
with some recursion: 
```
...
Map.new(map, fn {k, v} -> {String.to_existing_atom(k), atomize_map_keys(v)} end)

# and add a last match
defp atomize_map_keys(value), do: value
```

It could come handy. I faced this issue right now as well: https://github.com/trento-project/web/pull/3273
My solution is flawed, as we use the JSON conversion in some parts where we don't have plain json maps

lib/wanda/checks_customizations.ex
```diff
@@ -14,6 +16,15 @@ defmodule Wanda.ChecksCustomizations do
           value: any()
         }
 
+  @spec get_customizations(Ecto.UUID.t()) :: [CheckCustomization.t()]
+  def get_customizations(group_id) do
+    Repo.all(
+      from c in CheckCustomization,
+        where: c.group_id == ^group_id,
+        select: c
```
arbulu89, line 24: Do we need this `select`?

test/wanda/checks_customizations_test.exs
```diff
@@ -203,4 +203,69 @@ defmodule Wanda.ChecksCustomizationsTest do
...
+                 check_id: ^check_id_2,
+                 group_id: ^group_id,
+                 custom_values: [
+                   %{
+                     name: ^name3,
+                     value: ^value3
+                   }
+                 ]
+               }
+             ] = ChecksCustomizations.get_customizations(group_id)
```
arbulu89, line 268: We could have a more robuts test with `==` comparison. I think it would still pass and could catch future changes better

arbulu89 (approved):
Looks good @nelsonkopliku !
I dropped a question about making the `load` to work with nested structs. It could come handy in the future

lib/wanda/support/ecto/json.ex
nelsonkopliku: Here we go cad35dccb1e12d6defbb34b2157201e2c754b007.

I was just annoyed by that kind of issue 😄

test/wanda/checks_customizations_test.exs
nelsonkopliku: It adds the necessity to also match on dates, which are not really interesting at this point and quite troublesome 🙈

lib/wanda/checks_customizations.ex
nelsonkopliku: Thanks! da886aa77e3f903e1e949da32796d610ead3a35a

lib/wanda/support/ecto/json.ex
nelsonkopliku: We either go with [`String.to_atom`](https://hexdocs.pm/elixir/1.18.2/String.html#to_atom/1), in which case we'd need to be extra careful that what goes in the map keys is not user provided, or catch the error raised by `String.to_existing_atom` and fallback to plain string.

The problem with the second approach is that we might end up in mixed key maps like
```elixir
%{
  foo: "bar",
  "baz" => 42
}
```
A bit ugly 🙈

EMaksy (approved):
Ah great thanks for this!
LGTM!

PR #566, opened 2025-02-04
Add customization auth policy

nelsonkopliku:
# Description

This PR adds an authorization policy to allow or deny checks customization.

### DIFF
--- a/lib/wanda/catalog/customization_policy.ex
+++ b/lib/wanda/catalog/customization_policy.ex
@@ -1,0 +1,19 @@
+defmodule Wanda.Catalog.CustomizationPolicy do
+  @moduledoc """
+  Policy for checks customizations.
+  """
+  @behaviour Bodyguard.Policy
+
+  import Wanda.Support.AbilitiesHelper
+
+  alias Wanda.Catalog.CheckCustomization
+  alias Wanda.Users.User
+
+  def authorize(:apply_custom_values, %User{} = user, CheckCustomization),
+    do: has_checks_customization_abilities?(user)
+
+  defp has_checks_customization_abilities?(user),
+    do:
+      has_global_ability?(user) or
+        user_has_ability?(user, %{name: "all", resource: "check_customization"})
+end
--- a/test/wanda/catalog/customization_policy_test.exs
+++ b/test/wanda/catalog/customization_policy_test.exs
@@ -1,0 +1,47 @@
+defmodule Wanda.Catalog.CustomizationPolicyTest do
+  use ExUnit.Case
+
+  alias Wanda.Catalog.{CheckCustomization, CustomizationPolicy}
+  alias Wanda.Users.User
+
+  test "should allow checks customization only when required abilities are present" do
+    scenarios = [
+      %{
+        abilities: [],
+        expected: false
+      },
+      %{
+        abilities: [%{name: "foo", resource: "bar"}],
+        expected: false
+      },
+      %{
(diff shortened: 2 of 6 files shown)


lib/wanda/catalog/customization_policy.ex
```diff
@@ -0,0 +1,19 @@
...
+  Policy for checks customizations.
+  """
+  @behaviour Bodyguard.Policy
+
+  import Wanda.Support.AbilitiesHelper
+
+  alias Wanda.Catalog.CheckCustomization
+  alias Wanda.Users.User
+
+  def authorize(:apply_custom_values, %User{} = user, CheckCustomization),
```
arbulu89, line 12: We need to add the `defdelegate` in the `check_customizations.ex` file:
https://hexdocs.pm/bodyguard/readme.html#policies
Example: https://github.com/trento-project/web/blob/main/lib/trento/users/user.ex#L25

It works without it in some cases, but better to simply add it

test/wanda/catalog/catalog_test.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Wanda.CatalogTest do
+defmodule Wanda.Catalog.CatalogTest do
```
arbulu89, line 1: I don't think we need to change this name and file placement.
This tests `Wanda.Catalog` which is directly in `wanda` folder

test/wanda/catalog/checks_customizations_test.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Wanda.ChecksCustomizationsTest do
+defmodule Wanda.Catalog.ChecksCustomizationsTest do
```
arbulu89, line 1: The same. I think this should go in the `test/wanda` folder, as you are testing `wanda/checks_customizations.ex` file.

test/wanda/catalog/customization_policy_test.exs
```diff
@@ -0,0 +1,47 @@
+defmodule Wanda.Catalog.CustomizationPolicyTest do
+  use ExUnit.Case
+
+  alias Wanda.Catalog.{CheckCustomization, CustomizationPolicy}
+  alias Wanda.Users.User
+
+  test "should allow checks customization only to when required abilities are present" do
```
arbulu89, line 7: typo: `to` word is not needed

test/wanda_web/controllers/v1/checks_customization_controller_test.exs
```diff
@@ -263,7 +264,69 @@ defmodule WandaWeb.V1.ChecksCustomizationsControllerTest do
...
+          %{
+            name: "all",
+            resource: "check_customization"
+          }
+        ]
+      }
+    ]
+
+    for %{name: scenario_name, abilities: allowing_abilities} <- allowing_abilities do
+      @tag abilities: allowing_abilities
```
arbulu89, line 289: question. does this replace the `abilities` in the context? interesting

arbulu89 (changes requested):
Looks good!
I'm requesting changes as I see some file movements that surprise me

test/wanda_web/controllers/v1/checks_customization_controller_test.exs
nelsonkopliku: Yes, in `UserAwareConnCase` we either use default abilities or what has been defined in `@tag`

nelsonkopliku:
Thanks @arbulu89! Addressed all the feedbacks.

PR #573, opened 2025-02-13
Improve api routes

nelsonkopliku:
# Description

Routing improvements as discussed here https://github.com/trento-project/wanda/pull/572, at least the immediately actionables.

Pipeline fails on API bc check, obv 😄 Gonna force-merge.

### DIFF
--- a/lib/wanda_web/router.ex
+++ b/lib/wanda_web/router.ex
@@ -57,14 +57,18 @@ defmodule WandaWeb.Router do
         get "/groups/:id/executions/last", ExecutionController, :last
         post "/executions/start", ExecutionController, :start
         get "/catalog", CatalogController, :catalog
+      end
 
-        get "/groups/:group_id/catalog", CatalogController, :selectable_checks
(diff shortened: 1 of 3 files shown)


lib/wanda_web/router.ex
```diff
@@ -57,14 +57,18 @@ defmodule WandaWeb.Router do
...
+      end
+
+      scope "/groups" do
+        pipe_through [:api_v1, :protected_api]
 
-        get "/groups/:group_id/catalog", CatalogController, :selectable_checks
+        get "/:group_id/checks", CatalogController, :selectable_checks
 
-        post "/:check_id/customize/:group_id",
+        post "/:group_id/check/:check_id/customization",
```
arbulu89, line 67: question: shouldn't this and the delete have `checks` instead of `check`?

nelsonkopliku: whoopsie

arbulu89 (approved):

PR #575, opened 2025-02-17
Multiple amqp consumer publisher

arbulu89:
# Description
In order to work with operations, using individual amqp queues for this context looks appropriate.
This code gives the chance to start multiple consumers and producers based on macro code.
So each context will define its own consumer and producer.

We can replicate this model in `web` as well.

PD: I need to fix dialyzer erros but for that I need to update `gen_rmq` first

## How was this tested?

UT


config/config.exs
```diff
@@ -3,27 +3,50 @@ import Config
 config :wanda, Wanda.Messaging.Publisher, adapter: Wanda.Messaging.Adapters.AMQP
 
 config :wanda,
-  children: [Wanda.Messaging.Adapters.AMQP.Publisher, Wanda.Messaging.Adapters.AMQP.Consumer]
+  children: [
+    Wanda.Executions.Messaging.Publisher,
+    Wanda.Executions.Messaging.Consumer,
+    Wanda.Operations.Messaging.Publisher,
+    Wanda.Operations.Messaging.Consumer
+  ]
```
gagandeepb, line 11: I wonder if there's any impact to be considered from the order of starting these children? E.g. Does it matter that the operations Publisher is started before Consumer?

gagandeepb (approved):
Looks good; I have two questions of the informational sort.

config/config.exs
arbulu89: I don't think there would be any difference, but technically it is better to start the Publisher first.
The consumer might start the operation server that will use the publisher later on. So having the publisher always ready before consuming events and for the moment we have the dynamic server starts is required.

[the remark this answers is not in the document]
arbulu89: The operations dynamic server will start before those, but in any case, it won't never be used until we consume something from the queue, so in practical scenario we are covered, as the operations server will be ready to run when we start receiving events

PR #577, opened 2025-02-19
Operations messaging with web

arbulu89:
# Description

Consume and publish operations related amqp messages.
Basically, it consumes the `OperationRequested` event and publishes `OperationStarted/OperationCompleted` with different results.

I have renamed and updated some `mapper` functions to have an easier time to understand. The functions starting with `from` map messages coming in the events to internal types. Starting with `map` do the opposite

## How was this tested?

UT


CDimonaco (approved):
LGTM : 🚀

PR #584, opened 2025-03-03
Using custom values in checks executions

nelsonkopliku:
# Description

This PR allows customized values to be used during checks executions. 
Additionally:
- each `CheckResult` is marked as customized when necessary
- each `Value` in `AgentCheckResult` is marked as customized when necessary

With regards to OpenApi spec: 
- `values` entry spec was missing in `AgentCheckResult` so it has been added
- in order to avoid another version bump at this time, new fields have beed added in a backward compatible manner (although java heap is complaining 🤦 )

### DIFF
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -82,13 +82,11 @@ defmodule Wanda.Catalog do
     |> Enum.map(&map_to_selectable_check(&1, available_customizations))
   end
 
-  @spec get_current_selection([String.t()], String.t(), map()) :: [SelectedCheck.t()]
-  def get_current_selection(ids, group_id, env \\ %{}) do
+  @spec to_selected_checks([Check.t()], String.t()) :: [SelectedCheck.t()]
+  def to_selected_checks(checks, group_id) do
     available_customizations = ChecksCustomizations.get_customizations(group_id)
 
-    ids
-    |> get_checks(env)
-    |> Enum.map(&map_to_selected_check(&1, available_customizations))
+    Enum.map(checks, &map_to_selected_check(&1, available_customizations))
   end
 
   defp map_to_selectable_check(
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -3,6 +3,9 @@ defmodule Wanda.Catalog do
   Function to interact with the checks catalog.
   """
 
+  alias Wanda.Catalog.CustomizedValue
+  alias Wanda.Catalog.SelectedCheck
+
   alias Wanda.Catalog.{
     Check,
     CheckCustomization,
@@ -79,6 +82,15 @@ defmodule Wanda.Catalog do
     |> Enum.map(&map_to_selectable_check(&1, available_customizations))
   end
 
+  @spec get_current_selection([String.t()], String.t(), map()) :: [SelectedCheck.t()]
+  def get_current_selection(ids, group_id, env \\ %{}) do
+    available_customizations = ChecksCustomizations.get_customizations(group_id)
+
+    ids
+    |> get_checks(env)
+    |> Enum.map(&map_to_selected_check(&1, available_customizations))
+  end
+
   defp map_to_selectable_check(
          %Check{
            id: id,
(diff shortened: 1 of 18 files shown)


lib/wanda/catalog/selected_check.ex
```diff
@@ -0,0 +1,26 @@
...
+  defstruct [
+    :id,
+    :spec,
+    :customized,
+    :customizations
+  ]
+
+  @type t :: %__MODULE__{
+          id: String.t(),
+          spec: Check.t(),
```
arbulu89, line 19: nitpick: what if we name this `check`? It embeds a `Check`. Anyway, after understanding the whole flow, it is not that confusing.

lib/wanda/executions/evaluation.ex
```diff
@@ -109,19 +119,32 @@ defmodule Wanda.Executions.Evaluation do
...
+        },
+        customizations: customizations
+      } = selected_check
+
       value_evaluation_scope = add_scope(%{"env" => env}, "facts", facts)
 
-      evaluated_values = Enum.map(values, &eval_value(&1, value_evaluation_scope, engine))
+      resolved_values =
+        Enum.map(spec_values, fn %CatalogValue{name: name} = specified_value ->
+          customization = Enum.find(customizations, &(&1.name == name))
```
arbulu89, line 134: Maybe we could change the `Enum.find` to add a default value as 2nd parameter with the specified value.
This way, if the custom value is not there, we would get the initial value, and we could pass only one value to `eval_value`, and pattern match based on the type, instead of having 2 variables.
It looks a bit cleaner in my opinion having the `eval_value` without partially used variables

arbulu89 (approved):
Hey @nelsonkopliku 
Great job.

I left some personal comments. They are based on how I understood better the code. As always, they are mine and don't need to be better than the current code. Just posted just in case if you like them as well. No need to change.

PD: I don't know why the BC is failing, as we are adding new fields. I don't know what was the other alternative though

lib/wanda/executions/evaluation.ex
nelsonkopliku: the problem is that the _specified value_ has to be evaluated based on the default value and conditions.

lib/wanda/catalog.ex
```diff
@@ -79,6 +82,15 @@ defmodule Wanda.Catalog do
     |> Enum.map(&map_to_selectable_check(&1, available_customizations))
   end
 
+  @spec get_current_selection([String.t()], String.t(), map()) :: [SelectedCheck.t()]
```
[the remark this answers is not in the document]
nelsonkopliku: I was actually thinking to make get_checks private altogether as it's not used anywhere at this stage 😅 and keep only the `get_current_selection`

lib/wanda/executions/evaluation.ex
arbulu89: yes, don't we already do that? 
I mean, if we don't have customizations the currrent `eval_value` should do it (with minor changes).
At the end is just pattern matching:
```
Enum.map(spec_values, fn %CatalogValue{name: name} = specified_value ->
    customizations
    |> Enum.find(specified_value, &(&1.name == name))
    |> eval_value(value_evaluation_scope, engine)
end)

...
defp eval_value(
         %CatalogValue{
           name: name,
           default: default,
           conditions: conditions
         },
         evaluation_scope,
         engine
       ) do
    %Value{
      name: name,
      value: find_value(conditions, default, evaluation_scope, engine),
      customized: false
    }
  end

  defp eval_value(
         %CustomizedValue{name: name, value: custom_value},
         _,
         _
       ) do
    %Value{
      name: name,
      value: custom_value,
      customized: true
    }
  end
```

nelsonkopliku: Oh now I get it. I did not think of this option tbh.
The only _weird_ aspect is that the `eval_value` function expects two totally different input types as the first argument: either a `CatalogValue` or a `CustomizedValue`.

You know what, I don't dislike it. Thanks bf3fc7121582aa7dcb2a8cd61df9f83f4691d576

lib/wanda/catalog.ex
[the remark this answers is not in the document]
nelsonkopliku: hey @arbulu89 I made an attempt here 449054658f62daf0391686049ea687b83def85bf.
In this case I still would prefer encapsulating a functionality that returns a list of selected checks in the catalog.

Just wondering whether changing the initial name `get_current_selection` to something else would have helped.

In any case I am fine either way.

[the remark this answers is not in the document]
nelsonkopliku: Not going to change again 🙈 keeping the last changes 😉

PR #592, opened 2025-03-11
Expose resolved original values

nelsonkopliku:
# Description

This PR allows to expose the resolved original values based on current context.
This is handy during checks customization to tell users what they're going to customize.

Additionally `current_value` has been renamed to `original_value` as the former sounds confusing because a _current value_ could be either the original one as defined in the check spec or a custom value.

Alternatives to `original_value` could be `initial_value`, `builtin_value`, `specified_value`, etc... open for suggestions.

Notes:
- clearly the ci complains about a breaking change in the api
- a tiny update in the ui is required https://github.com/trento-project/web/pull/3380

### DIFF
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -170,17 +170,13 @@ defmodule Wanda.Catalog do
        ),
        do: Map.put(value, :default_value, default_value)
 
-  defp maybe_add_customization(%{customizable: customizable} = value, %ResolvedValue{
-         customized: customized
-       })
-       when not customized or not customizable,
-       do: value
-
   defp maybe_add_customization(
          %{customizable: true} = initial_value,
          %ResolvedValue{custom_value: custom_value, customized: true}
        ),
        do: Map.put(initial_value, :custom_value, custom_value)
+
+  defp maybe_add_customization(initial_value, _resolved_value), do: initial_value
 
   defp read_check(path, check_id) do
     case YamlElixir.read_from_file(path) do
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -157,18 +157,18 @@ defmodule Wanda.Catalog do
         name: value_name
       }
       |> add_value_customizability(value_spec, customization_globally_disabled?)
-      |> maybe_add_original_value(resolved_value)
+      |> maybe_add_default_value(resolved_value)
       |> maybe_add_customization(resolved_value)
     end)
   end
 
-  defp maybe_add_original_value(%{customizable: false} = value, _), do: value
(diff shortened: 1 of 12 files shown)


EMaksy (approved):

lib/wanda/catalog/selectable_check.ex
```diff
@@ -6,14 +6,14 @@ defmodule Wanda.Catalog.SelectableCheck do
   @type customized_value :: %{
           name: String.t(),
           customizable: true,
-          current_value: boolean() | number() | String.t(),
+          original_value: boolean() | number() | String.t(),
```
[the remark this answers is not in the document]
arbulu89: nitpick: I would rename this to `default_value` as it is the default value we fallback in case there is not any modification done

lib/wanda/catalog/resolved_value.ex
```diff
@@ -0,0 +1,26 @@
...
+  """
+  alias Wanda.Catalog.Value
+
+  @type value :: boolean() | number() | String.t()
+
+  @derive Jason.Encoder
+  defstruct [
+    :spec,
+    :name,
+    :original_value,
```
arbulu89, line 14: Remember to change this if you decide to change the other in the `SelectableCheck`

lib/wanda/catalog.ex
```diff
@@ -138,49 +137,50 @@ defmodule Wanda.Catalog do
...
-  defp maybe_add_current_value(%{customizable: true} = value, default_value),
-    do: Map.put(value, :current_value, default_value)
+  defp maybe_add_original_value(
+         %{customizable: true} = value,
+         %ResolvedValue{original_value: original_value}
+       ),
+       do: Map.put(value, :original_value, original_value)
 
-  defp maybe_add_customization(%{customizable: false} = value, _), do: value
+  defp maybe_add_customization(%{customizable: customizable} = value, %ResolvedValue{
```
arbulu89, line 173: nitpick: is it not simpler to put this match after the next one without any `when`?
```
  defp maybe_add_customization(initial_value, _resolved_value), do: initial_value
```
I mean, if both customizable and customized are true we take the custom, otherwise we take the default value

arbulu89 (approved):
It looks good @nelsonkopliku 
I vote for using `default_value` instead of `original_value`, as at the end, it is the fallback value we use if none are changed (and we use `Default` label in the frontend as well).

lib/wanda/catalog.ex
nelsonkopliku: Well yes 😄 4a265c21d90a38ac543941d68ca9e92402684b52

nelsonkopliku:
Thanks @arbulu89 `default_value` -> 9bcf5e220f11eccb61730e6d4aeef4a6a2756c83

PR #594, opened 2025-03-12
Check vendor deps on CI

balanza:
# Description

To avoid future issues like in https://github.com/trento-project/wanda/pull/593, we can check the rust dependencies are correctly vendorized when adding a PR

Depends on #593

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -408,7 +408,7 @@ jobs:
       - name: Setup rust
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.66
+          toolchain: 1.81
       - name: Get mix deps
         run: mix local.hex --force && mix local.rebar --force && mix deps.clean --all && mix deps.get
       - name: cargo vendor
@@ -488,7 +488,7 @@ jobs:
       - name: Setup rust
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.66
+          toolchain: 1.81
       - name: Get mix deps
         run: mix local.hex --force && mix deps.clean --all && mix deps.get
       - name: Cargo vendor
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -51,6 +51,44 @@ jobs:
           mix deps.compile --warnings-as-errors
           mix dialyzer --plt
 
+  check-vendor-deps:
```
nelsonkopliku, line 54: **question:** 
- how does this job affect the CI? 
- None of the other jobs is depending on this. Is this on purpose?

balanza: If the job fails the CI is marked as failed (see https://github.com/trento-project/wanda/pull/594/commits/396aa4cd302e5cdf2eac7b7ebf0d7b821607c2a2)

I don't think it will add more value to set this job as a dependency for any other job. It's just a canary to anticipate a possible build problem at build time

nelsonkopliku: I understand, we would see if cargo vendoring fails also in a PR where `obs-commit` and `obs-commit-rpm` are not run. It makes sense.

**thought:** an opportunity for improvement could be making the result of the `cargo vendor` of this job available to `obs-commit` and `obs-commit-rpm` when they're executed. At the end they run the same step.
However, it wouldn't change much in terms of CI performance, so this is good enough.

nelsonkopliku (approved):
I guess you might want to rebase, but LGTM.

stefanotorresi:
Why wasn't this merged for 2 months? 🫤 
I'm only noticing this now, and #611 will surely go in conflict with this.

By the way, while I'm okay adding the job to at least do a smoke test of cargo dependencies, problems like #593 are addressed in #611 by using `.tool-versions` as the only place where we set the rust version, so the workflows should never go out of sync again.

That said, I realize that something like this job is still needed because nowhere in our current testing pipeline we ever actually do anything with rust, apparently: it's only invoked when vendoring for the OBS packages.
Maybe we should explore the usage of [`RHAI_RUSTLER_FORCE_BUILD`](https://github.com/rhaiscript/rhai_rustler?tab=readme-ov-file#rustler-precompiled).

PR #599, opened 2025-03-18
Receive operator execution reports

arbulu89:
# Description

Receive and handler `OperatorExecutionCompleted` messages sent by the agent and process the result.

Now the `diff` or `error_message` sent in the report are stored as well in the database and exposed in the operation API.

I have changed the `agent_reports` field to be of `JSON` type, so we use `atom` entries everywhere as all the used keys are well known atoms.

## How was this tested?

UT and manual testing


CDimonaco (approved):
lgtm

PR #602, opened 2025-04-01
Publish checks customization messages

nelsonkopliku:
# Description

This PR adds message publishing for Checks customization:
- when a customization is applied
- when a customization is updated (same message as the previous)
- when a customization is reset

Notes:
- custom value type changed from number to integer, as it is elsewhere
- passing the user id around will follow up

### DIFF
--- a/lib/wanda/checks_customizations.ex
+++ b/lib/wanda/checks_customizations.ex
@@ -162,19 +162,21 @@ defmodule Wanda.ChecksCustomizations do
 
       {:error, :publish, reason, _} ->
         Logger.error(
-          "Error while publishing check customization message, error: #{inspect(reason)}"
+          "Error while publishing check customization message. check: #{check_id}, group_id: #{group_id}, error: #{inspect(reason)}"
         )
 
         {:error, reason}
 
       {:error, reason} = error ->
-        Logger.error("Error while applying custom values, error: #{inspect(reason)}")
+        Logger.error(
+          "Error while applying custom values. check: #{check_id}, group_id: #{group_id}, error: #{inspect(reason)}"
+        )
 
         error
     end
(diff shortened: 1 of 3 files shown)


lib/wanda/messaging/mapper.ex
```diff
@@ -261,6 +301,11 @@ defmodule Wanda.Messaging.Mapper do
 
   defp map_value(value), do: %{kind: {:string_value, value}}
 
+  defp map_inner_value(value) when is_integer(value), do: {:int_value, value}
```
arbulu89, line 304: comment: maybe I would put a comment on top of them explaining that these are not `Google.Protobuf.Value` values.
It took me a while to understand it.
Or rename to `map_check_custom_value`, so it is clear.
Yes, I would go with the renaming

EMaksy (approved):
LGTM, just one typo i think :)

lib/wanda/checks_customizations.ex
```diff
@@ -120,26 +126,128 @@ defmodule Wanda.ChecksCustomizations do
...
+        {:error, reason}
+
+      {:error, reason} = error ->
+        Logger.error("Error while applying custom values, error: #{inspect(reason)}")
+
+        error
     end
   end
+
+  defp reset_check_customization(%{} = metadata, check_id, group_id) do
```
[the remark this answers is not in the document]
nelsonkopliku: Sort of type hinting saying that `metadata` is a map. I can very well remove it

lib/wanda/messaging/mapper.ex
nelsonkopliku: will rename to `map_check_custom_value`

PR #604, opened 2025-04-07
Propagate user in messages

nelsonkopliku:
# Description

This PR makes sure the user id is propagated with checks customization messages.

need to re-bump contracts when https://github.com/trento-project/contracts/pull/97 is merged

---

For completeness I was also considering a different approach that might result to something like this in the controller
```elixir
check_id
|> ChecksCustomizations.customize(group_id, custom_values)
|> ChecksCustomizations.apply_transactionally(
  fn message ->
    case Messaging.publish(Publisher, "customizations", message, user_id: user_id) do
      :ok ->
        {:ok, :published}

      {:error, reason} ->
        {:error, reason}
    end
  end
)
```
It's an alternative to passing `opts` as a 4th argument to `ChecksCustomizations.customize`.

There is something I do like in this approach, but I felt more confident at this time doing something maybe more familiar to  what we usually do.
We might re-consider in the future.


test/wanda/checks_customizations_test.exs
```diff
@@ -616,4 +601,39 @@ defmodule Wanda.ChecksCustomizationsTest do
         where: c.group_id == ^group_id
     )
   end
+
+  defp get_scenario_opts(scenario) do
```
arbulu89, line 605: super nitpick: I personally find the usage of this helper of not much help unless you wrote it hehe. Kind of implicit logic with "hidden" defaults.

arbulu89 (approved):

test/wanda/checks_customizations_test.exs
nelsonkopliku: either that or the Map.get repeated 😄

PR #605, opened 2025-04-10
Deduplicate execution targets

nelsonkopliku:
# Description

Relates to this https://github.com/trento-project/web/pull/3447

Deduplicates targets for checks executions.


arbulu89 (approved):
I guess this is a just in case solution if we have some mistake in `web`, right?

PR #607, opened 2025-05-08
Add ipv6 bug fix

EMaksy:
# Description

This pr provides a fix for wanda, currently when wanda is started and the machine does not have ipv6 the app fails to start.
This happens because the [Cowboy webserver Plug](https://hexdocs.pm/plug_cowboy/Plug.Cowboy.html) requires either an ipv4 or ipv6 address. For this we can use the erlang inet getaddr function

Same fix as in [web ](https://github.com/trento-project/web/pull/3490)

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -36,13 +36,22 @@ if config_env() in [:prod, :demo] do
 
   port = String.to_integer(System.get_env("PORT") || "4000")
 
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):
Same comment as in web eheheh https://github.com/trento-project/web/pull/3490#pullrequestreview-2827413345

PR #608, opened 2025-05-12
Add saptune change solution operation to the catalog

nelsonkopliku:
# Description

Adds the saptune change solution operation spec to the operations catalog

Current tests already cover necessary scenarios.

Related to https://github.com/trento-project/web/pull/3488


arbulu89 (approved):
:green_heart:

PR #615, opened 2025-05-29
Fix deps cache key in CI

nelsonkopliku:
# Description

Fixes the elixir cache key reference.

Pipelines like this one fail because cached deps could not be found https://github.com/trento-project/wanda/actions/runs/15321794247/job/43107296198?pr=600

Here's a working example https://github.com/trento-project/wanda/actions/runs/15322336128/job/43108691057


arbulu89 (approved):

stefanotorresi (approved):
thanks for catching this!

PR #616, opened 2025-05-29
Clustermaintenancechange

arbulu89:
# Description

Add `clustermaintenancechange@v1` operation

### DIFF
--- a/lib/wanda/operations/catalog/clustermaintenancechange_v1.ex
+++ b/lib/wanda/operations/catalog/clustermaintenancechange_v1.ex
@@ -1,0 +1,20 @@
+defmodule Wanda.Operations.Catalog.ClusterMaintenanceChangeV1 do
+  @moduledoc false
+
+  use Wanda.Operations.Catalog.Operation,
+    operation: %Wanda.Operations.Catalog.Operation{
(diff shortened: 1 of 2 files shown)


lib/wanda/operations/catalog/clustermaintenancechange_v1.ex
```diff
@@ -0,0 +1,29 @@
...
+        is_dc (boolean): Whether the target is the designated controller of the cluster
+        resource_id (string): The ID of the cluster resource to change the state (this has precedence over node_id)
+        node_id (string): The ID of the cluster node to change the state
+      """,
+      required_args: ["maintenance", "is_dc"],
+      steps: [
+        %Wanda.Operations.Catalog.Step{
+          name: "Change maintenance state",
+          operator: "clustermaintenancechange@v1",
+          predicate: "is_dc==true"
```
nelsonkopliku, line 25: ocd alert 
```suggestion
          predicate: "is_dc == true"
```
besides this, also `predicate: "is_dc"` would work, right? not that I have preferences, just asking

nelsonkopliku (approved):
Great!

PR #618, opened 2025-06-03
[WIP] Demo operations

nelsonkopliku:
# Description

This PR aims to add the ability to have faked operations, similarly to what we have for checks executions.

The underlying idea is the same: we have a configuration file that specifies what to assume agents return - operation reports in this case.

Alternatives to this approach could be:
- fully faked agents (well, we don't have it)
- a more sophisticated mechanism where we spawn different processes (ie genservers) that behave like an agent:
  - listen to the same `OperatorExecutionRequested` message as an agent would do
  - determine which mocked report to emit
  - emit the mocked report via a `OperatorExecutionCompleted` message as an agent would do

### DIFF
--- a/lib/wanda/operations/reporting.ex
+++ b/lib/wanda/operations/reporting.ex
@@ -1,0 +1,138 @@
+defmodule Wanda.Operations.Reporting do
+  @moduledoc """
+  Reporting module for Wanda operations.
+  """
+  alias Wanda.Operations.{
+    AgentReport,
+    OperationTarget,
+    OperatorError,
+    OperatorResult,
+    StepReport
+  }
+
+  alias Wanda.EvaluationEngine
+
+  require Wanda.Operations.Enums.OperatorPhase, as: OperatorPhase
+  require Wanda.Operations.Enums.Result, as: Result
+
+  def initialize_report_results(targets, steps) do
+    not_executed_agent_reports =
+      Enum.map(targets, fn %OperationTarget{agent_id: agent_id} ->
+        %AgentReport{agent_id: agent_id, result: Result.not_executed()}
+      end)
+
+    agent_reports =
+      steps
+      |> Enum.with_index()
+      |> Enum.map(fn {_, index} ->
+        %StepReport{step_number: index, agents: not_executed_agent_reports}
+      end)
+
+    agent_reports
+  end
+
+  def evaluate_agent_reports_result(agent_reports) do
+    agent_reports
+    |> Enum.flat_map(fn %{agents: agents} -> agents end)
+    |> Enum.map(& &1.result)
+    |> Enum.map(&{&1, result_weight(&1)})
+    |> Enum.max_by(fn {_, weight} -> weight end)
+    |> elem(0)
+  end
+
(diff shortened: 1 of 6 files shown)


lib/wanda/operations/reporting.ex
```diff
@@ -0,0 +1,138 @@
+defmodule Wanda.Operations.Reporting do
```
nelsonkopliku, line 1: @arbulu89 I would appreciate feedback on this approach.

Decoupling this `Reporting` module enables the mocking we want to introduce, besides improving the general code structure in terms of responsibility of components.

If you look at checks execution a similar approach made it possible in the first place to have such mocking.

Naming can be further improved: both the module name and functions name (I mainly kept similar if not the same to the ones that were embedded in the server)

priv/demo/fake_operations_reports.yaml
```diff
@@ -0,0 +1,19 @@
...
+          solution: ""
+        after: 
+          solution: "HANA"
+  "saptunechangesolution@v1":
+    0:
+      target1: 
+        before: 
+          solution: "HANA"
+        after: 
+          solution: "S4HANA-DBSERVER"
```
nelsonkopliku, line 19: This structure is meant to define what report each `target` is going to return for a given operation/step.

References to `target1`, `target2` will be resolved with relevant uuids when parsing the file. 
It is handy to keep the yaml tidy and less noisy by avoiding repetition of uuids

arbulu89 (commented):
Hey @nelsonkopliku ,
I have been thinking about other alternative solution that maybe is simpler to implement. It consists on using the real operations server in the fake server and sending the `receive_reports` message from it.

It would we something like:
```
defmodule Wanda.Operations.FakeServer do

@behaviour Wanda.Operations.ServerBehaviour

@impl true
def start_operation(operation_id, group_id, operation, targets, config) do
    # maybe we need to handle the return from this, specially the already_running one
    Server.start_operation(operation_id, group_id, operation, targets, config)

    # we start reporting for each step and each target
    Enum.forEach(steps, ...
        Enum.forEach(targets, ...
            report = get_report() # we get report from yaml or created with faker
            Server.receive_operation_reports(operation_id, group_id, 0, agent_id_1, report)
            # sleep, other things...
            ...
end

@impl true
def receive_operation_reports(_, ...), do: :ok
```

Maybe it is simpler, we use all the already existing logic, so we don't need to extract things from the current code, and at the end, the difference between the demo and the real one are just the how the reports are received.

What do you think?

Other aspects:
Knowing that the demo environment is a fully controlled environment and we know exactly the state of all hosts, I think we could already have a predefined state of returns, including the states. For example:
- We don't need to check the predicate. We know that this step in this host should be skipped
- We can compose the whole report (result, diff, error (if it failed), etc)

nelsonkopliku:
Hey @arbulu89, thanks! It looks worth trying that path. Let me look into that.

I still have the feeling that some sort of decoupling of what happens inside the server might be beneficial, but for the sake of the task at hand - demoing things - such refactor can be not considered here.

PR #619, opened 2025-06-05
Add SAP instance start/stop operations

arbulu89:
# Description

Add SAP instance start/stop operations.

### DIFF
--- a/lib/wanda/operations/catalog/registry.ex
+++ b/lib/wanda/operations/catalog/registry.ex
@@ -7,12 +7,16 @@ defmodule Wanda.Operations.Catalog.Registry do
 
   alias Wanda.Operations.Catalog.{
     ClusterMaintenanceChangeV1,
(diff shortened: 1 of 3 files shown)


arbulu89:
> @arbulu89 just a question: do we run start/stop on _every_ node? 👀

If you send multiple targets, yes. But that's not the normal use case (surely not if the request is done by `web`).
This is more of a one host operation (like saptune operations).

nelsonkopliku:
> > @arbulu89 just a question: do we run start/stop on _every_ node? 👀
> 
> If you send multiple targets, yes. But that's not the normal use case (surely not if the request is done by `web`). This is more of a one host operation (like saptune operations).

I see, I am wondering whether it could be worth considering some different way to make it clear that some operation are host bound. First thing that might come to mind is either not having the `predicate` at all or have something similar to a target type.

nelsonkopliku (approved):

PR #622, opened 2025-06-09
Add enable/disable pacemaker operations spec

nelsonkopliku:
# Description

Add operations spec for:
- enabling pacemaker
- disabling pacemaker


lib/wanda/operations/catalog/disablepacemaker_v1.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Wanda.Operations.Catalog.DisablePacemakerV1 do
```
arbulu89, line 1: I guess we will update to `PacemakerDisable`

lib/wanda/operations/catalog/registry.ex
```diff
@@ -18,7 +20,9 @@ defmodule Wanda.Operations.Catalog.Registry do
     "sapinstancestart@v1" => SapInstanceStartV1.operation(),
     "sapinstancestop@v1" => SapInstanceStopV1.operation(),
     "saptuneapplysolution@v1" => SaptuneApplySolutionV1.operation(),
-    "saptunechangesolution@v1" => SaptuneChangeSolutionV1.operation()
+    "saptunechangesolution@v1" => SaptuneChangeSolutionV1.operation(),
+    "enablepacemaker@v1" => EnablePacemakerV1.operation(),
```
arbulu89, line 24: Let's try to keep and alphabethical order on the map.
It helps to me at least to find things when it gets big

arbulu89 (approved):
Good!
We need to update the names. Nothing besides that

lib/wanda/operations/catalog/disablepacemaker_v1.ex
nelsonkopliku: done

PR #626, opened 2025-06-23
Filter operations by status

arbulu89:
# Description

Filter operations by status in the `operations/executions` api call.
We need this to query the currently ongoing operations from wanda. Related: https://github.com/trento-project/web/pull/3591

Unfortunately, this is not as simple as just doing `where([status: ^status])`.
We need to take care of operations that were suddenly canceled without putting the `aborted` value in the `status` field.

I have update how we compute aborted operations, so it is done in the postgresql `sql` query, as we need for querying exact items per page. If we do later on, with other filter, we might get less values than given by the DB.

The logic is simple. Add the `timeout_at` value during operation creation.
With this, check if an operation is running and the timeout is already expired. In this case, the operation is set to `aborted`.

I have opted for this option as it looks the cleaner. We could've used the postgresql `values` utility to have "in memory" table joined to the current table with the timeouts, but it would make the code more volatile, as removing any entry from the registry would change the returned values.

Everything else is just testing parafernalia to have a mocked date.

## How was this tested?

UT

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -88,7 +88,8 @@ config :rustler_precompiled, :force_build, rhai_rustler: true
 config :wanda,
   cors_enabled: true,
   jwt_authentication_enabled: true,
-  operations_enabled: true
+  operations_enabled: true,
+  date_service: Wanda.Support.DateService
 
 config :bodyguard,
   default_error: :forbidden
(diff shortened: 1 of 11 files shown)


nelsonkopliku (approved):
TIL: parafernalia

PR #629, opened 2025-07-17
Add SAP system start/stop operations

arbulu89:
# Description

Add SAP system start/stop operations

I'm leaving the PR open until @abravosuse is back. I want to discuss with him if we need to have some more specific timeouts for the steps. Maybe the default 5 minute is not enough, as these operations might take more time than this one.


nelsonkopliku (approved):

arbulu89:
Setting 1 hour timeout for both operations after talking offline with @abravosuse 
We might decide to give the user the option to adjust a smaller timeout for the operator itself. But this value will come from `web` in any case

PR #632, opened 2025-07-29
Add database start/stop operations

arbulu89:
# Description

Add database start/stop operations.

They have a very large timeout, as start/stopping a production database might take a really long time.
Either way, the operator includes other `timeout` which we will use to give the user the capability to select a more specific timeout from the frontend.

### DIFF
--- a/lib/wanda/operations/catalog/databasestart_v1.ex
+++ b/lib/wanda/operations/catalog/databasestart_v1.ex
@@ -6,7 +6,9 @@ defmodule Wanda.Operations.Catalog.DatabaseStartV1 do
       id: "databasestart@v1",
       name: "Start HANA database",
       description: """
-      This operation starts a complete HANA database.
+      This operation starts a HANA database. If the database is in a system
+      replication configuration, it only starts the site where the given target
+      belongs to. Other sites are not impacted.
 
       Arguments:
         instance_number (string): Instance number to start the database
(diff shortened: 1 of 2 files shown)


lib/wanda/operations/catalog/databasestart_v1.ex
```diff
@@ -0,0 +1,28 @@
...
+      """,
+      required_args: ["instance_number"],
+      steps: [
+        %Wanda.Operations.Catalog.Step{
+          name: "Start HANA database",
+          operator: "sapsystemstart@v1",
+          predicate: "*",
+          # 12 hours
+          timeout: 12 * 60 * 60 * 1_000
+        }
```
nelsonkopliku, line 25: **question:** `timeout` argument is what will be provided to the operator while `timeout` in the Step specification is the time wanda would wait for the step to complete.

Is it safe to assume that the argument must always be less than 12 hours?

arbulu89: > **question:** `timeout` argument is what will be provided to the operator while `timeout` in the Step specification is the time wanda would wait for the step to complete.
> 
Yes. Basically to avoid having wanda hang forever.

> Is it safe to assume that the argument must always be less than 12 hours?
Yes, it is an assumption. 12 hours is an extra large timeout that shouldn't be surpassed in any case. I don't know how we will set the limits for the "user" timeout, but it shouldn't be bigger than 12 hours

PR #638, opened 2025-08-25
Elixir 1.18 and Erlang/OTP 27

gagandeepb:
# Description

Bumps upwards the application stack: Elixir and Erlang/OTP. Rust also bumped up towards 1.88. Depends on https://github.com/trento-project/contracts/pull/113 .


nelsonkopliku (approved):
As long as everything builds and works LGTM 😄

PR #642, opened 2025-09-11
[TRNT-3845] Run linters on the API spec and fix the errors (controller)

antgamdia:
# Description

This PR continues the work in https://github.com/trento-project/wanda/pull/634 and starts passing some linters (`redocly`, `vacuum` and `spectral`) over the API spec file. For now, the main goal is just to reduce the number of errors being found. Reaching zero errors/warnings is not yet intended.

Particularly, this PR performs some breaking changes that need a closer review.

Related # TRNT-3845

## How was this tested?

`openapi-diff`

### DIFF
--- a/test/wanda_web/controllers/v1/checks_customization_controller_test.exs
+++ b/test/wanda_web/controllers/v1/checks_customization_controller_test.exs
@@ -136,7 +136,7 @@ defmodule WandaWeb.V1.ChecksCustomizationsControllerTest do
           |> put_req_header("content-type", "application/json")
           |> post("/api/v1/groups/#{group_id}/checks/#{check_id}/customization", invalid_body)
           |> json_response(:unprocessable_entity)
-          |> assert_schema("JsonErrorResponse", api_spec)
+          |> assert_schema("UnprocessableEntity", api_spec)
 
         assert %{errors: ^expected_errors} = response
       end


lib/wanda_web/controllers/health_controller.ex
```diff
@@ -10,12 +10,15 @@ defmodule WandaWeb.HealthController do
   }
 
   operation :ready,
-    summary: "Wanda ready",
-    tags: ["Platform"],
-    description: "Check if Wanda is ready",
+    summary: "Wanda ready.",
+    tags: ["Wanda Platform"],
```
arbulu89, line 14: Could we leave this as `Platform`. Adding `Wanda` looks redundant

arbulu89 (commented):
Can we remove the `Wanda` word from the tags?
It is redundant, isn't it?

lib/wanda_web/controllers/health_controller.ex
antgamdia: I partially agree with it. The problem with `Platform` is that it overlaps with the same tag in TrentoWeb. As we might need to combine both specs in a single file (for the MCP), this name is not ok. On the other hand, adding `Wanda` is indeed redundant. 
What do you think of something like `Checks Platform` ?

PR #645, opened 2025-09-17
[TRNT-3845] Run API documentation linter in CI

antgamdia:
# Description

This PR adds a script we recently added for linting our API docs (using `redocly`, `vacuum` and `spectral`).

For now, it is merely informative; that is, it does not fail even if the linters are not passing. We'll have to revert the `|| true` once the API docs are ready.

Related # TRNT-3845

## How was this tested?

CI, example: TBD


lib/wanda_web/router.ex
```diff
@@ -30,6 +30,11 @@ defmodule WandaWeb.Router do
     plug OpenApiSpex.Plug.PutApiSpec, module: WandaWeb.Schemas.V3.ApiSpec
   end
 
+  pipeline :unversioned_api do
+    plug :api
+    plug OpenApiSpex.Plug.PutApiSpec, module: TrentoWeb.OpenApi.Unversioned.ApiSpec
```
arbulu89, line 35:
```suggestion
    plug OpenApiSpex.Plug.PutApiSpec, module: WandaWeb.Schemas.Unversioned.ApiSpec
```

arbulu89 (approved):
all good!
only a typo in one of the aliased modules

PR #649, opened 2025-09-25
[TRNT-3845] Publish merged API docs in CI

antgamdia:
# Description

This PR updates the script we have to run the linters and adds a "join-only" option. This approach is now used by CI, so, instead of `mix openapi.spec.json --start-app=false --spec TrentoWeb.OpenApi.V1.ApiSpec` (which only generates the V1 docs), it now delegates the generation to the script. This script yields an `openapi.json` file, combining all the specs in a single file.

Related # TRNT-3845

See https://github.com/trento-project/web/pull/3799

## How was this tested?

N/A


arbulu89:
Hey @antgamdia ,
I have been thinking about this. I have other proposal: https://github.com/trento-project/web/pull/3811
Which reduces the need for external tuning of our specs

PR #650, opened 2025-09-25
[TRNT-3845] Tag endpoints for MCP

antgamdia:
# Description

This PR adds the `MCP` tag in the API docs; besides, it tags some endpoints with it. The idea is to flag those endpoints that are ok to be exposed via LLM.

For now, this is just a proposal; we need to check with more stakeholders before merging.

## How was this tested?

CI


arbulu89 (commented):
good. The same as web, but I looks good to me (one endpoint to clarify)

arbulu89 (approved):
Ok!
Let's go with this 1st option at least.
It looks safe enough

PR #688, opened 2026-04-01
Lock targets parallel execution during operation

arbulu89:
# Description

Lock targets parallel execution during operation.

**Disclaimer:**
This implementation wouldn't cover all the scenarios right now, as it only works properly if all operations send all the targets composing one resource (example: all instances of a SAP system). 
This doesn't happen always. For example, for SAP operations, we only send one target, as it is enough to execute the operation.

I opened this PR just to show one option, but I have zero attachments, and seeing it doesn't cover everything, I'm open to discuss any other alternative.
____

Until now, parallel execution is only blocked by `group_id`.
For most of the cases this is enough, but we still need to have a additional protection against having parallel operations in the same target.

I have decided to continue using the `:global` registry and lock the target resources once the operations starts. When the operation is finished the locks are released again.
Using `:global` has the advantage of:
- It works in a distributed erlang cluster
- It avoids any race-condition as locking is a atomic operation

I have considered other options:
- `Registry`: I have discarded this as the `Registry` is a local node registry, so it wouldn't work in a distributed system
- Using the db to store the targets: we actually have the list of targets of currently running operations. But it would require a more complex usage, as first you need to read the currently running operations and get the targets, and after checking if all the targets in the new operation are not there, start a new one, saving the new targets there. Here we would have some race-condition if not managed properly
- Some external dependency like `Horde`: I didn't want to include this kind of "big" dependency for such a simple task

## How was this tested?

UT and manual testing

## Did you update the documentation?

No need

### DIFF
--- a/lib/wanda/operations/server.ex
+++ b/lib/wanda/operations/server.ex
@@ -102,12 +102,24 @@ defmodule Wanda.Operations.Server do
   end
 
   @impl true
-  def init(%State{operation_id: operation_id} = state) do
+  def init(%State{operation_id: operation_id, targets: targets} = state) do
     Logger.debug("Starting operation: #{operation_id}", state: inspect(state))
 
     Process.flag(:trap_exit, true)
 
-    {:ok, state, {:continue, :start_operation}}
+    target_ids = Enum.map(targets, & &1.agent_id)
+
+    # lock targets usage in the init, so the process is self contained and all
+    # the locks are released once the operation itself is finished
+    case lock_targets(target_ids) do
+      :ok ->
+        {:ok, state, {:continue, :start_operation}}
+
+      {:error, :target_already_in_use} ->
+        Logger.error("Operation already in progress in some of the requested targets")
+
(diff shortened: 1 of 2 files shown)


lib/wanda/operations/server.ex
```diff
@@ -400,6 +412,42 @@ defmodule Wanda.Operations.Server do
...
+    # List of all nodes in a distributed cluster
+    nodes = [node() | Node.list()]
+    # The requester_id is set to self as this is a new GenServer process.
+    # If this function was used outside of the short lived GenServer,
+    # self() might not be a good choice
+    requester_id = self()
+
+    locking_result =
+      target_ids
+      |> Enum.sort()
```
antgamdia: suggestion: not sure if possible, but should we filter out duplicates just in case?

antgamdia (commented):
Thanks!
I know it is still a draft, so feel free to disregard any comments that are no longer relevant. I have mainly added some questions to better understand it.

PR #698, opened 2026-04-09
Exclude our common workflows from cooldown

skrech:
I suggest we exclude our own common re-usable workflows from the cooldown period, so we can rapidly update our CI.

Warning: I haven't tested this change. I'm not entirely sure if this is the correct "package name" for what we want to exclude.

PS. Excuse this sloppy PR description, I'm making it from my phone. :)

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -6,6 +6,7 @@ updates:
       interval: "weekly"
     cooldown:
       default-days: 14
+      exclude: "trento-project/.github*"
   - package-ecosystem: "mix"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


antgamdia (approved):

gagandeepb (commented):
I am not sure I understand the "what" of this PR. Can you please explain /provide some examples of situations where this specific exclusion will execute? Additionally, what are the specific dependencies that this exclusion will exclude ?

antgamdia:
I guess it is to see if Dependabot can send PRs as soon as possible when one of the shared reusable workflows we own (at https://github.com/trento-project/.github) is updated,
Since we are actively changing those workflows these days, having Dependabot send PRs ASAP would avoid extra manual work of fanning out the version bump-up in every repo.

skrech:
Thanks Antonio, yes this is the reason.

gagandeepb (approved):
I am not sure about the naming scheme for excludes/ or whether this will work as intended. But happy to help iterate, if needed/necessary.

PR #712, opened 2026-05-11
[TRNT-4376] Align DB dependencies

antgamdia:
# Description

I noticed some tests are failing ([example](https://github.com/trento-project/wanda/actions/runs/24457669610/job/71463254264)); this PR is to try to discover and fix the issue causing it.

This PR is aligning the versions we use in development and in CI to the ones we use in the Helm Chart (manually for now, though). Using `latest` was causing some issues in out test cases.

Besides, it extracts some additional versions to env dependencies, for maintainability. 

Fixes # TRNT-4376

## How was this tested?

CI

Before the fix, it fails ([example](https://github.com/trento-project/wanda/actions/runs/25674566418/job/75368842001?pr=712)); after the fix, it works ([example](https://github.com/trento-project/wanda/actions/runs/25675187863/job/75371143719?pr=712)).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,6 +19,8 @@ env:
   MIX_ENV: test
   ELIXIR_BC: "1.15.7-otp-26"
   ERLANG_BC: "26.2.1"
+  NODEJS_BC: "22.15.1"
+  OPENAPI_DIFF_VERSION: "2.1.7" # https://github.com/OpenAPITools/openapi-diff/releases
 
 jobs:
   setup-matrix-env:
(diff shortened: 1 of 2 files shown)


vicenteqa (approved):
lgtm, I know this development is about aligning, but I see that we're still pinning the postgres version in two different places, perhaps we need to think in some way to have a single source of truth.

antgamdia:
> I see that we're still pinning the postgres version in two different places, perhaps we need to think in some way to have a single source of truth.

Yes, absolutely. Besides, this is also applicable to Web :(   ~I'll be creating a PR shortly~. See https://github.com/trento-project/web/pull/4290.

https://github.com/trento-project/web/blob/main/docker-compose.yaml#L62
https://github.com/trento-project/web/blob/main/.github/workflows/ci.yaml#L35

We could do it in a shared repo, set up a GHA to, once it changes, perform some replacements in every affected repo? But at least, the bare minimum is to ensure, even manually, that we are using the same version in the same repo 😅

nelsonkopliku (approved):

PR #722, opened 2026-06-08
Add x-ai-tool information to MCP tagged operations

nelsonkopliku:
# Description

This PR introduces an `x-ai-tool` [extension](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specification-extensions) entry to relevant AI related operations (MCP tagged).

The idea is to be able to drive `tool name` and `display text` - and possibly more in the future - as we do in https://github.com/trento-project/web/pull/4330 so that we can better control AI integration.

As an extra I also untagged deprecated endpoints from being MCP. I couldn't find solid reasons to keep deprecated endpoint at use of AI agents.

AI feature-set is quite new and access to latest operations should be enough. Additionally it might mitigate the risk of hallucinations of LLMs as there would not be ambiguity.

Let me know what you think.

## How was this tested?

No specific tests needed here.

### DIFF
--- a/lib/wanda_web/controllers/info_controller.ex
+++ b/lib/wanda_web/controllers/info_controller.ex
@@ -11,7 +11,7 @@ defmodule WandaWeb.InfoController do
     summary: "Wanda service information.",
     tags: ["Checks/Operations Platform"],
     description:
-      "This endpoint returns information about this Wanda instance for service discovery purposes, including name and version",
+      "This endpoint returns information about this Wanda instance for service discovery purposes, including name and version.",
     security: [],
     responses: [
       ok:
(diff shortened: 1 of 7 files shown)


arbulu89 (approved):
Looks good to me!

One question. Did you follow certain pattern to set the `name` values? Or is it just a "meaningful" name?
Just asking to know what the expect. 

PD:
You linked openAPI 3.1 extension there.
The extension system is the same for 3.0 anyway.
Just as nit

balanza (approved):

nelsonkopliku:
> One question. Did you follow certain pattern to set the `name` values? Or is it just a "meaningful" name? Just asking to know what the expect.
> 
whatever is meaningful and unique.

> PD: You linked openAPI 3.1 extension there. The extension system is the same for 3.0 anyway. Just as nit
>
Thanks! Yes we are on 3.0.

PR #729, opened 2026-07-02
Handle check exclude predicate 

balanza:
Following RFC https://github.com/trento-project/docs/pull/188

In this PR, an optional `exclude` predicate is evaluated before check execution to exclude. For agents for which the predicate verifies, the check is not evaluated against them and they are reported as "excluded by policy" in the check result.

Please note that the fact-gathering dispatch mechanism has not been touched, because it would have introduced a disruption too big to fit in the scope of these changes. It might be refined later.

### DIFF
--- a/lib/wanda/executions/agent_check_result.ex
+++ b/lib/wanda/executions/agent_check_result.ex
@@ -16,13 +16,19 @@ defmodule Wanda.Executions.AgentCheckResult do
   @derive Jason.Encoder
   defstruct [
     :agent_id,
+    :status,
+    :exclude_expression,
     values: [],
     facts: [],
     expectation_evaluations: []
   ]
 
+  @type status :: :excluded_by_policy | nil
+
   @type t :: %__MODULE__{
           agent_id: String.t(),
+          status: status(),
+          exclude_expression: String.t() | nil,
           facts: [Fact.t()],
           values: [Value.t()],
           expectation_evaluations: [ExpectationEvaluation.t() | ExpectationEvaluationError.t()]
--- a/lib/wanda/executions/agent_check_result.ex
+++ b/lib/wanda/executions/agent_check_result.ex
@@ -16,14 +16,14 @@ defmodule Wanda.Executions.AgentCheckResult do
   @derive Jason.Encoder
   defstruct [
     :agent_id,
-    :status,
     :exclude_expression,
+    status: :executed,
     values: [],
     facts: [],
     expectation_evaluations: []
   ]
 
-  @type status :: :excluded_by_policy | nil
+  @type status :: :executed | :excluded
 
   @type t :: %__MODULE__{
           agent_id: String.t(),
--- a/lib/wanda/executions/agent_check_result.ex
+++ b/lib/wanda/executions/agent_check_result.ex
@@ -19,7 +19,7 @@ defmodule Wanda.Executions.AgentCheckResult do
   defstruct [
     :agent_id,
     :exclude_expression,
-    status: :executed,
+    status: AgentCheckStatus.excluded(),
     values: [],
     facts: [],
     expectation_evaluations: []
--- a/lib/wanda/executions/agent_check_result.ex
+++ b/lib/wanda/executions/agent_check_result.ex
@@ -13,6 +13,8 @@ defmodule Wanda.Executions.AgentCheckResult do
     Value
   }
 
+  require Wanda.Executions.Enums.AgentCheckStatus, as: AgentCheckStatus
+
   @derive Jason.Encoder
   defstruct [
     :agent_id,
@@ -23,11 +25,9 @@ defmodule Wanda.Executions.AgentCheckResult do
     expectation_evaluations: []
   ]
 
-  @type status :: :executed | :excluded
-
   @type t :: %__MODULE__{
           agent_id: String.t(),
-          status: status(),
+          status: AgentCheckStatus.t(),
           exclude_expression: String.t() | nil,
           facts: [Fact.t()],
           values: [Value.t()],
--- a/lib/wanda/executions/evaluation.ex
+++ b/lib/wanda/executions/evaluation.ex
@@ -435,6 +435,10 @@ defmodule Wanda.Executions.Evaluation do
         _ -> false
       end)
 
+  defp aggregate_execution_result(%Result{check_results: []} = execution) do
+    %Result{execution | result: ResultEnum.passing()}
+  end
+
   defp aggregate_execution_result(%Result{check_results: check_results} = execution) do
     result =
       check_results
--- a/lib/wanda/executions/evaluation.ex
+++ b/lib/wanda/executions/evaluation.ex
@@ -24,6 +24,7 @@ defmodule Wanda.Executions.Evaluation do
   alias Wanda.EvaluationEngine
 
   require Wanda.Catalog.Enums.ExpectType, as: ExpectType
+  require Wanda.Executions.Enums.AgentCheckStatus, as: AgentCheckStatus
   require Wanda.Executions.Enums.Result, as: ResultEnum
 
   @default_failure_message "Expectation not met"
@@ -142,6 +143,7 @@ defmodule Wanda.Executions.Evaluation do
 
       %AgentCheckResult{
         agent_id: agent_id,
+        status: AgentCheckStatus.executed(),
         facts: facts,
         values: resolved_values,
         expectation_evaluations:
--- a/lib/wanda/executions/evaluation.ex
+++ b/lib/wanda/executions/evaluation.ex
@@ -13,6 +13,7 @@ defmodule Wanda.Executions.Evaluation do
     AgentCheckError,
     AgentCheckResult,
     CheckResult,
+    ExcludedCheckResult,
     ExpectationEvaluation,
     ExpectationEvaluationError,
     ExpectationResult,
@@ -35,10 +36,20 @@ defmodule Wanda.Executions.Evaluation do
           [SelectedCheck.t()],
           map(),
           %{String.t() => boolean() | number() | String.t()},
+          [ExcludedCheckResult.t()],
           [String.t()],
           Rhai.Engine.t()
         ) :: Result.t()
-  def execute(execution_id, group_id, checks, gathered_facts, env, timeouts \\ [], engine) do
+  def execute(
+        execution_id,
+        group_id,
+        checks,
+        gathered_facts,
+        env,
+        excluded_checks,
+        timeouts \\ [],
+        engine
+      ) do
     %Result{
       execution_id: execution_id,
       group_id: group_id,
@@ -50,6 +61,7 @@ defmodule Wanda.Executions.Evaluation do
       env,
       engine
     )
+    |> add_excluded_checks_result(excluded_checks, checks)
     |> aggregate_execution_result()
   end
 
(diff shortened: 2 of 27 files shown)


nelsonkopliku:
Still haven't gone deeper in the changes, but we're gonna need something like this https://github.com/trento-project/wanda/pull/732 to have a better mockability/demoability of the feature and in general of the checks execution.

lib/wanda/executions/server.ex
```diff
@@ -228,9 +282,137 @@ defmodule Wanda.Executions.Server do
...
+  defp inject_excluded_checks(%Result{check_results: _check_results} = result, []),
+    do: result
+
+  defp inject_excluded_checks(%Result{check_results: check_results} = result, excluded) do
+    excluded_by_check = Enum.group_by(excluded, & &1.check_id)
+
+    new_check_results =
+      Enum.map(check_results, fn %CheckResult{check_id: check_id} = check_result ->
+        inject_excluded_agents(check_result, Map.get(excluded_by_check, check_id))
+      end)
```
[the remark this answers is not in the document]
nelsonkopliku: we might have an issue here indeed: if we select more than one check for a cluster (`checkA` and `checkB`) and one of those checks - ie `checkB` - has to be excluded from all the hosts of the cluster (can be tested by having a check with `exclude: "true"`), then `checkB` will not be part of the checks result as if it wasn't selected at all

lib/wanda/executions/result.ex
```diff
@@ -16,13 +16,15 @@ defmodule Wanda.Executions.Result do
     :group_id,
     :result,
     check_results: [],
+    excluded_checks: [],
```
nelsonkopliku: **question:**

this new field seems always set to an empty list
- `excluded_checks: []` [server.ex:154](https://github.com/trento-project/wanda/pull/729/changes#diff-c2ddcabead7c24091a7652708c728613e146c5ff6acfe9e93bfa7f97ee16d2ffR154)
- `|> Map.put(:excluded_checks, [])` [server.ex:234](https://github.com/trento-project/wanda/pull/729/changes?w=1#diff-c2ddcabead7c24091a7652708c728613e146c5ff6acfe9e93bfa7f97ee16d2ffR234)
- `|> Map.put(:excluded_checks, [])` [server.ex:266](https://github.com/trento-project/wanda/pull/729/changes?w=1#diff-c2ddcabead7c24091a7652708c728613e146c5ff6acfe9e93bfa7f97ee16d2ffR266)

It seems always empty. Who is consuming that field?

nelsonkopliku:
@balanza I prepared a draft PR here https://github.com/trento-project/wanda/pull/733 that addresses the following
- check absent when all nodes of a cluster are excluded https://github.com/trento-project/wanda/pull/729#discussion_r3518286488
- supposedly unused `Wanda.Executions.Result.excluded_checks` field https://github.com/trento-project/wanda/pull/729#discussion_r3537451056

I find the first more compelling, though.

Feel free to take that draft as a source of suggestions.

lib/wanda/executions/result.ex
balanza: fixed in https://github.com/trento-project/wanda/pull/729/commits/112b4dd9456bda96fc62727aa5796682439e65d2

lib/wanda/executions/server.ex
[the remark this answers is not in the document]
balanza: Fixed in https://github.com/trento-project/wanda/pull/729/commits/2e34c79f5bb34a2749440a8d3030915d80b2b94e

lib/wanda/executions/excluded_check_result.ex
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-License-Identifier: Apache-2.0
+
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
balanza: fixed in 9bc9b30f

lib/wanda/messaging/mapper.ex
```diff
@@ -416,4 +426,20 @@ defmodule Wanda.Messaging.Mapper do
...
   end
+
+  defp unwrap_proto_value(%{kind: {:string_value, s}}), do: s
+
+  defp unwrap_proto_value(%{kind: {:bool_value, b}}), do: b
+
+  defp unwrap_proto_value(%{kind: {:number_value, n}}), do: n
+
+  defp unwrap_proto_value(%{kind: {:null_value, _}}), do: nil
+
```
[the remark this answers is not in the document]
balanza: fixed in 9bc9b30f

test/wanda/executions/server_test.exs
```diff
@@ -317,4 +317,342 @@ defmodule Wanda.Executions.ServerTest do
...
+          attributes: %{"any_attribute" => "including_value"}
+        )
+      ]
+
+      expect(Wanda.Messaging.Adapters.Mock, :publish, 2, fn _, _, _, _ -> :ok end)
+
+      assert :ok = Server.start_execution(UUID.uuid4(), group_id, targets, "cluster", %{})
+
+      # Wait for handle_continue to run
+      :timer.sleep(50)
```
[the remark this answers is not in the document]
balanza: fixed in 9bc9b30f

test/wanda/executions/server_test.exs
```diff
@@ -317,4 +317,342 @@ defmodule Wanda.Executions.ServerTest do
...
+      start_supervised!(
+        {Server,
+         execution_id: UUID.uuid4(),
+         group_id: group_id,
+         targets: targets,
+         checks: [selected_check],
+         env: %{}}
+      )
+
+      :timer.sleep(50)
```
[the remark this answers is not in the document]
balanza: fixed in 9bc9b30f

test/wanda/executions/server_test.exs
```diff
@@ -317,4 +317,342 @@ defmodule Wanda.Executions.ServerTest do
...
```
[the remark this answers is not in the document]
balanza: fixed in 9bc9b30f

test/wanda/executions/server_test.exs
```diff
@@ -317,4 +317,342 @@ defmodule Wanda.Executions.ServerTest do
...
```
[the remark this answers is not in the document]
balanza: fixed in 9bc9b30f

lib/wanda/executions/agent_check_result.ex
```diff
@@ -16,13 +16,19 @@ defmodule Wanda.Executions.AgentCheckResult do
   @derive Jason.Encoder
   defstruct [
     :agent_id,
+    :status,
```
arbulu89: I'm not sure if `status` is a proper name for this concept. It looks to generic for a such concrete event.
Specially, since returning `nil` will be the normal thing, and `status: nil` doesn't look really well.
Maybe we should be more concrete on this, and have something like `execution_excluded: true/false`.
I assume that you added `status` forseeing future additions, but what this could be?

lib/wanda/executions/evaluation.ex
```diff
@@ -435,6 +435,10 @@ defmodule Wanda.Executions.Evaluation do
         _ -> false
       end)
 
+  defp aggregate_execution_result(%Result{check_results: []} = execution) do
```
arbulu89, line 518: question:
this is a safeguard to protect against some execution where all agents are excluded, right?
Even though in practice shouldn't happen (we shouldn't have checks excluding all agents), it protects some corner cases that for A or B it happens.

lib/wanda/messaging/mapper.ex
```diff
@@ -416,4 +426,25 @@ defmodule Wanda.Messaging.Mapper do
   defp from_value(map) when is_map(map) do
     Enum.into(map, %{}, fn {key, value} -> {key, from_value(value)} end)
   end
+
+  defp unwrap_proto_value(%{kind: {:string_value, s}}), do: s
```
arbulu89: question:
Doesn't this reimplement the `from_value` function?
The intention of that function is the same of this `unwrap_proto_value` at least.
Is there maybe some tiny difference?
At the end, both are `map<string, google.protobuf.Value>` and `from_value` should cover all use cases

lib/wanda/messaging/mapper.ex
```diff
@@ -97,10 +97,20 @@ defmodule Wanda.Messaging.Mapper do
         target_type: target_type,
         env: env
       }) do
+    plain_targets =
+      Enum.map(targets, fn %{agent_id: agent_id, checks: checks} = item ->
```
arbulu89: question:
couldn't we simplify this to? At the end, `attributes` will always come, or not?
```
plain_targets =
    Enum.map(targets, fn %{agent_id: agent_id, checks: checks, attributes: attributes}  ->
        %{agent_id: agent_id, checks: checks, attributes: from_value(attributes)}
    end)
```

If you are adding the `attributes` logic for backward compatibility, I think protobuf itself guarantees that.
If old `web` with old contract without the `attributes` field sends a request, the receiver will default to an empty map.

arbulu89 (commented):
Hey @balanza ,
A preliminary review, specially on the wrapper thing.
I wonder we did you choose to create the new `unwrap_` function and not use the `from_value` one.

lib/wanda/executions/agent_check_result.ex
arbulu89: Commented offline, but adding here.
I finally liked the `status` name. Maybe instead of being with current values `:excluded_by_policy | nil`, we could make it an enum to `[:executed, :excluded]`. This way we don't have the nil value, which i think is better.

lib/wanda/messaging/mapper.ex
balanza: It's a good point, it was there in the beginning and then with refactors it became useless. Fixed in https://github.com/trento-project/wanda/pull/729/commits/a693e5d1a0da0c888dcfbe4d9488dde4813697a4

lib/wanda/messaging/mapper.ex
balanza: Done in https://github.com/trento-project/wanda/pull/729/commits/a693e5d1a0da0c888dcfbe4d9488dde4813697a4

lib/wanda/executions/evaluation.ex
balanza: Yes, it's a safeguard for edge cases

> in practice shouldn't happen (we shouldn't have checks excluding all agents)

I was on the same page. However, you never know what the exclude predicate is about... today we are only considering majority maker flags, but in the future we can have less static values (i.e. health?) and to exclude all the host can be actually a thing

lib/wanda/executions/agent_check_result.ex
balanza: It's a fair concern. Having `nil` makes the `status` field not exist in normal execution. Which is something I did to not change the payload on tests that do not use `exclude`.

However, we can achieve the same result using a sane default on the consumer (i.e. trento-web). 

if we want to change, I'd probably go with the enum
```
status: executed | excluded_by_policy
```
Or we can rename `status` with `execution` or `execution_status` if it's more specific.

What do you think? (It will require changing https://github.com/trento-project/web/pull/4439 as well, but it's trivial)

balanza: Anyway, I just refactored in https://github.com/trento-project/wanda/pull/729/commits/78f7f2eaed0fceefef18a9b68e0ed9981719f3f6 and consequently in https://github.com/trento-project/web/pull/4439

arbulu89:
Hey @balanza ,
As commented offline, I have appended some commits which basically do:
- Move the `executed/excluded` atoms to an enum module
- Move the addition of the excluded checks results to the `evaluation`module this. We way, we can reuse it in the initial `if/else` condition where we create an empty check results if all checks are excluded.
- Refactor the `evaluate_exclusions`. Simply I think it is simpler (my perspective, of course).
What do you think?

arbulu89 (approved):
Green light from my side!

PR #731, opened 2026-07-02
Document `exclude` predicate

balanza:
Following RFC https://github.com/trento-project/docs/pull/188

Depends on #729 

Document the usage of the `exclude` predicate

### DIFF
--- a/guides/specification.adoc
+++ b/guides/specification.adoc
@@ -385,7 +385,7 @@ metadata:
 An optional Rhai expression evaluated against each host's metadata (`+host+`
 link:#evaluation-scope[scope]) before the check evaluation is performed.
 Hosts for which the predicate returns `+true+` are removed from the target
-set and reported as `+excluded_by_policy+` in the execution results.
+set and reported as `+excluded+` in the execution results.
 
 *Default:* if no `+exclude+` is provided, no host is excluded based on this
 predicate.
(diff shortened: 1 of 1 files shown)


abravosuse (commented):
Thank you @balanza . Am I understanding the provided example in the wrong way?

`exclude: host.is_majority_maker == false`

When does predicate `host.is_majority_maker == false` returns `true`? When the host is a majority maker or when it is not?

balanza:
> Thank you @balanza . Am I understanding the provided example in the wrong way?
> 
> `exclude: host.is_majority_maker == false`
> 
> When does predicate `host.is_majority_maker == false` returns `true`? When the host is a majority maker or when it is not?

It was actually a typo, sorry. `host.is_majority_maker == true` excludes the majority maker nodes, as you may expect

abravosuse (approved):
Thank you!

guides/specification.adoc
```diff
@@ -376,6 +378,47 @@ metadata:
   qux: [foo, bar, baz]
 ----
 
+==== exclude
+
+An optional Rhai expression evaluated against each host's metadata (`+host+`
+link:#evaluation-scope[scope]) before the check evaluation is performed.
+Hosts for which the predicate returns `+true+` are removed from the target
+set and reported as `+excluded_by_policy+` in the execution results.
```
arbulu89: can we change this to simply `excluded`?
```suggestion
set and reported as `+excluded` in the execution results.
```

arbulu89 (approved):
Thank you!

nelsonkopliku (approved):

PR #732, opened 2026-07-03
Improve Checks execution mockability

nelsonkopliku:
### Description

This PR adds a `FactsGathering` abstraction that allows us to have a better mockability of checks executions.

Additionally:
- the fake gathering is made the default behavior in dev
- previous `FakeServer` required some duplication to behave as the actual server. Now it is gone
- added hostnames in `priv/demo/fake_facts.yaml` that reflect the actual hostname in the demo scenario.
- actual mocking point moved closer to edge (amqp exchange with agents) so that prod/demo drift cannot happen

### How was this tested?

IRL + tests


lib/wanda/executions/facts_gathering/facts_gathering_fake.ex
```diff
@@ -0,0 +1,76 @@
...
+defmodule Wanda.Executions.FactsGathering.Fake do
+  @moduledoc """
+  Demo/dev facts gathering source: it does not reach out to any agent. Instead
+  it synthesizes fake facts (see `Wanda.Executions.FakeGatheredFacts`) and feeds
+  them back into `Wanda.Executions.Server` through the same `receive_facts/4`
+  entry point the real agents use.
+
+  Because of this the execution flows through the exact same evaluation,
+  exclusion and finalization path as in production: `exclude` predicates are
+  honoured and `excluded_by_policy` results are produced in demo just like with
```
arbulu89: outdated:
Let's replace `exclude_by_policy` by simply `excluded`

lib/wanda/executions/facts_gathering/fake.ex
```diff
@@ -0,0 +1,76 @@
+# SPDX-FileCopyrightText: SUSE LLC
```
arbulu89, line 1: question:
any reason not to name the file` fake.ex`? (just following the normal convention we have to name the module name after the file name).
Otherwise, i would rename at least the `fake_test.exs` to have the same name.

test/wanda/executions/facts_gathering/fake_test.exs
```diff
@@ -0,0 +1,146 @@
...
+    previous_server = Application.get_env(:wanda, Server)
+    previous_facts = Application.get_env(:wanda, Wanda.Executions.FakeGatheredFacts)
+
+    Application.put_env(:wanda, Server,
+      facts_gathering_impl: Wanda.Executions.FactsGathering.Fake
+    )
+
+    Application.put_env(:wanda, Wanda.Executions.FactsGathering.Fake, sleep: 0)
+
+    Application.put_env(:wanda, Wanda.Executions.FakeGatheredFacts,
```
arbulu89: question:
isn't this the default configuration for `test`? configured in `test.exs`.
maybe it was set to avoid issues with other concurrent tests, but the test is marked as sinc.
we could put as weel the `sleep: 0` as default in `test.exs` configuration.
at the end, i don't understand match the usage of `previous_facts` thing. Do we have other fake facts for test?

arbulu89 (approved):
I like the changes!

PD:
some minor nitpicks. Ignore if you see fit!

nelsonkopliku:
Thanks @arbulu89 for the feedback! Addressed here 693928d615dfc65a5c5715b81b4ee084da413a83

PR #733, opened 2026-07-07
[WIP] Improvements to check exclusion

nelsonkopliku:
### Description

This PR addresses two specific comments in https://github.com/trento-project/wanda/pull/729
- check hidden when all nodes of a cluster are excluded https://github.com/trento-project/wanda/pull/729#discussion_r3518286488
- supposedly unuses `Wanda.Executions.Result.excluded_checks` field https://github.com/trento-project/wanda/pull/729#discussion_r3537451056

image

---

As a side note I realized that if on a cluster wide check (aka `expect_same`) all the nodes are excluded, we still se the aggregated result for the cluster, however that might be a false positive/negative because if none of the nodes of a cluster were taken into account for a cluster wide check, then the cluster wide check expectation does not make sense anymore.

image

That said, probably this is an edge case we might not want to consider just now. Leaving it here for reference anyway.

### How was this tested?

Automated and IRL.

### DIFF
--- a/lib/wanda/catalog.ex
+++ b/lib/wanda/catalog.ex
@@ -256,6 +256,7 @@ defmodule Wanda.Catalog do
        remediation: remediation,
        metadata: Map.get(check, "metadata"),
        when: Map.get(check, "when"),
+       exclude: Map.get(check, "exclude"),
        severity: map_severity(check),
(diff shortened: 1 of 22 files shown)


nelsonkopliku:
Suggestions should be landed to a certain extend and form in main already.

PR #752, opened 2026-08-14
Adds explicit system env variable for ipv4/ ipv6 handling

gagandeepb:
### Description

Explicit system env. variable for ipv4 vs ipv6 support. Exposed as operator configuration.
Rest of the motivation is same as that for https://github.com/trento-project/web/pull/4615
(This is the wanda equivalent of the changes seen previously on web.)

### How was this tested?

The test package/PTF will be tested by the case reporter, after this PR is reviewed by all involved.

### Documentation changes

Docs changes will be needed and will be done subsequently.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -51,14 +51,19 @@ if config_env() in [:prod, :demo] do
 
   port = String.to_integer(System.get_env("PORT") || "4000")
 
-  # Returns an IP address to bind the HTTP server, based on IPv6 support.
+  # Returns an IP address to bind the HTTP server, based on explicit operator choice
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
Same as in sibling PR in web.

If nobody has objections about adding an extra environment variable I am also fine.

Let's see if we need to adjust the following:
- https://github.com/trento-project/wanda/blob/main/packaging/suse/rpm/systemd/trento-wanda.example

gagandeepb (commented):
@nelsonkopliku Can you please verify the backport target/version milestone here, and approve if it looks fine?

nelsonkopliku (approved):
Thanks @gagandeepb, looks good!

PR #756, opened 2026-08-28
Install libopenssl1_1 for erlang NIFs

nelsonkopliku:
### Description

Adding libopenssl1_1 to have erlang use correct NIFs

Should fix https://github.com/trento-project/web/actions/runs/33155460512

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,7 @@ LABEL org.opencontainers.image.base.digest="latest"
 LABEL io.artifacthub.package.logo-url="https://www.trento-project.io/images/trento-icon.svg"
 LABEL io.artifacthub.package.readme-url="https://raw.githubusercontent.com/trento-project/wanda/refs/heads/main/packaging/suse/container/README.md"
 # Erlang runtime dependencies
-RUN zypper -n in libsystemd0 libopenssl3
+RUN zypper -n in libsystemd0 libopenssl1_1 libopenssl3
 WORKDIR /app
 COPY --from=release /build/_build/$MIX_ENV/rel/wanda .
 VOLUME /usr/share/trento/checks


gagandeepb (commented):
In my understanding, openssl is a build dependency of erlang/otp and subsequently elixir. Do we have some verification that this change fixes the observed behaviour/failures?

nelsonkopliku:
> In my understanding, openssl is a build dependency of erlang/otp and subsequently elixir. Do we have some verification that this change fixes the observed behaviour/failures?
>

Apparently it is a runtime dep since the artifact builds anyway.

How to reproduce:
- build the image against main and run the container (or just run `ghcr.io/trento-project/trento-wanda:demo` `ghcr.io/trento-project/trento-wanda:rolling`) and you'll get `Failed to load NIF library /app/lib/crypto-5.4/priv/lib/crypto: 'libcrypto.so.1.1: cannot open shared object file: No such file or directory'`
- build the image at this branch, run the container and it works

gagandeepb (commented):
Unless I am mistaken, openssl1_1 has been EOL for quite some time now. Do we know why the built artifacts rely on this? This might be something to address in our upstream build and packaging scheme of things rather installing an additional runtime dep here.

nelsonkopliku:
> Unless I am mistaken, openssl1_1 has been EOL for quite some time now
>
That's correct, and erlang26 should support openssl3. That is very strange 😕

skrech:
It seems that the problem with `libcrypt.so` is related to zypper picking up `erlang26` from BCI repositories, instead of our `builddeps` repo.
Apparently, erlang26 was built with openssl1.1, thus its crypto lib is dependent on openssl1.1.

So, maybe we should drop openssl3 for openssl1.1... :( 

Or maybe fiddle around with our erlang build in builddeps and drop the cutting-edge 3.5 openssl version with some `Prefer` directives?

PR #771, opened 2026-09-04
Disable fail-fast in the CI for most jobs

skrech:
### Description

Disable fail-fast on most jobs in the standard CI workflow.

This would help in pinpointing  which actual run is failing and which not. Also, with occasional failures of a single check, the whole group needs to be re-run which takes time on subsequent re-trigger, which unnecessarily wastes dev time.

We have the same in trento-web: https://github.com/trento-project/web/pull/4708

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,6 +62,7 @@ jobs:
     needs: [setup-matrix-env, elixir-deps]
     runs-on: ubuntu-24.04
     strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Thanks! Perhaps I should also add some of the changes from https://github.com/trento-project/web/pull/4466 to Wanda as well.

skrech:
@antgamdia Yes, if they are approved for `trento-web`, it makes sense to add them to wanda as well.

PR #781, opened 2026-09-08
Revert backport workflow back to normal

skrech:
### Description

Reverting the backport workflow to be triggered on `pull_request` event, instead of `pull_request_target`. The backport of older PRs is complete, so every new one would be picked by the standard safe `pull_request` event.


arbulu89 (approved):
🥳

